PR #3, opened 2015-02-16
Bug 1414221: Warning "The log was not applied to the intended LSN" should optionally be an error

gl-sergei:
Bug 1414221: Warning "The log was not applied to the intended LSN" should optionally be an error

Instead of warning print an error message.
Print value of srv_start_lsn right after the error message.

Jenkins build http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/291/

Test bug723318.sh is an artifact left after bzr -> git conversion. It is not present in bzr version of repo.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -6039,11 +6039,13 @@ next_node:
 	       && srv_start_lsn < metadata_last_lsn)) {
 		msg(
 "xtrabackup: ########################################################\n"
-"xtrabackup: # !!WARNING!!                                          #\n"
+"xtrabackup: # !!ERROR!!                                            #\n"
 "xtrabackup: # The transaction log file is corrupted.               #\n"
 "xtrabackup: # The log was not applied to the intended LSN!         #\n"
 "xtrabackup: ########################################################\n"
 		    );
+		msg("xtrabackup: Log applied to lsn " LSN_PF "\n",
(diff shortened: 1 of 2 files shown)


gl-sergei:
Trivial change. Self approving.

jsoriano:
Hi @gl-sergei,
We had this error recovering a backup, but apart of that everything looked perfectly fine, we could even do the copy-back.
The backup we were trying to restore is from a small database in a cluster with very low activity, we suspect that during the backup there were no writes, so there were no log to apply. Could it be possible that it reported this error because it didn't apply any log (even if it was correct)?

gl-sergei:
Hi @jsoriano,
Actually this warning/error indicates that InnoDB redo log copied incorrectly or copied log is corrupted. It might be a result of networking issue or caused by one of following bugs https://bugs.launchpad.net/percona-xtrabackup/+bug/1470847, https://bugs.launchpad.net/percona-xtrabackup/+bug/1464608, https://bugs.launchpad.net/percona-xtrabackup/+bug/1403237 or even another not yet discovered bug.

jsoriano:
I don't think it can be a network issue in our case, we could reproduce it multiple times, and the only common denominator compared to recoveries that work, is that the cluster whose recovery fails didn't have any write during the backup.

It could be https://bugs.launchpad.net/percona-xtrabackup/+bug/1403237 but the thing is that with the warning only everything ended up correctly (showing the warning) and now it shows the error and fails when everything looks perfectly fine.

PR #12, opened 2015-02-25
fixes bug: https://launchpad.net/bugs/1413044

george-lorch:
Deleted line that does 'rm' on existing .qp and .xbcrypt files when using
inobackupex --decrypt and/or --decompress.

Also needs doc updates that if someone is using innobackupex with
--decompress and/or --decrypt, they will need to manually verify
that the result is correct (check to see if xtrabackup_checkpoints
is a readable file that makes sense) and manually delete or move
their original .qp and .xbcrypt files before performing a prepare
and --copy_back.

Jenkins : http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/297/

### DIFF
--- a/storage/innobase/xtrabackup/innobackupex.pl
+++ b/storage/innobase/xtrabackup/innobackupex.pl
@@ -1780,7 +1780,6 @@ sub decrypt_decompress_file {
   }
   print STDERR "$prefix $file_cmd\n";
   system("$file_cmd") && die "$file_cmd failed with $!";
-  system("rm -f $file");
 }
 
 #


gl-sergei:
Goerge, could you please file JIRA request for doc changes or link existing one if you did already. Maybe we should also file feature request to store CRC/MD5/other hash of original chunk within encrypted payload to be able to fail with proper error message in case of wrong encryption key supplied?

gl-sergei:
Doc request is https://jira.percona.com/browse/DOC-141
Blueprint for proper fix is https://jira.percona.com/browse/PXB-173

PR #17, opened 2015-02-26
PXB-170: xtrabackup should not allow to apply incremental backup on  top of the backup prepared without --apply-log-only

gl-sergei:
New status 'log-applied' introduced for backup prepared with
 --apply-log-only. Incremental can be applied only to backup in
'log-applied' state, but not to 'full-prepared' as it was earlier.

http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/295/


george-lorch:
Looks good to me, surprised this wasn't in there from the beginning.

PR #18, opened 2015-02-26
PXB-170: xtrabackup should not allow to apply incremental backup on top of the backup prepared without --apply-log-only

gl-sergei:
Automerge from 2.2
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/35/
Fixes for kill_long_selects are on the way. Will be ported from rewirite-innobackupex-in-c branch.


george-lorch:
Looks good as well, odd that there are 2 commits here unless there was some conflict during the merge forward.

gl-sergei:
Hi George,

It seems just how git merge works. Here is the description of difference between "fast forward" and "merge commit" http://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging

PR #23, opened 2015-03-18
Fixed LP Bug #1401087 by using the openssl command (2.2)

hrvojem:
Fixed LP Bug #1401087 by using the openssl command


gl-sergei:
Approve

PR #25, opened 2015-03-18
PXB-172: Changed page bitmaps support does not work in innobackupex

gl-sergei:
Code responsible for discovery of CHANGED PAGE BITMAPS support
always returned that support is not found.

Fixed discovery of CHANGED PAGE BITMAPS support and adjusted test
backup_locks to reflect the fact that additional flush command is
needed when CHANGED PAGE BITMAPS is in use.

### DIFF
--- a/storage/innobase/xtrabackup/innobackupex.pl
+++ b/storage/innobase/xtrabackup/innobackupex.pl
@@ -3050,6 +3050,19 @@ sub mysql_query {
   }
 }
 
+#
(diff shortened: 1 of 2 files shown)


gl-sergei:
Jenkins: http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/298/

gl-sergei:
Change is trivial. Self approving.

PR #27, opened 2015-03-19
Fixed LP Bug #1307961 - safe-slave-backup explanation is not clear (2.2)

hrvojem:
LP bug: https://bugs.launchpad.net/percona-xtrabackup/2.2/+bug/1307961

### DIFF
--- a/storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
+++ b/storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
@@ -218,11 +218,11 @@ Options
 
 .. option:: --safe-slave-backup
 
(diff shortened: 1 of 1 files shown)


gl-sergei:
Hi Hrvoje!

When I read this paragraph I imagine it works this way:
1. Stop slave and place FTWRL (order is unclear)
2. START and STOP slave until number of slave temp tables is zero or timeout is reached
3. Backup myisam files
4. Release FTWRL

But what really is happening:
1. Stop slave
2. if safe-slave-backup, then repeat START slave, wait some time and STOP slave again, until number of slave temp tables is zero or timeout is reached
3. FTWRL
4. Backup myisam files
5. Release FTWRL

Slave sql thread is indeed stopped during FTWRL stage, but we ensure that Slave_open_temp_tables is 0 before placing FTWRL, which requires some start/stop flapping to give sql thread a chance to advance replication.

gl-sergei:
The reason why order is important here is that given that default timeout is 300 seconds no one will want to use this feature if it holds FTWRL all these 300 seconds.

PR #30, opened 2015-03-22
innobackupex rewritten in C

gl-sergei:
Few decisions which needs to be explained.

I don't use system provided libmysqlclient because it has too much
conflicts with built-in embedded server library. Among the others are
encodings, mysys global and thread-local variables, functions with
the same name implemented differently in different libmysqlclient
versions (distros use client libraries from MySQL 5.1, 5.5, 5.6 and
MariaDB). I managed to make it work, but it looked too bad and was
not trivial to maintain. I didn't look into alternative client
libraries because I don't like this idea in general. I think it not
too bad to use built-in client library because it in fact is already
here and is ready to use.

Innobackupex is kind set of auxiliary functions while main() function
remains in xtrabackup.cc. This is really more cosmetic thing which
doesn't make much sense to me. Besides it is easy to change by moving
main into innobackupex.cc and invoking xtrabackup_backup,
xtrabackup_prepare and so on from there.

innobackupex.cc follows closely after innobacupex.pl. This is done
because it is a rewrite, and doing things this way makes it much
easier not to miss something.

version_check left as a Perl script. I compiled it into the binary as
string literal and invoking it with fpipe(). I definitely don't want
to rewrite version_check it Perl. In fact we can distribute it as a
perl script but it makes it more difficult to locate, because
different distributions may place it into different places. If there
are any other options I can reconsider.

http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/59/


akopytov:
Before I proceed to in-depth reviews, I'd like to clarify the following. My original idea was to make xtrabackup a full replacement for the innobackupex script as a result of the merge. Similar to how mysqlbackup is a one-stop executable in MEB. That is, the xtrabackup binary should support all features provided by innobackupex (with minor syntax differences), while innobackupex should be a compatibility layer emulating the Perl script and its syntax, and all new options being added _only_ to the xtrabackup binary.

As I understand, this PR just rewrites innobackupex in C, but does not touch the xtrabackup binary functionality. Is this a deliberate decision, and if so, what's the reasoning behind it?

gl-sergei:
Alexey,

In fact MEB's mysqlbackup is more like innobackupex at the moment. In fact current implementation is somewhat what you described, but with innobackupex as the full featured command and xtrabackup as the compatibility mode command. Number of customers use their own scripting around xtrabackup (not innobackupex) to backup only InnoDB tables without connecting to server and worrying about MyISAM data, there are also scripts our support and consulting teams use which also based on xtrabackup, I am not sure about RDBA.

I understood that the final goal is to replace both xtrabackup and innobackupex with a single command (xtrabackup or innobackupex). To achieve this we just need to add options to xtrabackup because all code is already in place for it. I used ibx_\* prefix, but it is easy to get rid of this prefix and split things into number of smaller files. One for copying non-innodb tables, another one for dealing with locks, third for gathering server capabilities and so on.

Do you think it is OK to loose compatibility of xtrabackup command for 2.3 release (I'd rather do it for 3.0 release)?

akopytov:
I don't see any issues. Direct xtrabackup usage has not been recommended for a long time. There are important bits of functionality which were only provided by innobackupex even if only InnoDB tables are being backed up or restored.

Any major release is a good time for changes like that, no matter if it's called 2.3, 3.0 or 10.0.

gl-sergei:
So, we are going to change recommendation? Until 2.3 we recommended innobackupex. Now it will be xtrabackup, right?

akopytov:
Yes, and on top of that, I would add a warning when the binary is called by the innobackupex name, saying that innobackupex is provided for compatibility and will be removed in the next major release.

gl-sergei:
Alexey,
Can you start this review with the current code and maybe I will submit other changes as separate MP or it is better to rework this MP? It should not take too much time to rework.

gl-sergei:
I better rework first to save your time.

gl-sergei:
Done.

Xtrabackup now supports all features of innobackupex. They are currently almost identical in terms of features and syntax.

I added few tests to cover some of xtrabackup commands. Few tests adjusted.

Here will be the build http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/61/

PR #31, opened 2015-03-25
Changed the way backups stored in cloud.

gl-sergei:
xbcloud store each chink as separate object with name
backup_name/database/table.ibd.NNNNNNNNNNNNNNNNNNNN,
where NNN... is 0-padded serial number of chunk within file.
Size of chunk produced by xtrabackup and xbstream changed
to 10M.

Each chunk read fully from input stream into the buffer. Then it
uploaded via HTTP. If input chunk is corrupted, then upload aborted
and partially uploaded backup will remain in cloud storage.

Following has been implemented along the way:
1. xbcloud refuses to overwrite the backup with the same name.
2. added ability to download only selected files.

jenkins: http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/62/

Jenkins currently can not test xbcloud. I am checking the status of the task to set up the swift for testing with Jenkins.

### DIFF
--- a/storage/innobase/xtrabackup/src/CMakeLists.txt
+++ b/storage/innobase/xtrabackup/src/CMakeLists.txt
@@ -18,6 +18,7 @@ INCLUDE(curl)
 INCLUDE(libev)
 
 ADD_SUBDIRECTORY(libarchive)
+ADD_SUBDIRECTORY(jsmn)
 
 FIND_GCRYPT()
 FIND_CURL()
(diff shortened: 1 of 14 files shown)


george-lorch:
+1 - I can't really comment much on the protocol interactions correctness as I am pretty ignorant of the real format, but the remainder of things look good to me.

The only thing I would bring up is the larger hard coded value of page to read in a chunk. At one point in the past, we had talked about allowing this to be a configurable option. In some tests, a much larger value proved to improve overall backup performance and in others, a smaller value was desired for other reasons. If you start getting into an XB w/ multiple threads + compression + encryption + xbstream + xbcloud, the amount of memory required as a direct result of the size of this buffer can grow to be pretty big due to the fact that IIRC there is one instance of this buffer per thread, then another buffer of this size for each encryption and compression thread and I think another for xbstream, then more in xbcoud.

I do not remember though why we abandoned the idea, maybe just not enough demand or lost in the noise of other work. So I would maybe suggest adding another option to allow changing the default size of this block (in pages) as part of a new feature.

PR #32, opened 2015-03-27
Changes in cmake to generate source tarball from git tree and not bzr

tplavcic:
This is a change so that "make dist" command works with git tree. This change is in upstream mysql 5.6.23 so it will eventually get merged when xtrabackup is based on that version.
Without it "make dist" creates a source tarball which includes the whole git source tree which is too big so our jenkins release jobs make no sense.

Here's the test release build:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-xtrabackup-2.2-RELEASE/65/

Here's the param build:
http://jenkins.percona.com/job/percona-xtrabackup-2.2-param/299/


gl-sergei:
Looks good. Thank you!

PR #37, opened 2015-04-29
Bug 1448447: Fix for bug 1403237 is incomplete (2.2)

gl-sergei:
At the end of the log copying process XtraBackup invokes
log_group_read_checkpoint_info to grab latest checkpoint LSN which
resets group->lsn_offset to wrong value. As the result last copied
batch of log records is copied from wrong location.

Fix is following:
1. Remember if we chose to use alternative LSN offset location.
2. Allow to try different offset at any time, not only when
   reading the first log block. It handles the case when copying
   started with offset lower than 4G and finished with offset
   higher than 4G.

http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/301/


george-lorch:
+1

PR #42, opened 2015-05-07
Bug 1452307: --decrypt and --decompress do not work with xtrabackup 2.3

gl-sergei:
opt_decrypt and opt_decompress were not properly set by xtranackup.cc.
Relevant test were using innobackupex to decompress/decrypt.

Fix is to set options properly and to adjust test cases.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/90/

PR #43, opened 2015-05-07
Bug 1452375: innobackupex symlink is not created by make install in 2.3

gl-sergei:
Fixed by adding appropriate command to CMakeLists.txt


gl-sergei:
tested locally and http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/91/

PR #48, opened 2015-05-11
Implementation of 'swift-auth-options' blueprint

gl-sergei:
--swift-url renamed to --swift-auth-url
Many memory bugs fixed.
Test case modified and adjusted to able to run in our Jenkins environment.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/103/

PR #49, opened 2015-05-13
bugfix xtrabackup crash during prepare phase, if there are

fungo:
bug #1399471
bugfix xtrabackup prepare crash.


gl-sergei:
Thank you for your contribution! I have few comments:
- we need test case in order to accept this patch, you can base yours
   on comment #4 I posted here:
   https://bugs.launchpad.net/percona-xtrabackup/+bug/1399471
- is there a need to pass `dict_locked` into
   `dict_check_if_table_exists`? `dict_check_if_table_exists` invoked
   only once with `dict_sys->mutex` locked. I would suggest to
   simplify this function.
- change in `row0mysql.cc` is essentially a double copy-paste of
   existing SQL statements. Instead, I suggest to two invoke `DELETE
   FROM SYS_COLUMNS ...` and `DELETE FROM SYS_TABLES ...` in separate
   `que_eval_sql`.
  
  Please let me know if you are wiling to work further on your patch,
  otherwise I will go ahead and make these changes by myself.

fungo:
- OK, testcase will be added shortly
- `dict_check_if_table_exists` is a newly added function, cause I didn't found existed function in InnoDB code to check whether table exists, so I made this one, and I also want to make this function more generic, maybe it will be used again someday in the future, that's why there is `dict_locked` parameter. For this situation only, one tablename parameter is enough.
- yeah, change in `row0mysql.cc` seemed redundant，but I don't think two invoke of  `que_eval_sql`  is a more grace way , cause invoke 2 times of the same function to do one thing is double cost, maybe a sql string prepare before invoke `que_eval_sql` is a better one.

gl-sergei:
- well, actually there is a function `dict_check_if_system_table_exists` which perfectly fits for this patch. It is trying to locks `dict_sys->mutex`, but this is not an issue (see next item)
- you don't really need to check if `SYS_TABLESPACES` exists for every table you want to drop. It may give even worse overhead comparing to additional `que_eval_sql`. You can check it once upon dictionary initialization.
- fil0fil.cc has similar SQL in `fil_remove_invalid_table_from_data_dict`, should we update it as well? It seems like a bit buggy for MySQL 5.6

fungo:
Sorry for my late response, I have noticed that you submitted a  [new commit](https://github.com/percona/percona-xtrabackup/commit/32dcc43d83537ec870ad45f3e64dad1247979269) to fix this bug, so this pull request can be closed.

PR #51, opened 2015-05-14
Bug #1454815: xtrabackup prints warnings for tables, not included into partial backup

darnaut:
Whenever preparing a partial backup, several warning messages are
printed about tables that don’t exist. This is due to tables that are
not physically part of the backup (that is, no ibd files) but still
exist in InnoDB’s data dictionary.

Apart from these warnings being misleading to casual users, it can also
lead to overly large and verbose logs if the partial backup is taken on
a instance which has a large number of tables that are not part of the
backup.

When a filtering option (--databases, --tables, etc) is provided to
prepare, assume that the backup can be partial and do not print warnings
about tables that don't exist.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/312/

gl-sergei:
Thank you for contribution!

PR #56, opened 2015-05-21
Bug 1444541: XtraBackup 2.2.10 does not work with MariaDB 10

gl-sergei:
MariaDB lists INNODB_CHANGED_PAGES in INFORMATION_SCHEMA.PLUGINS, but
it doesn't support 'FLUSH NO_WRITE_TO_BINLOG CHANGED_PAGE_BITMAPS'.

This is regression after fixing bug #1436793 (changed page bitmaps
support was not correctly detected).

Workaround is to turn off changed pages flush for MariaDB.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/315/

PR #58, opened 2015-05-21
Bug 1449834: innobackupex does not record binlog file and pos when GT…

gl-sergei:
…ID is enabled

Sometimes it is useful to know file based binlog coordinates even
though GTID is enabled.

Fix is to print file based coordinates and to store them in
xtrabackup_binlog_info regardless of GTID mode.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/316/

gl-sergei:
UPD: http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/320/

PR #63, opened 2015-05-22
Bug 1399471: XtraBackup 2.2.x crash on prepare when backing up MySQL/…

gl-sergei:
…Percona Server 5.5.x

When InnoDB rolling back active transactions on recovery it also
removes temporary tables. When temporary table removed from data
dictionary it also removed from system tables like SYS_TABLES,
SYS_FOREIGN, etc. The culprit here are SYS_DATAFILES and
SYS_TABLESPACES which were introduced in 5.6 and absent in backup
taken from MySQL 5.1/5.5.

The fix is not remove from SYS_DATAFILES and SYS_TABLESPACES if they
are absent. Patch also updates
fil_remove_invalid_table_from_data_dict to match
row_drop_table_for_mysql.

Patch contributed by Fungo Wang.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -41,6 +41,7 @@ Created 10/25/1995 Heikki Tuuri
 #include "mtr0mtr.h"
 #include "mtr0log.h"
 #include "dict0dict.h"
+#include "dict0priv.h"
 #include "page0page.h"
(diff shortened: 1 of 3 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/319/

PR #70, opened 2015-06-03
Bug 1461129: Target sql_embedded should depend on GenDigestServerSource

gl-sergei:
Source file 'sql_yacc.yy' needs generated file 'lex_token.h'. This
file is generated by 'GenDigestServerSource', which is dependency for
'sql', but not for 'sql_embedded'.


gl-sergei:
Builds normally http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/321/

PR #71, opened 2015-06-03
Bug 1461129: Target sql_embedded should depend on GenDigestServerSource (2.3)

gl-sergei:


gl-sergei:
Builds normally http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/321/

PR #75, opened 2015-06-19
Fix bug1466888 Switch to git instead of bzr in RPM build

roidelapluie:


tplavcic:
Thank you very much for contribution!
Notice I've created a little bit different PR's (linked) so I'll just close this one without directly merging.

PR #79, opened 2015-06-23
Bug 1467574: xtrabackup should initialize current_thd

gl-sergei:
Set thread-local variable THR_THD to NULL. This patch also fixes and
adds test case for bug #1415191.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/324/

PR #80, opened 2015-07-05
Bug 1460138: xtrabackup segfaults during --prepare of certain FTS pages

gl-sergei:
xtrabackup crashed while trying to get value of THDVAR
innodb_support_xa.

Fixed by checking whether trx->mysql_thd is NULL (it should always be
NULL for xtrabackup) and disabling XA transactions in this case.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/325/

PR #81, opened 2015-07-08
Bug 1470847: Percona Xtrabackup: Starting LSN is bigger than ending LSN

gl-sergei:
xtrabackup completes successfully even if redo log has not been
copied for whatever reason. It leads to confusion when backup
considered correct while it actually corrupt.

Fixed by checking that copied redo log contains the latest checkpoint
LSN.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/330/

PR #85, opened 2015-07-23
Removed python-sphinx10 build dependency in rpm for centos 5/6 (JEN-378) - 2.3

tplavcic:
This is for: https://jira.percona.com/browse/JEN-378

There no longer exists python-sphinx10 package in epel repo for centos 5/6 so we need to remove this dependency on those packages.
If someone wants to build xtrabackup on their own they will need to install sphinx and docutils python modules in some python way.
Centos 7 is not affected because there is native package for python-sphinx and python-docutils.


tplavcic:
Just to clarify - this dependency is for build only - not for installation.
It was used for building man pages.

PR #86, opened 2015-07-27
Bug 1475487: xtrabackup handling of space id conflicts

gl-sergei:
When there are two or more tablespaces with the same ID in the data
directory, xtrabackup picks up the first one by lexical order.

Better choice would be to pick one, actually referenced by data
dictionary and that is what MySQL does after clean shutdown.

However, we cannot rely on data dictionary on "prepare" and
"apply-log" stages. MySQL on crash recovery just refuses to start
with error message.

Xtrabackup silently skips tablespaces with uninitialized space_id,
however because of wrong condition it also skips space_id conflicts.

Fixed by changing condition in fil_load_single_table_tablespace to
not only verify return code form
fil_validate_single_table_tablespace, but also check that space_id is
not defined before silently skip tablespace.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -4261,7 +4261,8 @@ fil_load_single_table_tablespace(
 			os_file_close(remote.file);
 			mem_free(remote.filepath);
 
-			if (srv_backup_mode) {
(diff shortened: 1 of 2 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/335/

PR #87, opened 2015-07-27
Bug 1464608: Failing assertion: byte_offset % OS_MIN_LOG_BLOCK_SIZE =…

gl-sergei:
…= 0, lsn_offset = 18446744073709551615

Fix for bug 1403237 was to try to read redo log from different
offset. However, the patch reverted the fix for bug 1206309. As a
result xtrabackup tried to read redo log from incorrect offset.

Fix is to check condition for the bug 1206309 before attempting to
switch to different offset.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/336/

PR #98, opened 2015-09-10
Bug 1493015: Sporadic bug 1410339 failures in Jenkins

gl-sergei:
For older bash versions killing child bash process caused cleanup to
be invoked which led to backup directory being deleted and xtrabackup
was unable to create 'xtrabackup_log_copied' file and innobackupex
falsely assumed that backup completed with errors.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/340/#showFailuresLink
Shows no failures of bug1410339.sh.

PR #99, opened 2015-09-15
Bug 1495367: xtrabackup 2.3 does not set wait_timeout session variable

gl-sergei:
xtrabackup needs to set wait_timeout in order to prevent server to
kill the connection while xtrabackup is copying data files.


gl-sergei:
http://jenkins.percona.com/job/percona-xtrabackup-2.3-param/123/

PR #100, opened 2015-09-16
Bug 1242309: xtrabackup_logfile not compressed in XtraBackup 2.1+

gl-sergei:
xtrabackup_logfile goes to "ds_meta" datasink, which is not
compressed allowing to use compressed backup as a base for
incremental.

xtrabackup_logfile cannot go "ds_data" datasink, because it needs to
be saved in temporary directory and then streamed at the end of the
backup process, otherwise it will interfere with the innobackupex
stream.

Fix is to create new "ds_redo" datasink, which goes the same path as
"ds_meta", except for it is compressed if compressed backup created.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/341/

george-lorch:
+1

PR #102, opened 2015-09-16
Bug 1457016: Out of source tree builds are broken in 2.3 trunk

gl-sergei:
INSTALL_SYMLINK macro messed up with dependencies and caused error
for out of source build. Fix is to replace INSTALL_SYMLINK with
INSTALL.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/125/

PR #103, opened 2015-09-17
Bug 1454692: xtrabackup 2.3 should add timestamps to stderr output

gl-sergei:
Introduce two functions for logging messages msg() which doesn't add
timestamps and msg_ts() which does. Tools like xbstream, xbcrypt,
etc. use msg() and their output don't contain timestamps. xtrabackup
adds timestamps when needed (when file started or ended copying, when
FTWRL issued, etc.).


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/126/

PR #104, opened 2015-09-18
Bug 1452387: xtrabackup never verifies return code of ds_close

gl-sergei:
Added ds_close return code handling.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/127/

george-lorch:
+1

PR #106, opened 2015-09-30
Bug 1500508: xbcloud fails to create container with "error: curl_easy_perform() failed: Failed sending data to the peer"

gl-sergei:
Fixed by doing following:
1. Always set read data callback for PUT method
2. Set input file size when appropriate

Change test case to run it only on PS 5.6. This will both provide
sufficient coverage and reduce load on swift server.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/135/

PR #108, opened 2015-10-05
Add support for encrypted and page compressed tables supported

janlindstrom:
by MariaDB 10.1

In backup encrypted tables remain encrypted. Similarly, page
compressed tables are still compressed.

### DIFF
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -1304,6 +1304,37 @@ fsp_page_create(
 }
 
 /**********************************************************************//**
+Compute offset after xdes where crypt data can be stored
+@return	offset */
+ulint
+fsp_header_get_crypt_offset(
+/*========================*/
+	ulint   zip_size, /*!< in: zip_size */
+	ulint*  max_size) /*!< out: free space available for crypt data */
+{
+	ulint pageno = 0;
+	/* compute first page_no that will have xdes stored on page != 0*/
+	for (ulint i = 0;
+	     (pageno = xdes_calc_descriptor_page(zip_size, i)) == 0; )
+		i++;
+
+	/* use pageno prior to this...i.e last page on page 0 */
+	ut_ad(pageno > 0);
+	pageno--;
+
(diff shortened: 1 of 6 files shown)


gl-sergei:
This PR breaks at least following test cases:
bug1291299 ib_part_include ib_part_tf_innodb ib_part_tf_innodb_stream innodb_page_size remote_tablespaces

Related to remote tablestaces and pages with non default size.

janlindstrom:
Hi,

How you run those tests ?

R: Jan

gl-sergei:
You need to 
1. build xtrabackup
2. cd storage/innobase/xtrabackup/test
3. run ./bootstrap.sh <innodb-flavor> (innodb56 for example) or ./bootstrap.sh server-tarball-name.tar.gz
4. then ./run.sh -f to run all tests or ./run.sh -t t/test-name.sh to run specific tests
Currently we don't run test suite on MariaDB because we don't officially support it, so you will likely see many failed tests.

hydrapolic:
@janlindstrom, any update on this please?

janlindstrom:
This needs more work before next PR, ability to understand that pages are encrypted is not enough. Log records should be decrypted to be able to apply them at restore is required. This needs some kind of encryption plugin for backup.

hydrapolic:
@janlindstrom, thanks for the info. That would be great as it would allow us to deploy page compression on production servers.

janlindstrom:
Support for page compression is easier to add, do you need only that?

hydrapolic:
Yes, page compression would be enough. Encryption is not needed at the moment, but would be nice to have a single backup tool for all purposes of course. Also for bringing up slaves, etc. I think using mysqldump is not an option for production databases with a fair amount of data.

deepvoice:
+1

disaster123:
+1 for page compression in xtrabackup 2.4

hydrapolic:
Any update on this?

janlindstrom:
Hi,

This has been on hold a while to find out better solution. I will try to update the PR soonish.

hydrapolic:
@janlindstrom, thank you very much!

PR #111, opened 2015-10-14
Bug 1503964: innobackupex stream corrupted "xb_stream_read_chunk():

gl-sergei:
wrong chunk magic at offset 0x0"

innobackupex wrote error message to STDOUT instead of STDIN which
broke xbstream and tar stream.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/345/

PR #113, opened 2015-10-14
Bug 1505865: Incremental backups do not work with MariaDB below 10.1.6

gl-sergei:
MariaDB has support for changed page bitmaps but lacks support for
FLUSH CHANGED_PAGE_BITMAPS statement in versions below 10.1.6

Fix is to turn off FLUSH CHANGED_PAGE_BITMAPS for MariaDB versions
below 10.1.6.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.2/job/percona-xtrabackup-2.2-param/344/

PR #125, opened 2015-10-23
Fixed --swift-url

NoodlesNZ:
--swift-url has been replaced with --swift-auth-url. Some of the examples don't reflect this change.


gl-sergei:
Thank you for contribution!

PR #129, opened 2015-12-09
Bug 1509812: backup fails with history on server enabled and binary l…

gl-sergei:
…og disabled

When trying to store history record on server, xtrabackup is trying
to obtain binary log position from mysql_binlog_position variable
which is set only if binary log is enabled, otherwise it is NULL.

When mysql_binlog_position xtrabackup fails when trying to invoke
strlen(mysql_binlog_position).


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/156/

PR #145, opened 2016-01-14
Bug 1533722: xtrabackup crashed during apply 5.5 backup set with utf8_general50_ci collation

fungo:
Charset collation 'utf8_general50_ci'(253) was added in PS 5.1/5.5,
and is not supported by Oracle/MySQL against which xtrabackup is
compiled. This will lead to crash when recovring.

Fix:
Print warning when encounter 'utf8_general50_ci' charset collation,
and leave the decisions(continue or discard recovery) to users.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -1610,6 +1610,12 @@ innobase_get_cset_width(
 				sql_print_warning(
 					"Unknown collation #%lu.", cset);
 			}
+		} else if (cset == 253) {
+			/* Charset collation utf8_general50_ci is not support in Oracle MySQL */
+			ut_print_timestamp(stderr);
+			fprintf(stderr,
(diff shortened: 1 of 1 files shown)


gl-sergei:
Hi! Thank you for contribution. Instead of bailing out with error message, I would prefer to port this patch to XtraBackup: https://code.launchpad.net/~akopytov/percona-server/utf8_general50_ci-5.5/+merge/67427

fungo:
Hi, PS 5.6 also removed this collation, so maybe we don't need port this collation, it won't do much harm for PXB to apply log.

Besides, users may want use backup to upgrade, such as restore a PS 5.5 backup to Oracle/PS 5.6, it will be helpful to notice user there exists unsupported collation, new version mysqld will crash when table contain this collation is used. So I think error message is necessary.

gl-sergei:
This have been fixed in https://github.com/percona/percona-xtrabackup/pull/438

PR #160, opened 2016-02-04
Bug 1533542: innobackupex silently skip extra arguments

fungo:
Should print error about extra arguments and exit.


gl-sergei:
Thank you!

PR #179, opened 2016-03-04
Bug 1508450: secure_auth forced client side in 2.3.2

gl-sergei:
Option --secure-auth added, set by default.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/162/

PR #188, opened 2016-03-25
Bug 1554235: Innobackupex fails when using --throttle

gl-sergei:
Race condition between main thread and io_watching_thread could cause
accessing memory used by wait_throttle after it was already freed up.

Bug analysis by Sunguck Lee.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/163/

PR #189, opened 2016-03-25
Fix for bug 1554235 in 2.4

gl-sergei:


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.4-param/23/

PR #194, opened 2016-04-06
Bug #1566228: XB crashes with insufficient param.

gl-sergei:
Fix includes:
- fixed memory error when no files were copied on copy-back/move-back
- error is printed when 'datadir' from SHOW VARIABLES doesn't exist
  (aborts backup when remote host is specified)
- error is printed when no 'datadir' is specified for
  copy-back/move-back


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/164/

PR #196, opened 2016-04-15
Bug 1570560: xtrabackup-v2 SST is broken on MariaDB 10.1/Galera  w/ b…

gl-sergei:
…inary logs enabled

Condition for have_flush_engine_logs was wrong.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -374,7 +374,7 @@ get_mysql_vars(MYSQL *connection)
 		have_galera_enabled = true;
 	}
 
-	if (strcmp(version_var, "5.5") >= 0) {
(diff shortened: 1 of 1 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/165/

abg:
I don't think this fixes the MariaDB / Galera problem though, which needs flush engine logs to happen before copying the binary log.

The binlog gets copied here:

https://github.com/gl-sergei/percona-xtrabackup/blob/2.3-xb-bug1570560/storage/innobase/xtrabackup/src/backup_copy.cc#L1271

But the flush engines happens later:

https://github.com/gl-sergei/percona-xtrabackup/blob/2.3-xb-bug1570560/storage/innobase/xtrabackup/src/backup_copy.cc#L1281

After the (possibly incomplete) binlog was already archived.

PR #198, opened 2016-05-01
Bug 1555423: Fix for 1322658 not ported to 2.3 and 2.4 of Xtrabackup

gl-sergei:
The piece of --copy-back logic copying remote tablespaces back to the
place specified in .isl files was missing, but remote_tablespaces.sh
did not fail because InnoDB was still able to find such tablespaces
inside datadir.

The fix includes:
- attempt to find .isl file and copy .ibd into correct destination
  directory
- check presence of t.ibd inside of remote directory in
  remote_tablespaces.sh


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/166/

PR #201, opened 2016-05-12
Bug 1570560: Fix ordering of FLUSH ENGINE LOGS for MariaDB 10.x

abg:
This extends PR https://github.com/percona/percona-xtrabackup/pull/196 to ensure that FLUSH NO_WRITE_BINLOG ENGINE LOGS is run before copying the last binlog for GTID state during the handling of --galera-info bits in order to fix a problem with MariaDB/GTID.

Presently, FLUSH ENGINE LOGS is not run at all for MariaDB and even with PR https://github.com/percona/percona-xtrabackup/pull/196,  FLUSH ENGINE LOGS is only run after the binlog is copied.   Under load, this can result in MariaDB failing binlog recovery since the copied binlog may have only a single Binlog_checkpoint event referencing a previous (missing) binlog file.  Executing FLUSH ENGINE LOGS prior to this copy ensures a Binlog_checkpoint event is present for the current binlog under consideration.

This avoids a serious headaches when using MariaDB/Galera + xtrabackup-v2 for SST.

See https://jira.mariadb.org/browse/MDEV-9423 for more details on the MariaDB problem.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1277,11 +1277,7 @@ backup_start()
 		write_binlog_info(mysql_connection);
 	}
 
-	if (have_flush_engine_logs) {
-		msg_ts("Executing FLUSH NO_WRITE_TO_BINLOG ENGINE LOGS...\n");
-		xb_mysql_query(mysql_connection,
-			"FLUSH NO_WRITE_TO_BINLOG ENGINE LOGS", false);
-	}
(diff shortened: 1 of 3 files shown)


koopernick:
If it's possible to push this fix. I really need it for my set up.
Thanks for the thread btw.

starouille:
i also don't understand why this is not comitted in a new version?   Without any fix, xtrabackup is not working in full sync when you have lot of writes.. :(

gl-sergei:
See https://bugs.launchpad.net/percona-xtrabackup/+bug/1570560/comments/10

PR #211, opened 2016-05-24
Bug 1582130: Database page corruption detected at page 1 with

gl-sergei:
```
         encryption enabled table full backup
```

When table is both compressed and encrypted xtrabackup did not
detect them well. Problem was that xtrabackup passed incorrect buffer
to is_compressed function.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.4-param/51/

PR #214, opened 2016-06-08
Bug 1590351: xtrabackup --backup crashes with empty keyring file

gl-sergei:
Xtrabackup was trying to print error message using logger which has
not been initialized yet.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.4-param/53/

PR #215, opened 2016-06-20
Fixed lp:1222777. Changed FSF address for new one

gen1us2k:


gl-sergei:
Thank you for contribution. We need GCA branches in order to be able to upmerge changes. I'll create proper ones for 2.2, 2.3 and 2.4 based on your patch.

gl-sergei:
#221, #222, #223

PR #224, opened 2016-07-04
Bug 1517629: xtrabackup fails with binlogs in non-standard directory

gl-sergei:
Bug is specific to `MariaDB' with`Galera' when binary logs are stored
outside of `datadir'.`MariaDB' doesn't expose variable
`log_bin_basename' which is used by`xtrabackup' to determine the
location of binary logs. When this variable isn't available,
`xtrabackup' falls back to`datadir'.

Fix changes behaviour as following:
1. Use `log-bin' if it is set in`my.cnf' or passed as command line
   argument and contains directory part.
2. Use `log_bin_basename' if present.
3. Fallback to `./' which is`datadir'.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/177/

PR #226, opened 2016-07-06
Bug 1599397: equal_paths should not crash if paths do not exist

gl-sergei:
equal_paths returns false if any of parameters points to path which
does not exist.

directory_exists now respects `create` parameter.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/180/

PR #233, opened 2016-07-10
add option to reverse regex match

lalor:
Recently, I want to backup database but exclude some tables , for example slow_log, general_log, I can't do this through `--tables` option. therefore, I have add a new option `--reverse-regex-match` to implement that.


gl-sergei:
Thank you for contribution, we decided to go with --exclude option (released in 2.4.7 https://www.percona.com/doc/percona-xtrabackup/2.4/release-notes/2.4/2.4.7.html)

PR #234, opened 2016-07-11
Bug 1568009: Assertion failure in xtrabackup.cc: lsn_chosen == 1

gl-sergei:
Percona Server 5.5 had support for redo logs larger than 4G. MySQL added
support for redo logs larger than 4G in 5.6. PS 5.5 and MySQL 5.6 store
LSN offset for large log files at different places inside redo log
header.

Xtrabackup is trying to guess which offset is better to use by trying to
read from each one and compare log block numbers. Assert "lsn_chose ==
1" is displayed when both LSNs look correct, but they are different. In
this case we simply cannot pick up any.

Since innobackupex was merged with xtrabackup, we can now use additional
information to make the decision. Server version and its flavor are
available for us at the backup time.

When server version and flavor are known, xtrabackup will make decision
based on this information.

The fix includes some cleanup of version detection code, such as using
version numbers where possible instead of version strings.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -43,7 +43,6 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <mysql.h>
 #include <mysqld.h>
 #include <my_sys.h>
-#include <fnmatch.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/184/

PR #236, opened 2016-07-11
Bug 1550322: Prepare error: Tablespace size stored in header is X pages,

gl-sergei:
```
         but the sum of data file sizes is only Y pages
```

This is a regression of 5.7 merge. 5.6-based xtrabackup skipped the
check for tablespace size in the case of autoextend system tablespace.
Tablespace will be automatically extended as needed.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.4-param/73/

PR #237, opened 2016-07-11
Bug 1582456: Incremental restore in XB 2.4 requires 1+GB of RAM

gl-sergei:
With upstream change of `UNIV_PAGE_SIZE_MAX` from 16K to 64K, the size
of `incremental_buffer` in `xtrabackup_apply_delta` becomes 1G. In fact,
we don't need to allocate such a large buffer for smaller pages. What we
really need is `page_size / 4` pages.

The fix doesn't address the issue for tablespaces with 64K page size,
but it works better in default case.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -5931,13 +5931,12 @@ xtrabackup_apply_delta(
 
 	os_file_set_nocache(dst_file, dst_path, "OPEN");
 
-	/* allocate buffer for incremental backup (4096 pages) */
+	/* allocate buffer for incremental backup */
 	incremental_buffer_base = static_cast<byte *>
-		(ut_malloc_nokey((UNIV_PAGE_SIZE_MAX / 4 + 1) *
-			   UNIV_PAGE_SIZE_MAX));
+		(ut_malloc_nokey((page_size / 4 + 1) * page_size +
+			UNIV_PAGE_SIZE_MAX));
 	incremental_buffer = static_cast<byte *>
-		(ut_align(incremental_buffer_base,
-			  UNIV_PAGE_SIZE_MAX));
+		(ut_align(incremental_buffer_base, UNIV_PAGE_SIZE_MAX));
 
 	msg("Applying %s to %s...\n", src_path, dst_path);
 


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.4-param/68/

andvgal:
Please also add malloc return value check as allocation failure leads to not so descriptive SIGBUS.

gl-sergei:
Actually, `ut_alloc_nokey` reports OOM errors. See `ut0new.h`:

``` c++
        for (size_t retries = 1; ; retries++) {

            if (set_to_zero) {
                ptr = calloc(1, total_bytes);
            } else {
                ptr = malloc(total_bytes);
            }

            if (ptr != NULL || retries >= alloc_max_retries) {
                break;
            }

            os_thread_sleep(1000000 /* 1 second */);
        }

        if (ptr == NULL) {
            ib::fatal_or_error(m_oom_fatal)
                << "Cannot allocate " << total_bytes
                << " bytes of memory after "
                << alloc_max_retries << " retries over "
                << alloc_max_retries << " seconds. OS error: "
                << strerror(errno) << " (" << errno << "). "
                << OUT_OF_MEMORY_MSG;
            if (throw_on_error) {
                throw(std::bad_alloc());
            } else {
                return(NULL);
            }
        }
```

and `m_oom_fatal` is set `true` by default. To make allocator don't crash, one must use `.set_oom_not_fatal()`.

Will crash on OOM error:

``` c++
#define ut_malloc_nokey(n_bytes)    static_cast<void*>( \
    ut_allocator<byte>(PSI_NOT_INSTRUMENTED).allocate( \
        n_bytes, NULL, __FILE__, false, false))
```

Will not crash on OOM error:

``` c++
#define ut_zalloc_nokey_nofatal(n_bytes)    static_cast<void*>( \
    ut_allocator<byte>(PSI_NOT_INSTRUMENTED). \
        set_oom_not_fatal(). \
        allocate(n_bytes, NULL, __FILE__, true, false))
```

andvgal:
OK. Perhaps, exception is a good solution for clean code in general, but then I would add extra `const char *` parameter for description of allocation in progress to be included in exception message.

Of course, this is a more generic note for improvements then for this specific issue.

gl-sergei:
Then you should file bug report to http://bugs.mysql.com/ for InnoDB team :)

knocte:
Is there a workaround for this if we cannot wait for a new version of xtrabackup? (I'm using v2.4.3 and being affected by this bug.) Thanks

andvgal:
@knocte you can try to add >1GB of spare swap before calling xtrabackup.

knocte:
thanks, but just realized 2.4.4 is available for download here: https://www.percona.com/downloads/XtraBackup/LATEST/ , would `percona-xtrabackup-24_2.4.4-1.xenial_amd64.deb` include this fix?

gl-sergei:
@knocte yes, this fix is included in 2.4.4.

knocte:
Thanks guys, just tested it, and the bug has indeed gone away. If you give me a btc address I'll give a small tip to the committer :)

PR #242, opened 2016-07-21
Bug 1604367: Failed to take full backup -> CORRUPT LOG RECORD FOUND

gl-sergei:
During the backup when master key rotated, one could see the warning:

"############### CORRUPT LOG RECORD FOUND ##################"

It means the MySQL server is trying to store the key into the tablespace
header. This key is encrypted with newly generated master key, which is
not accessible by xtrabackup yet.

The fix is to omit this check. Xtrabackup does quick integrity checks of
the log file when copying which are sufficient.


gl-sergei:
Looks like the issue is deeper and the same error can be seen during prepare.

PR #245, opened 2016-08-31
Bug 1611568: target-dir no longer relative to current directory

gl-sergei:
`my_load_path` changed its behavior in 5.7, when directory name is not
started with "./", it doesn't expand it.

During the merge, `my_load_path` has been replaced with `fn_format`
which doesn't work as expected when directory name isn't started with
"./" and is not exists (`fn_format` expands directory when it is exists,
that is why `bug489290.sh` test passed).

This patch reverts back to `my_realpath`, but now we pass current
working directory as prefix, so it will be prepended to the path when it
isn't started with "./".

Test case `bug489290.sh` expanded to check various combinations of
relative and absolute paths, including directories starting with "./"
and "../", existing and non-existing directories.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -7658,6 +7658,7 @@ int main(int argc, char **argv)
 {
 	int ho_error;
 	char **argv_defaults;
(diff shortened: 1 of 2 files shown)


ioquatix:
Having this problem too, behaviour was unexpected.

PR #246, opened 2016-09-06
Fixed: Bug #1594990, Bug #1620521, and DOC-59 (2.3)

hrvojem:
Updated full backup example to use xtrabackup
Fixed doc8 issues for part of the documents


gl-sergei:
LGTM

PR #247, opened 2016-09-07
Fixed: Bug #1594990, Bug #1620521, and DOC-59 (2.4)

hrvojem:
Updated full backup example to use xtrabackup
Fixed doc8 issues for part of the documents


storage/innobase/xtrabackup/doc/source/using_xtrabackup/privileges.rst
```diff
@@ -68,7 +70,7 @@ alteration, see :option:`mysql --help` for details.
 
    In case of multiple server instances the correct connection parameters
    (port, socket, host) must be specified in order for |xtrabackup| to talk to
-   the correct server. 
+   the correct server.
 
 
 Permissions and Privileges Needed
```
gl-sergei, line 76: Should this page also mention PROCESS privilege for 2.4 since this PR fixing https://bugs.launchpad.net/percona-xtrabackup/+bug/1594990 ?

hrvojem: It does, check line: 128.

gl-sergei: ah... PROCESS was listed already... then we need to change its description

```
PROCESS in order to see all threads which are running on the server (see Improved FLUSH TABLES WITH READ LOCK handling),
```

to mention `SHOW ENGINE INNODB STATUS`

PR #249, opened 2016-09-19
Bug 1624473: Issue with --safe-slave-backup option in 2.4.4

gl-sergei:
Safe slave backup algorithm performed too short delays between retries.
Original interval was 3 seconds, but 2.3 and 2.4 version performed 3
milliseconds delays.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/186/

PR #250, opened 2016-09-23
Bug 1623210: xtrabackup failed to write metadata but the backup still…

gl-sergei:
… succeeded

Bug 1557027: Error message about missing metadata is misleading

Abort xtrabackup when it was unable to write or read metadata file.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/188/

PR #258, opened 2016-10-31
Fixes for 1638017 & 1638020

txominb:
1638017 - EMC ECS 204 auth status
1638020 - Etag tracking case sensitive

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud.cc
@@ -602,7 +602,8 @@ swift_temp_auth(const char *auth_url, swift_auth_info *info)
 			goto cleanup;
 		}
 		curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &http_code);
-		if (http_code != 200) {
+		if (http_code != 200 &&
+			http_code != 204) {
 			fprintf(stderr, "error: authentication failed "
 				"with response code: %ld\n", http_code);
 			res = CURLE_LOGIN_DENIED;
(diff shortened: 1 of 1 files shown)


storage/innobase/xtrabackup/src/xbcloud.cc
```diff
@@ -1079,9 +1080,9 @@ size_t upload_header_read_cb(char *ptr, size_t size, size_t nmemb,
 	connection_info *conn = (connection_info *)(data);
 	char etag[33];
 
-	if (get_http_header("Etag: ", ptr, etag, array_elements(etag))) {
+	if (get_http_header("ETag: ", ptr, etag, array_elements(etag))) {
```
gl-sergei: If you take a look at the [API documentation](http://developer.openstack.org/api-ref/object-storage/?expanded=create-or-replace-object-detail), you will see this example response:

```
HTTP/1.1 201 Created
Last-Modified: Fri, 17 Jan 2014 17:28:35 GMT
Content-Length: 0
Etag: 8b1a9953c4611296a827abf8c47804d7
Content-Type: text/html; charset=UTF-8
X-Trans-Id: tx4d5e4f06d357462bb732f-0052d96843
Date: Fri, 17 Jan 2014 17:28:35 GMT
```

At the same time the doc tells to pass `ETag` header from client. Probably different implementations are using different case. It is not correct behaviour, because according to RFC2616 header names are case sensitive:

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

But, I think most HTTP clients do not bother. So, I would change `get_http_header` implementation to use `strncasecmp` instead of `strncmp`.

storage/innobase/xtrabackup/src/xbcloud.cc
```diff
@@ -602,7 +602,8 @@ swift_temp_auth(const char *auth_url, swift_auth_info *info)
 			goto cleanup;
 		}
 		curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &http_code);
-		if (http_code != 200) {
+		if (http_code != 200 &&
+			http_code != 204) {
```
gl-sergei: Please change indentation to align second `http_code` with the first one. See the line 772 for example.

gl-sergei (changes requested):

txominb:
New commit includes the changes as described.  Tested the change to get_http_header and that does represent a better solution.

gl-sergei:
LGTM. Thank you!

gl-sergei (approved):

PR #261, opened 2016-11-01
Bug 1633448: xtrabackup 2.4 doesn't check the log block checksums

gl-sergei:
Fixed the check and added new test case verifying 5.7's
`innodb_log_checksums' setting.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/94/

PR #264, opened 2016-11-09
Bug 1639764: Test t/bug1254227.sh failing on PXC 5.7

gl-sergei:
This test intended to check that xtrabackup rolls back prepared XA
transactions. Since PXC doesn't work well with XA transactions (and PXC
5.7 started to complain with error message "WSREP doesn't support XA
transaction"), this test has to be skipped for PXC.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/201/

PR #272, opened 2016-11-17
Update xtrabackup.cc

elshadaghazade:
Bug fix for https://bugs.launchpad.net/percona-xtrabackup/2.3/+bug/1642329


gl-sergei:
This patch actually breaking xtrabackup :) Now if I run `xtrabackup --backup --target-dir=./` it tells me `Parameter is wrong`. But, indeed, no more crash!

elshadaghazade:
:D :D

PR #274, opened 2016-11-22
Bug 1641612: New compilation warnings with GCC 6

gl-sergei:
Fixed by backporting pieces of code from MySQL (commits c6a1902,
f254870b, f38d67f7, 7b899ac) and PS (fix for bug #1641612).


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/213/

PR #289, opened 2016-11-30
try to hide password from ps output the same way mysql client does

hholzgra:
now trying to overwrite the password in the same way as the mysql command line client does

https://bugs.launchpad.net/percona-xtrabackup/+bug/907280


gl-sergei:
Thank you for contribution.
The fix looks good to me. Can you please create PR for 2.3? It then will be merged into 2.4 without issues.

hholzgra:
Done as requested, it will not directly merge into 2.4 without issues though due to the change in my_strdup() arguments for P_S

PR #292, opened 2016-12-09
Bug 1648322: xtrabackup --move-back is not always restoring out-of-da…

gl-sergei:
…tadir tablespaces to their original directories

When copy-back walks files and copies them back to datadir, it is
looking for corresponding .isl files as it passing by .ibd file. If .isl
file is found, .ibd will be copied to the destination pointed by .isl.

It works for copy-back, but it doesn't work for move-back, because we
have no control over the order in which files are handled. It can happen
that .isl file already moved when we are handling corresponding .ibd
file.

To fix this, map has been introduced to store association between .ibd
files and their destinations. Now when we stumble upon .ibd or .isl file
we store destination path for .ibd file as been read from .isl. And when
we handle .ibd, we simply lookup the destination from that map.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -58,7 +58,9 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 
 
 /* list of files to sync for --rsync mode */
-std::set<std::string> rsync_list;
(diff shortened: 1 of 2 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/228/

PR #297, opened 2016-12-12
Bug 1642826: incremental backup fails with relative path

gl-sergei:
Fix is to expand not only `target-dir', but also `incremental-basedir',
`incremental-dir', `incremental-lsndir'.

Besides current directory expansion, also expand tilde.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/229/

gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/229/

PR #305, opened 2017-02-03
Do not print WARNING messages into backup stream.

kedazo:
Hello there,

Found a new issue with xtrabackup when server has SSL enabled with our own keys... Now it printed the following warning into the backup stream:

XBSTCK01Extrabackup_infoWARNING: no verification of server certificate will be done. Use --ssl-mode=VERIFY_CA or VERIFY_IDENTITY.

This patch fixes the issue (changed the printout to stderr), but well I do have in [xtrabackup] ssl_mode=DISABLED, so I really have no clue why this warning is even needed here then...

PS.: the original bug report we've created: https://bugs.launchpad.net/percona-xtrabackup/+bug/1647340

Best regard,
David


gl-sergei:
Thank you for your patch! Looks good to me.

PR #310, opened 2017-02-14
Bug1658692: Missing .isl file for general tablespace from backup

VasilyNemkov:
Considering files inside datadir itself for backup, since the general tablespace .isl file is located there.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -3272,6 +3272,12 @@ xb_load_single_table_tablespace(
 	const char *filname,
 	bool is_remote)
 {
+	/* Ignore .isl files on XtraBackup recovery. All tablespaces must be
+	local. */
+	if (is_remote && !srv_backup_mode) {
+		return;
+	}
+
 	/* The name ends in .ibd or .isl;
 	try opening the file */
 	char*	name;
@@ -3455,7 +3461,6 @@ xb_load_single_table_tablespaces(bool (*pred)(const char*, const char*))
 				bool is_remote;
 
 				if (fileinfo.type == OS_FILE_TYPE_DIR) {
-
 					goto next_file_item;
 				}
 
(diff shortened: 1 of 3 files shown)


gl-sergei:
Jenkins param build: http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/135/

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3272,6 +3272,12 @@ xb_load_single_table_tablespace(
 	const char *filname,
 	bool is_remote)
 {
+	/* Ignore .isl files on XtraBackup recovery. All tablespaces must be
+	local. */
+	if (is_remote && !srv_backup_mode) {
```
gl-sergei, line 3277: after adding this condition, should we also remove similar one in `xb_load_single_table_tablespaces`?

storage/innobase/xtrabackup/test/t/bug1658692.sh
```diff
@@ -0,0 +1,66 @@
...
+
+count=`ls $mysql_datadir/*.isl | wc -l`
+vlog  "$count .isl files in datadir, expecting 1"
+test $count -eq 1
+
+vlog "Performing a backup to $target_dir"
+xtrabackup \
+    --backup \
+    --target-dir=$target_dir \
+    --host=127.0.0.1 \
```
gl-sergei: we don't need to specify host and port. xtrabackup is a wrapper function defined `common.sh`, it passing all needed connection parameters

storage/innobase/xtrabackup/test/t/bug1658692.sh
```diff
@@ -0,0 +1,66 @@
...
+    --port=$MYSQLD_PORT
+
+vlog  "Verifying backup directory"
+count=`ls $target_dir | wc -l`
+vlog "$count files in backup, expecting more than 0"
+test $count -gt 0
+
+count=`ls $target_dir/*.isl | wc -l`
+vlog  "$count *.isl files in backup, expecting 1"
```
gl-sergei: this test case looks like a subset of `create_tablespaces.sh`, it only adds this line to check that `.isl` file copied. Can we simply add this check to `create_tablespaces.sh` instead of adding new test case?

gl-sergei (changes requested):

gl-sergei:
Please squash your commits into one and describe what was the bug and what the fix is in commit message.

storage/innobase/xtrabackup/src/xtrabackup.cc
VasilyNemkov: fixed

VasilyNemkov:
Updated, new Jenkins param build: http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/136/

gl-sergei (approved):

PR #314, opened 2017-02-16
Bug 1665239: -h shortcut seems to be not working for xtrabackup

gl-sergei:
xtrabackup -h localhost

aborted with error message

    xtrabackup: Error: unknown argument: 'localhost'

Here is why. -h is a shortcut for --datadir, which is server option
and 'localhost' was recognized as it's argument and handled
correctly.

Then xtrabackup was trying to pick client options and 'localhost'
option was marked as not handled.

Then xtrabackup checked options which were not handled during
previous step and complained that 'localhost' is incorrect.

Fix is to abort only in case when option was not recognized as client
option neither as server option.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/240/

PR #327, opened 2017-03-06
Bug 1669592: MySQL 5.6 fails to start on the incremental backup prepa…

gl-sergei:
…red with 2.4 and non-default `innodb_data_file_path`

This is 2.3 part of the fix. The problem is that command

    xtrabackup --prepare --incremental-dir=... --target-dir=...

creates redo logs inside of the `incremental-dir`.

The directory to look redo logs for is set up inside of the
`innodb_init_param`. When `incremental-dir` is set, we consider that
we are going to apply redo log from incremental backup on top of the
full backup, but InnoDB engine is started twice, first one indeed to
apply incremental log, the second one to create log files.

The fix is to reset `incremental-dir` before the second start, as we
are no longer going to work with the incremental backup.

http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/241/

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -6639,6 +6639,12 @@ next_node:
 
 	if (!xtrabackup_apply_log_only) {
 
+		/* xtrabackup_incremental_dir is used to indicate that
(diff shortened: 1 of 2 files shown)


VasilyNemkov (approved):
Looks good to me

PR #328, opened 2017-03-06
Fix bug 1669592 in 2.4

gl-sergei:
Besides fixing the directory which xtrabackup creates redo logs in,
patch for 2.4 includes the changes in `fil_write_flushed_lsn` to make
it to write the flushed LSN to the page header of the first page of
**each** data file instead of 5.7's default behaviour when flushed LSN
is written to the first page of the first datafile only.

http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/138/


VasilyNemkov (approved):
Looks good

PR #330, opened 2017-03-13
Reset page type for page 6 if needed

dveeden:
https://bugs.launchpad.net/percona-xtrabackup/+bug/1672374


gl-sergei:
We probably need to do the same trick for pages 3 and 7

dveeden:
Could someone merge this or tell me why merging this would be wrong?

gl-sergei:
Fixed by PR #379. Thank you for contribution and bug report!

PR #334, opened 2017-03-18
Bug 1664405: Change default for innodb_checksum_algorithm in 2.4

gl-sergei:
MySQL 5.6 using "INNODB" checksums for datafiles pages which is slow.
MySQL 5.7 using CRC32-C hecksums for datafiles pages which can be
accelerated using Intel's crc32 instructions. xtrabackup defaults to
INNODB. When xtrabackup found checksum mismatch it trying another
algorithm. As a result when xtrabackup copies MySQL 5.7 datafile it
calculates two checkums for each page.

With this patch xtrabackup in backup mode will try to read checksum
algorithm from running server. In prepare mode it will try to guess
checksum algorithm looking at redo log format when it is not found in
`backup-my.cnf`.

### DIFF
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -32,6 +32,7 @@ typedef enum {REDO_LOG_V0, REDO_LOG_V1} redo_log_version_t;
 extern ulint redo_log_version;
 
 extern bool innodb_log_checksum_algorithm_specified;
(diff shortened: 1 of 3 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/143/

VasilyNemkov (commented):
Looks good

PR #335, opened 2017-03-18
Use hardware accelerated CRC32

gl-sergei:
Blueprint:

<https://blueprints.launchpad.net/percona-xtrabackup/+spec/use-hw-crc32>

xtrabackup is using CRC32 in two places:

-   when producing and reading xbstream format
-   when producing and reading xbcrypt format

Since both places are pure xtrabackup, this patch simply replaces
invocation of `crc32` with `crc32_iso3309` without touching server
code.

`crc32_iso3309` invokes hardware optimized CRC32 implementation taken
from `libgrypt` when `PCLMUL` instruction is available. When `PCLMUL`
isn't available, it falls back to software CRC32 from Zlib.

### DIFF
--- a/storage/innobase/xtrabackup/src/CMakeLists.txt
+++ b/storage/innobase/xtrabackup/src/CMakeLists.txt
@@ -18,6 +18,7 @@ INCLUDE(curl)
 INCLUDE(libev)
 
 ADD_SUBDIRECTORY(libarchive)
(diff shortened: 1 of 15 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/251/

VasilyNemkov (approved):
Looks good

PR #340, opened 2017-03-24
Bug 1089681: xtrabackup export - support tables with more than 31 ind…

VasilyNemkov:
…exes

On export, .exp-file generation is explicitly aborted if table has more
than 31 index, this also aborts generation of .cfg file;
The fix is to generate .cfg first, and then try to generate .exp (which
still may fail).


VasilyNemkov:
Jenkins param-build: http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/263/

gl-sergei (approved):
Looks good. Please submit 2.4 version before merging this one

PR #345, opened 2017-03-28
Added option decrypt for xbstream

gl-sergei:
Blueprint <https://blueprints.launchpad.net/percona-xtrabackup/+spec/xbstream-decrypt>

Added options:
   --decrypt=ALGO
   --encrypt-threads=N
   --encrypt-key=KEY
   --encrypt-key-file=KEYFILE

Implemented by adding datasink ds_decrypt and using it when extracting
files ending with .xbcrypt


storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+static void destroy_worker_threads(crypt_thread_ctxt_t *threads, uint n);
+static void *decrypt_worker_thread_func(void *arg);
+
+static const uint encrypt_mode = GCRY_CIPHER_MODE_CTR;
+static uint encrypt_key_len = 0;
+static size_t encrypt_iv_len = 0;
+
+static
+ds_ctxt_t *
+decrypt_init(const char *root)
```
VasilyNemkov: This function looks almost identical to **`encrypt_init()`** from **`ds_encrypt.c`**, I suggest moving common code to the general-purpose function, say `my_bool gcrypt_init(unit algo, const char* encryption_key, const char* encryption_key_file)`.

storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+
+	destroy_worker_threads(crypt_ctxt->threads, crypt_ctxt->nthreads);
+
+	my_free(ctxt->root);
+	my_free(ctxt);
+}
+
+static
+crypt_thread_ctxt_t *
+create_worker_threads(uint n)
```
VasilyNemkov, line 543: There are several issues with this function.
First - it looks almost identical to the **`create_worker_threads()`** from **`ds_encrypt.c`**
Second - memory (`threads`), mutexes, condition variables and cipher handles do leak if N-th thread initialization fails, since there is no cleanup code after **`err:`** label.

storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+
+	return threads;
+
+err:
+	return NULL;
+}
+
+static
+void
+destroy_worker_threads(crypt_thread_ctxt_t *threads, uint n)
```
VasilyNemkov, line 598: This one too, looks almost identical to the `destroy_worker_threads()` from `ds_encrypt.c`.

storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+	checksum = crc32(0, thd->from, thd->from_len);
+	if (checksum != checksum_exp) {
+		msg("%s:%s invalid checksum at offset 0x%llx, "
+		    "expected 0x%lx, actual 0x%lx.\n", my_progname,
+		    __FUNCTION__, thd->offset, checksum_exp, checksum);
+		result = XB_CRYPT_READ_ERROR;
+		goto exit;
+	}
+
+	thd->offset += thd->from_len;
```
VasilyNemkov, line 308: IMO bookkeeping **`thd->offset`** is tedious and error prone, can we just do `thd->offset += *bytes_processed` before returning anything to the invoker?

storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+
+
+	file = (ds_file_t *) my_malloc(sizeof(ds_file_t) +
+				       sizeof(ds_decrypt_file_t),
+				       MYF(MY_FAE|MY_ZEROFILL));
+
+	crypt_file = (ds_decrypt_file_t *) (file + 1);
+
+	/* Remove the .xbcrypt extension from the filename */
+	strcpy(new_name, path);
```
VasilyNemkov: Consider using `strncpy()`

storage/innobase/xtrabackup/src/ds_decrypt.c
```diff
@@ -0,0 +1,845 @@
...
+
+#define CHECK_BUF_SIZE(ptr, size, buf, len) \
+	if (ptr + size - buf > (ssize_t) len) { \
+		result = XB_CRYPT_READ_INCOMPLETE; \
+		goto exit; \
+	}
+
+static
+xb_rcrypt_result_t
+parse_xbcrypt_chunk(crypt_thread_ctxt_t *thd, const uchar *buf, size_t len,
```
VasilyNemkov, line 183: This looks similar to **`xb_crypt_read_chunk()`** from **`xbcrypt_read.c`**

VasilyNemkov (changes requested):
Please update existing code instead of copying it. There are at least 4 good candidates for update/move generic code to shared functions :
- decrypt_init()
- parse_xbcrypt_chunk()
- create_worker_threads()
- destroy_worker_threads()

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: agree with this one

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: Code is written with assumption that path is no longer than FN_REFLEN. However I agree that strncpy is safer choice.

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: Offset is used only in `parse_xbcrypt_chunk` in order to print error message. Incrementing it once simply doesn't work for that purpose.

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: Same pattern with worker threads is common for ds_encrypt, ds_decrypt and ds_compress. Lets leave it as it is for now.

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: Best way is to rewrite xbcrypt to use ds_decrypt. I logged https://blueprints.launchpad.net/percona-xtrabackup/+spec/xbcrypt-ds-decrypt for it.

storage/innobase/xtrabackup/src/ds_decrypt.c
gl-sergei: For separating threadpool from datasink logged https://blueprints.launchpad.net/percona-xtrabackup/+spec/datasink-threadpool

PR #347, opened 2017-03-30
Bug1089681: xtrabackup export - support tables with more than 31 indexes

VasilyNemkov:
On export, .exp-file generation is explicitly aborted if table has more
than 31 index, this also aborts generation of .cfg file;
The fix is to generate .cfg first, and then try to generate .exp (which
still may fail).


gl-sergei (approved):
Looks good to me

PR #365, opened 2017-05-03
Bug 1192834: Crash during apply with index compaction enabled

gl-sergei:
There were several issues with compact backups:

1. Prepare failed on rollback and purge phases because secondary indexes
   are not rebuilt by the time of rollback/purge. Fixed by skipping
   rollback and purge for all secondary indexes.
2. Pages in buffer pool could mistakenly be marked as compacted and redo
   log would not be replayed on such pages which led to corrupt backups
   and crashes. Fixed by clearing `is_compacted' for every page read
   into the buffer pool.
3. Compacted pages which have been freed up and repurposed by server
   were not fully reset during prepare which let to corrupt backups.
4. Index rebuild process could start in parallel with rollback phase
   which led to crashes and corrupt backups. Fix makes sure that index
   rebuild process isn't started until rollback is fully complete.

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -2848,6 +2848,9 @@ files_checked:
 	os_fast_mutex_free(&srv_os_test_mutex);
 
 	if (srv_rebuild_indexes) {
+		while (trx_rollback_or_clean_is_active) {
+			os_thread_sleep(100000);
+		}
 		xb_compact_rebuild_indexes();
 	}
 
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -303,6 +303,7 @@ static char *xtrabackup_debug_sync = NULL;
 
 my_bool xtrabackup_compact = FALSE;
 my_bool xtrabackup_rebuild_indexes = FALSE;
+my_bool xtrabackup_compact_need_expand = FALSE;
 
 my_bool xtrabackup_incremental_force_scan = FALSE;
 
@@ -1962,9 +1963,11 @@ xtrabackup_read_metadata(char *filename)
 	/* Optional fields */
 
 	if (fscanf(fp, "compact = %d\n", &t) == 1) {
-		xtrabackup_compact = (t == 1);
+		xtrabackup_compact = (t != 0);
+		xtrabackup_compact_need_expand = (t == 1);
 	} else {
-		xtrabackup_compact = 0;
+		xtrabackup_compact = FALSE;
+		xtrabackup_compact_need_expand = FALSE;
 	}
 
 	if (fscanf(fp, "recover_binlog_info = %d\n", &t) == 1) {
@@ -1995,7 +1998,11 @@ xtrabackup_print_metadata(char *buf, size_t buf_len)
 		 metadata_from_lsn,
 		 metadata_to_lsn,
 		 metadata_last_lsn,
-		 MY_TEST(xtrabackup_compact == TRUE),
+		 /* compact = 1 means backup is compact, not expanded
+		    compact = 2 means backup is compact, expanded */
+		 xtrabackup_compact ?
+			(xtrabackup_backup ||
+				xtrabackup_compact_need_expand ? 1 : 2) : 0,
 		 MY_TEST(opt_binlog_info == BINLOG_INFO_LOCKLESS));
 }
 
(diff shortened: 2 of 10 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/277/

storage/innobase/xtrabackup/test/t/bug1192834-purge.sh
```diff
@@ -0,0 +1,33 @@
...
+load_dbase_schema sakila
+load_dbase_data sakila
+
+run_cmd $MYSQL $MYSQL_ARGS -e "DELETE FROM payment LIMIT 100" sakila
+
+sleep 2
+
+backup_dir="$topdir/backup"
+
+innobackupex --no-timestamp --compact $backup_dir
```
VasilyNemkov: Do you use `innnobackupex` here intentionally? If yes, please put a comment why, IIRC it is going to be deprecated.

storage/innobase/xtrabackup/test/t/bug1192834-purge.sh
```diff
@@ -0,0 +1,33 @@
...
+# Bug 1192834: Crash during apply with index compaction enabled          #
+##########################################################################
+
+. inc/common.sh
+
+start_server --innodb_file_per_table
+load_dbase_schema sakila
+load_dbase_data sakila
+
+run_cmd $MYSQL $MYSQL_ARGS -e "DELETE FROM payment LIMIT 100" sakila
```
VasilyNemkov: I personally find `mysql` shortcut function from `inc/common.sh` somewhat more convenient and less verbose.

storage/innobase/xtrabackup/test/t/bug1192834-rollback.sh
```diff
@@ -0,0 +1,44 @@
...
+START TRANSACTION;
+DELETE FROM payment;
+SELECT SLEEP(10000);
+EOF
+}
+
+start_uncomitted_transaction &
+job_master=$!
+
+sleep 2
```
VasilyNemkov, line 23: Why do we need to sleep here?

storage/innobase/xtrabackup/test/t/bug1192834-rollback.sh
```diff
@@ -0,0 +1,44 @@
...
+
+backup_dir="$topdir/backup"
+
+innobackupex --no-timestamp --compact $backup_dir
```
VasilyNemkov: Do you use `innnobackupex` here intentionally? If yes, please put a comment why, IIRC it is going to be deprecated.

storage/innobase/xtrabackup/test/t/bug1192834-rollback.sh
```diff
@@ -0,0 +1,44 @@
...
+
+. inc/common.sh
+
+start_server --innodb_file_per_table
+load_dbase_schema sakila
+load_dbase_data sakila
+
+function start_uncomitted_transaction()
+{
+    run_cmd $MYSQL $MYSQL_ARGS sakila <<EOF
```
VasilyNemkov: Please consider using `mysql` function instead.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1995,7 +1998,7 @@ xtrabackup_print_metadata(char *buf, size_t buf_len)
 		 metadata_from_lsn,
 		 metadata_to_lsn,
 		 metadata_last_lsn,
-		 MY_TEST(xtrabackup_compact == TRUE),
+		 xtrabackup_compact ? (xtrabackup_prepare ? 2 : 1) : 0,
```
VasilyNemkov: This implicitly controls the value of `xtrabackup_compact_need_expand` on next read in `xtrabackup_read_metadata()` please add a comment here why.

storage/innobase/srv/srv0start.cc
```diff
@@ -2848,6 +2848,9 @@ innobase_start_or_create_for_mysql(void)
 	os_fast_mutex_free(&srv_os_test_mutex);
 
 	if (srv_rebuild_indexes) {
+		while (trx_rollback_or_clean_is_active) {
```
VasilyNemkov, line 2851: It looks like that we might need proper synchronization here with a mutex or at least by making `trx_rollback_or_clean_is_active` volatile or by playing around with atomic.

VasilyNemkov (changes requested):
Looks Ok, but few minor modifications required.

storage/innobase/xtrabackup/test/t/bug1192834-rollback.sh
gl-sergei: I want DELETE to actually delete some rows before backup starts, I could grep SHOW PROCESSLIST output, but it would be overkill in this case.

storage/innobase/srv/srv0start.cc
gl-sergei: If we were need volatile for extern variables used in loops like this one, InnoDB would probably stop working all together. Compiler will not optimize this loop because the call to `os_thread_sleep` can actually change any global variable as side effect. And there is a syscall inside of it, which makes things even more uncertain for compiler.

VasilyNemkov (approved):
Looks good

PR #366, opened 2017-05-03
Fix FIND_GCRYPT macro.

satori:
There were several mistypes in variable names which prevented macro to
use custom include path.

### DIFF
--- a/cmake/gcrypt.cmake
+++ b/cmake/gcrypt.cmake
@@ -15,11 +15,11 @@
 
 MACRO (FIND_GCRYPT)
 
-  IF (NOT GGRYPT_INCLUDE_PATH)
+  IF (NOT GCRYPT_INCLUDE_PATH)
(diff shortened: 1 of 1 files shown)


gl-sergei:
Hi! Sorry for late reply. Can you please file bug report describing the problem you trying to fix here: https://bugs.launchpad.net/percona-xtrabackup/

gl-sergei:
It is probably the case when grypt is installed to `/usr/local` ?

satori:
I have libgcrypt and libgpg-error headers installed into custom path.
The only way to specify custom headers path in this case is to provide `-DGCRYPT_INCLUDE_PATH=` option.
As this macro has few mistypes in it – it basically doesn't work properly.

I've created JIRA ticket additionally: https://jira.percona.com/browse/PXB-1504

altmannmarcelo:
Hi @satori .

I just checked it is still an issue on 2.4 and 8.0 (currently supported versions). I ported your fix to PR's #1135 and #1136 .

Thanks for your contribution.

PR #368, opened 2017-05-17
Bug 1690823: streaming backups of large databases fail when receiving

gl-sergei:
side runs pxb 2.3.8

xbstream does not close file descriptors when EOF chunk encountered.

`my_hash_delete' has been lost when patch adding parallel extract for
xbstream has been implemented. Fix is to simply put it back.


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/287/

VasilyNemkov (approved):
Looks good

PR #370, opened 2017-05-23
Bug 1651505: GTID position not recorded when --binlog-info=AUTO

gl-sergei:
With introduction of lockless binlog info support in xtrabackup, binary
log coordinates are no longer saved into `xtrabackup_binlog_info' file if
support of safe binlog info by server detected. Instead, binlog
coordinates are restored during prepare from information saved by InnoDB.

It causes trouble in case when GTIDs are in use instead of binlog
coordinates. Since InnoDB doesn't save GTID executed set, xtrabackup has
no place to restore it from.

This patch makes xtrabackup to fallback to `binlog-info=ON' and save
GTID coordinates during the backup when GTID mode is ON.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -407,16 +407,6 @@ get_mysql_vars(MYSQL *connection)
 		have_backup_locks = true;
 	}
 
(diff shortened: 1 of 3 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/290/

VasilyNemkov (approved):
Looks good

PR #372, opened 2017-05-29
Bug 1135431: Check privileges on backup start

VasilyNemkov:
Added an option to force check database user privileges.
Implemented a test to verify that option behaviour.

Bug: https://bugs.launchpad.net/percona-xtrabackup/+bug/1135431
param-build: http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/296/

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -194,6 +194,19 @@ xb_mysql_query(MYSQL *connection, const char *query, bool use_result,
 	return mysql_result;
 }
 
+my_ulonglong
+xb_mysql_numrows(MYSQL *connection, const char *query, bool die_on_error)
+{
+	my_ulonglong rows_count = 0;
+	MYSQL_RES *result = xb_mysql_query(connection, query, true,
+		die_on_error);
+	if (result) {
+		rows_count = mysql_num_rows(result);
+		mysql_free_result(result);
+	}
+	return rows_count;
+}
+
 
 struct mysql_variable {
 	const char *name;
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -352,6 +352,7 @@ my_bool opt_noversioncheck = FALSE;
 my_bool opt_no_backup_locks = FALSE;
 my_bool opt_decompress = FALSE;
 my_bool opt_remove_original = FALSE;
+static my_bool opt_check_privileges = FALSE;
 
 static const char *binlog_info_values[] = {"off", "lockless", "on", "auto",
 					   NullS};
@@ -396,6 +397,9 @@ my_bool opt_ssl_verify_server_cert = FALSE;
 char *opt_server_public_key = NULL;
 #endif
 #endif
+
+static void
+check_all_privileges();
 
 /* Whether xtrabackup_binlog_info should be created on recovery */
 static bool recover_binlog_info;
@@ -611,6 +615,7 @@ enum options_xtrabackup
 
   OPT_XTRA_TABLES_EXCLUDE,
   OPT_XTRA_DATABASES_EXCLUDE,
+  OPT_XTRA_CHECK_PRIVILEGES,
 };
 
 struct my_option xb_client_options[] =
@@ -1012,6 +1017,10 @@ struct my_option xb_client_options[] =
   {"secure-auth", OPT_XB_SECURE_AUTH, "Refuse client connecting to server if it"
     " uses old (pre-4.1.1) protocol.", &opt_secure_auth,
     &opt_secure_auth, 0, GET_BOOL, NO_ARG, 1, 0, 0, 0, 0, 0},
+
+  {"check-privileges", OPT_XTRA_CHECK_PRIVILEGES, "Check database user "
+    "privileges before performing any query.", &opt_check_privileges,
+   &opt_check_privileges, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
 #include "sslopt-longopts.h"
 
@@ -3548,22 +3557,31 @@ xb_register_table(
 }
 
 static
+bool compile_regex(
+	const char* regex_string,
+	const char* error_context,
+	xb_regex_t* compiled_re)
+{
+	char	errbuf[100];
+	int	ret = xb_regcomp(compiled_re, regex_string, REG_EXTENDED);
+	if (ret != 0) {
+		xb_regerror(ret, compiled_re, errbuf, sizeof(errbuf));
+		msg("xtrabackup: error: %s regcomp(%s): %s\n",
+			error_context, regex_string, errbuf);
+		return false;
+	}
+	return true;
+}
+
+static
 void
 xb_add_regex_to_list(
 	const char* regex,  /*!< in: regex */
 	const char* error_context,  /*!< in: context to error message */
 	regex_list_t* list) /*! in: list to put new regex to */
 {
-	char			errbuf[100];
-	int			ret;
-
 	xb_regex_t compiled_regex;
-	ret = xb_regcomp(&compiled_regex, regex, REG_EXTENDED);
-
-	if (ret != 0) {
-		xb_regerror(ret, &compiled_regex, errbuf, sizeof(errbuf));
-		msg("xtrabackup: error: %s regcomp(%s): %s\n",
-			error_context, regex, errbuf);
+	if (!compile_regex(regex, error_context, &compiled_regex)) {
 		exit(EXIT_FAILURE);
 	}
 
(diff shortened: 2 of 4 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1013,6 +1015,10 @@ struct my_option xb_client_options[] =
     " uses old (pre-4.1.1) protocol.", &opt_secure_auth,
     &opt_secure_auth, 0, GET_BOOL, NO_ARG, 1, 0, 0, 0, 0, 0},
 
+  {"check-privileges", OPT_XTRA_CHECK_PRIVILEGES, "Check database user "
+    "privileges before perorming any query.", &opt_check_privileges,
```
gl-sergei: s/perorming/performing/

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6998,6 +7012,206 @@ xb_init()
...
+	PRIVILEGE_ERROR = 2,
+};
+
+static
+int check_privilege(const std::list<std::string> &granted_priv,
+	const char* required,
+	const char* target,
+	int error = PRIVILEGE_ERROR)
+{
+	char buf[64*2 + 1]; //max length of table and db names + separator
```
gl-sergei: Please use NAME_LEN from mysql_com.h

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6998,6 +7012,206 @@ xb_init()
...
+Checks DB user privileges according to the intended actions. */
+static void
+check_all_privileges()
+{
+	if (!mysql_connection) {
+		// Not connected, no queries is going to be executed.
+		return;
+	}
+
+	// Fetch effective privileges.
```
gl-sergei: Please replace C++-style comments with C-style comments

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7375,6 +7589,10 @@ int main(int argc, char **argv)
 	}
 #endif
 
+	if (opt_check_privileges) {
```
gl-sergei: It doesn't make sense to check privileges for any mode except --backup. Probably xtrabackup_backup_func is better place?

storage/innobase/xtrabackup/test/t/bug1135431.sh
```diff
@@ -0,0 +1,52 @@
...
+# Bug 1135431: Check privileges on backup start
+#
+
+user=dummy
+
+# Function to check privileges verification output, pass PXB arguments
+#  as function arguments, and give list of permissions to check via stdin.
+function check_privileges()
+{
+    run_cmd_expect_failure ${XB_BIN} ${XB_ARGS} \
```
gl-sergei: Test case doesn't check successful completion. This option isn't intended to fail in 100% cases, right :)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6998,6 +7012,206 @@ xb_init()
...
+
+static
+int check_privilege(const std::list<std::string> &granted_priv,
+	const char* required,
+	const char* target,
+	int error = PRIVILEGE_ERROR)
+{
+	char buf[64*2 + 1]; //max length of table and db names + separator
+
+	strncpy(buf, target, min(sizeof(buf), strlen(target) + 1));
```
gl-sergei: maybe even better to make `check_privilege` to take db_name and table_name as separate arguments?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3548,22 +3553,31 @@ xb_register_table(
 }
 
 static
+bool compile_regex(
+		const char* regex_string,
```
gl-sergei: InnoDB coding style is to use single tab to indent arguments

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6998,6 +7012,206 @@ xb_init()
...
+}
+
+enum {
+	PRIVILEGE_OK = 0,
+	PRIVILEGE_WARNING = 1,
+	PRIVILEGE_ERROR = 2,
+};
+
+static
+int check_privilege(const std::list<std::string> &granted_priv,
```
gl-sergei: Here is how typical InnoDB function declaration looks like:

```
/******************************************************************//**
Outputs to a file the output of the InnoDB Monitor.
@return FALSE if not all information printed
due to failure to obtain necessary mutex */
UNIV_INTERN
ibool
srv_printf_innodb_monitor(
/*======================*/
	FILE*	file,		/*!< in: output stream */
	ibool	nowait,		/*!< in: whether to wait for the
				lock_sys_t:: mutex */
	ulint*	trx_start_pos,	/*!< out: file position of the start of
				the list of active transactions */
	ulint*	trx_end)	/*!< out: file position of the end of
				the list of active transactions */
```

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
VasilyNemkov: IMO, `xtrabackup_backup_func` is big enough, and I would prefer not to extend it any further.

storage/innobase/xtrabackup/src/xtrabackup.cc
laurynas-biveinis: I'm not sure if that's relevant for you, but new InnoDB code uses a different style, i.e. https://github.com/mysql/mysql-server/blob/8.0/storage/innobase/fsp/fsp0sysspace.cc#L130

storage/innobase/xtrabackup/src/xtrabackup.cc
gl-sergei: Good point. Maybe xb_init then? Right after connect and before reading server variables?

storage/innobase/xtrabackup/src/xtrabackup.cc
gl-sergei: Yeah, comments are different now, but code formatting remains the same, argument list starts on new line. I don't mind to put all arguments in one line, but if each occupies separate line, they should follow InnoDB conventions.

storage/innobase/xtrabackup/src/xtrabackup.cc
VasilyNemkov: It is more reliable to put that check after reading server variables, since some queries are executed only if some specific option is enabled on server. E.g. `have_backup_locks` and `have_flush_engine_logs` control if `FLUSH NO_WRITE_TO_BINLOG ENGINE LOGS` and `LOCK BINLOG FOR BACKUP` are going to be executed.

gl-sergei: This is why I suggested `xb_init`:

```
	if (xtrabackup_backup) {

		if (!opt_noversioncheck) {
			version_check();
		}

		if ((mysql_connection = xb_mysql_connect()) == NULL) {
			return(false);
		}

		if (!get_mysql_vars(mysql_connection)) {
			return(false);
		}
<-------------- somewhere here?
		history_start_time = time(NULL);

	}

```

gl-sergei: ah, yes, right after reading server variables I meant

VasilyNemkov: Ok, that sounds reasonable.

VasilyNemkov:
Considered comments and fixed issues.

VasilyNemkov:
new param-build: http://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/300/

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -194,6 +194,22 @@ xb_mysql_query(MYSQL *connection, const char *query, bool use_result,
...
 
+my_ulonglong
+xb_mysql_numrows(MYSQL *connection, const char *query, bool die_on_error)
+{
+	my_ulonglong rows_count = 0;
+	MYSQL_RES *result = xb_mysql_query(connection, query, true,
+		die_on_error);
+	if (result) {
+		MYSQL_ROW row;
+		while ((row = mysql_fetch_row(result))) {
```
gl-sergei: Can `mysql_num_rows` be used somehow?

storage/innobase/xtrabackup/src/xtrabackup.h
```diff
@@ -163,6 +163,7 @@ extern uint		opt_lock_wait_timeout;
 extern uint		opt_lock_wait_threshold;
 extern uint		opt_debug_sleep_before_unlock;
 extern uint		opt_safe_slave_backup_timeout;
+extern my_bool		opt_check_privileges;
```
gl-sergei: Lets try to make this variable static for `xtrabackup.cc`

gl-sergei (changes requested):
Looks good! Just some minor comments

storage/innobase/xtrabackup/src/xtrabackup.h
VasilyNemkov: Fixed

storage/innobase/xtrabackup/src/backup_mysql.cc
VasilyNemkov: Sure, now using `mysql_num_rows` instead.

storage/innobase/xtrabackup/test/t/bug1135431.sh
VasilyNemkov: Added some success cases.

gl-sergei (approved):

PR #378, opened 2017-06-25
Bug 1671437: Waiting for master thread to be suspended

gl-sergei:
Commit c5aca380511 attempted to fix assertion
`srv_get_active_thread_type() == SRV_NONE', but did it wrong way by
exiting purge thread while there was still some work to do. It caused
transaction not being able to rollback and xtrabackup would loop
indefinitely.

This commit reverts that change and instead makes server don't switch to
`SRV_SHUTDOWN_FLUSH_PHASE' until there is no active transactions.

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -2303,7 +2303,9 @@ loop:
 
 	active_thd = srv_get_active_thread_type();
 
-	if (active_thd != SRV_NONE) {
+	if (active_thd != SRV_NONE
+		|| (srv_fast_shutdown != 2
+			&& trx_rollback_or_clean_is_active)) {
 
 		if (active_thd == SRV_PURGE) {
 			srv_purge_wakeup();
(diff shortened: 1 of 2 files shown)


gl-sergei:
http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/214/

storage/innobase/log/log0log.cc
```diff
@@ -2303,7 +2303,9 @@ logs_empty_and_mark_files_at_shutdown(void)
 
 	active_thd = srv_get_active_thread_type();
 
-	if (active_thd != SRV_NONE) {
+	if (active_thd != SRV_NONE
+		|| (srv_fast_shutdown != 2
+			&& trx_rollback_or_clean_is_active)) {
```
VasilyNemkov, line 2308: This may cause a crash, please take a look at `case SRV_NONE:` below.

VasilyNemkov (changes requested):
There is an issues that has to be addressed.

storage/innobase/log/log0log.cc
gl-sergei: Thank you! Indeed, I missed one more bit for this patch. Should be fine now

gl-sergei:
Fixed. New Jenkins http://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param/220/

VasilyNemkov (approved):
Looks good now.

PR #392, opened 2017-07-16
Bug 1192834: disable compact backups in 2.4

gl-sergei:
Merge test cases for bug 1192834, but disable them. Change `compact.sh` to check that `--compact` option no longer available.

### DIFF
--- a/storage/innobase/xtrabackup/test/t/bug1192834-purge.sh
+++ b/storage/innobase/xtrabackup/test/t/bug1192834-purge.sh
@@ -1,0 +1,33 @@
+##########################################################################
+# Bug 1192834: Crash during apply with index compaction enabled          #
+##########################################################################
+
+. inc/common.sh
+
+skip_test "Enable when bug #1192834 is fixed in 2.4"
+
+start_server --innodb_file_per_table
+load_dbase_schema sakila
+load_dbase_data sakila
+
+mysql -e "DELETE FROM payment LIMIT 100" sakila
+
+backup_dir="$topdir/backup"
+
+xtrabackup --backup --compact --target-dir=$backup_dir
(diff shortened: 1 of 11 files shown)


gl-sergei:
http://jenkins.percona.com/job/percona-xtrabackup-2.4-param-medium/10/

storage/innobase/xtrabackup/test/t/bug1192834-purge.sh
```diff
@@ -0,0 +1,33 @@
+##########################################################################
+# Bug 1192834: Crash during apply with index compaction enabled          #
+##########################################################################
+
+. inc/common.sh
+
+skip_test "Enable when bug #1192834 is fixed in 2.4"
```
VasilyNemkov, line 7: If I get it right, this PR would fix the 1192834. Should we enable this test now?

storage/innobase/xtrabackup/test/t/bug1192834-rollback.sh
```diff
@@ -0,0 +1,46 @@
+##########################################################################
+# Bug 1192834: Crash during apply with index compaction enabled          #
+##########################################################################
+
+. inc/common.sh
+
+skip_test "Enable when bug #1192834 is fixed in 2.4"
```
VasilyNemkov, line 7: Same here, enable test?

VasilyNemkov (commented):
Mostly Ok, but please take a second look at the tests

gl-sergei:
I was going to mark bug 1192834 as "Won't fix" in 2.4. But was to going to keep the tests in case if we change our mind. The bug will still be valid for 2.4 even though one will not be able to take compact backup with 2.4, he can take the backup using 2.3 and crash 2.4

PR #394, opened 2017-07-18
Bug 1704636: xtrabackup 2.4.7 crashes while backing up MariaDB 10.2.x

gl-sergei:
with ftwrl-* options

Unlike MySQL or Percona Server, MariaDB can return NULL as Time in SHOW
PROCESSLIST. Fix is to treat NULL as 0.


gl-sergei:
http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-light/21/

VasilyNemkov (approved):
Looks good

PR #402, opened 2017-07-25
Make chunk size produced by innobackupex/xtrabackup adjustable

VasilyNemkov:
https://blueprints.launchpad.net/percona-xtrabackup/+spec/adjustable-chunk-size

Added option '--datafile_buffer_size_multiplicator' that allows to
increase size of the datafile read buffer.

### DIFF
--- a/storage/innobase/xtrabackup/src/fil_cur.cc
+++ b/storage/innobase/xtrabackup/src/fil_cur.cc
@@ -240,8 +240,9 @@ xb_fil_cur_open(
 	cursor->page_size_shift = page_size_shift;
 	cursor->zip_size = zip_size;
 
+	ut_a(opt_read_buffer_size >= UNIV_PAGE_SIZE);
 	/* Allocate read buffer */
-	cursor->buf_size = XB_FIL_CUR_PAGES * page_size;
+	cursor->buf_size = opt_read_buffer_size;
 	cursor->orig_buf = static_cast<byte *>
 		(ut_malloc(cursor->buf_size + UNIV_PAGE_SIZE));
 	cursor->buf = static_cast<byte *>
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -391,6 +391,7 @@ uint opt_safe_slave_backup_timeout = 0;
 
 const char *opt_history = NULL;
 my_bool opt_decrypt = FALSE;
+uint opt_read_buffer_size = 0;
 
 #if defined(HAVE_OPENSSL)
 my_bool opt_ssl_verify_server_cert = FALSE;
@@ -617,6 +618,7 @@ enum options_xtrabackup
   OPT_XTRA_TABLES_EXCLUDE,
   OPT_XTRA_DATABASES_EXCLUDE,
   OPT_XTRA_CHECK_PRIVILEGES,
+  OPT_XTRA_READ_BUFFER_SIZE,
 };
 
 struct my_option xb_client_options[] =
@@ -1022,6 +1024,15 @@ struct my_option xb_client_options[] =
   {"check-privileges", OPT_XTRA_CHECK_PRIVILEGES, "Check database user "
     "privileges before performing any query.", &opt_check_privileges,
    &opt_check_privileges, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
+
+  {"read_buffer_size",
+   OPT_XTRA_READ_BUFFER_SIZE,
+   "Set datafile read buffer size, given value is scaled up to page size."
+   " Default is 10Mb.",
+   &opt_read_buffer_size,
+   &opt_read_buffer_size,
+   0, GET_UINT, OPT_ARG, 10*1024*1024,
+   UNIV_PAGE_SIZE_MAX, UINT_MAX, 0, UNIV_PAGE_SIZE_MAX, 0},
 
 #include "sslopt-longopts.h"
 
--- a/storage/innobase/xtrabackup/test/t/read_buffer_size.sh
+++ b/storage/innobase/xtrabackup/test/t/read_buffer_size.sh
@@ -1,0 +1,106 @@
+start_server
+
+########################################################################
+# Restore backup from given directory and . If given
+# multiple directories, first one is threated as base, rest as incremental.
+# Synopsis:
+#   restore_from <backup_dir> [<incremental_backup_dir> ...]
+#########################################################################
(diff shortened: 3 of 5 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1023,6 +1025,13 @@ struct my_option xb_client_options[] =
     "privileges before performing any query.", &opt_check_privileges,
    &opt_check_privileges, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
+  {"datafile_buffer_size_multiplicator",
+   OPT_XTRA_DATAFILE_READ_BUFFER_MULTIPLICATOR,
```
gl-sergei: I would prefer to specify read buffer size in bytes. Using multiplier feels confusing. Xtrabackup could divide it by page size. It is also possible to restrict this parameter so that it will be multiple of 16k. Maybe something like `read-buffer-size`?

gl-sergei (changes requested):

VasilyNemkov:
Updated param-build: http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-medium/15/

storage/innobase/xtrabackup/src/xtrabackup.cc
VasilyNemkov: Fixed

VasilyNemkov: Fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1023,6 +1025,14 @@ struct my_option xb_client_options[] =
     "privileges before performing any query.", &opt_check_privileges,
    &opt_check_privileges, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
+  {"read_buffer_size",
+   OPT_XTRA_DATAFILE_READ_BUFFER_MULTIPLICATOR,
+   "Set datafile read buffer size, given value is scaled up to page size."
+   " Use 0 for default, which is 10Mb.",
```
gl-sergei: I think we can just set default value 10M

storage/innobase/xtrabackup/src/fil_cur.cc
```diff
@@ -242,6 +242,19 @@ xb_fil_cur_open(
 
 	/* Allocate read buffer */
 	cursor->buf_size = XB_FIL_CUR_PAGES * page_size;
+	if (opt_read_buffer_size != 0) {
+		ulint buf_size = cursor->buf_size;
+		const char *factors[] = {"b", "Kb", "Mb", NULL};
```
gl-sergei: As far as I remember my_getopt recognise suffixes

gl-sergei: This buffer is for innodb files only. What about other ones? MyISAM files in particular can actually be quite large.

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/fil_cur.cc
VasilyNemkov: Sure, but suffices here are not for parsing option values, but for reporting custom buffer size.

storage/innobase/xtrabackup/src/xtrabackup.cc
VasilyNemkov: The whole point of having default value of `0` was to be able distinguish user set value from default one. In my opinion, having specific value set as default value would complicate things a bit:
* that would create unnecessary dependency of option-specifying code to the engine-stuff
* default value may change in future engine versions (or different engines may have different default values)
* it is hard to tell if value was explicitly set by the user or was that a default one.

storage/innobase/xtrabackup/src/fil_cur.cc
gl-sergei: Well, if it is done, it should be done globally. `innodb_log_file_size`, `use-mem` and many more can be expressed in KiBs MiBs and GiBs... I don't think it improves user experience a lot. It loses in precision however.

storage/innobase/xtrabackup/src/xtrabackup.cc
gl-sergei: One more thing - constant `OPT_XTRA_DATAFILE_READ_BUFFER_MULTIPLICATOR ` should be changed to match option name.
OK. Lets compare. User read this description as "set 0 if you want 10Mb". When user run xtrabackup without arguments, xtrabackup will output default values (and ones from my.cnf) and user will see "read_buffer_size=0". It makes confusion.
Dependency. When you change default value you are going to change help text as well. Isn't it the same level of dependency as change the value itself? Actually I think it is less obvious dependency and one which much easier to miss.
Are there any specific cases when we want to distinguish user-set value and default one? If there are, then we *must* do that in `get_one_option` and not rely on magic value of 0.

VasilyNemkov:
Fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1023,6 +1025,15 @@ struct my_option xb_client_options[] =
     "privileges before performing any query.", &opt_check_privileges,
    &opt_check_privileges, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
+  {"read_buffer_size",
+   OPT_XTRA_READ_BUFFER_SIZE,
+   "Set datafile read buffer size, given value is scaled up to page size."
+   " Default is 10Mb.",
+   &opt_read_buffer_size,
+   &opt_read_buffer_size,
+   0, GET_UINT, OPT_ARG, UNIV_PAGE_SIZE_MAX*640,
```
gl-sergei: UNIV_PAGE_SIZE_MAX is different between 2.3 and 2.4. Please make sure to take it to attention.

storage/innobase/xtrabackup/test/t/read_buffer_size.sh
```diff
@@ -0,0 +1,39 @@
+start_server
+
+load_dbase_schema incremental_sample
+multi_row_insert incremental_sample.test \({1..100},100\)
```
gl-sergei: How much data is it? Is it big enough to test 100M chunk size?

gl-sergei (changes requested):
Please also check incremental backup in the test case. Will it produce chunks of specified size as well?

gl-sergei (approved):
LGTM

PR #405, opened 2017-07-25
Added parallel copyback.

VasilyNemkov:
Blueprint:
[[https://blueprints.launchpad.net/percona-xtrabackup/+spec/add-parallel-copyback-2.3](https://blueprints.launchpad.net/percona-xtrabackup/+spec/add-parallel-copyback-2.3)]
--parallel option now affects --copy-back, setting number of threads used
for copying/moving files.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -903,13 +903,15 @@ backup_file_printf(const char *filename, const char *fmt, ...)
 
 static
 bool
-run_data_threads(datadir_iter_t *it, os_thread_func_t func, uint n)
+run_data_threads(datadir_iter_t *it, os_thread_func_t func, uint n,
+		 const char *thread_description)
 {
 	datadir_thread_ctxt_t	*data_threads;
 	uint			i, count;
 	os_ib_mutex_t		count_mutex;
 	bool			ret;
 
+	ut_a(thread_description);
 	data_threads = (datadir_thread_ctxt_t*)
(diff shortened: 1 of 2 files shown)


VasilyNemkov:
Param-build: http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-medium/8/

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1746,85 +1851,19 @@ copy_back()
 
 	it = datadir_iter_new(".", false);
```
gl-sergei, line 1852: few lines above we copy ibdata1, undo and redo logs. We should mention in blueprint that we don't copy them in parallel.

gl-sergei (approved):
Looks good to me.

gl-sergei:
Please describe this feature in blueprint so that it can be documented.

storage/innobase/xtrabackup/src/backup_copy.cc
VasilyNemkov: This is the same way `xtrabackup_backup_func` works. But sure, let me put that in the blueprint.

gl-sergei: It is not true for backup stage. When backing up, we copy redo logs in separate thread. We also build fli_space dictionary and ibdata1 is just one of the tablespaces, it is copied in parallel with file-per-table tablespaces. Same is true for undo tablesaces. Your implementation is fine for copy-back. I just wanted to explain how backup works.

PR #406, opened 2017-07-28
Add deprecation notice for innobackupex

dveeden:
As a follow up for: https://bugs.launchpad.net/percona-xtrabackup/+bug/1706314

innobackupex is the compatibility layer of xtrabackup to be compatible with the way
the original perl script was called.

From the bug report:
"innobackupex is deprecated we don't add new options to this binary and don't fix bugs in it"

Not that most options are identical, so switching should be easy.

Also note the lie in the manpage: "The innobackupex tool is a Perl script"


hrvojem (approved):
LGTM

gl-sergei:
Thank you for contribution! I'll merge it as it is.

PR #411, opened 2017-08-14
BUG 1590320: Flush binary logs for consistent GTID

abg:
GTIDs are only flushed to the mysql.gtid_executed on binlog rotate or
on a clean shutdown.  During a backup, the contents of this table may be
lagging significantly behind @@global.gtid_executed.  This means that
without flushing binlogs that a restored instance's
@@global.gtid_executed is not guaranteed to contain the correct
coordinates and may require manual steps to fix.

With this change under MySQL 5.7.5+ with gtid replication enabled,
xtrabackup flushes binary logs to ensure up-to-date contents of the
mysql.gtid_executed table.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -97,6 +97,9 @@ MYSQL *mysql_connection;
 /* Whether LOCK BINLOG FOR BACKUP has been issued during backup */
 static bool binlog_locked;
 
+/* Whether FLUSH BINARY LOGS has been issued during backup */
+static bool binlog_flushed;
+
 /* Whether LOCK TABLES FOR BACKUP / FLUSH TABLES WITH READ LOCK has been issued
 during backup */
 static bool tables_locked;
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -1577,6 +1592,9 @@ write_binlog_info(MYSQL *connection)
 	gtid = (gtid_executed != NULL ? gtid_executed : gtid_current_pos);
 
 	if (mariadb_gtid || mysql_gtid) {
+		if (mysql_gtid && mysql_server_version >= 50705) {
+			flush_binlogs_maybe(connection);
```
gl-sergei, line 1596: FLUSH BINARY LOGS potentially can take a while. It is done under either FTWRL or LOCK BINLOG FOR BACKUP which locks all commits, so it potentially can affect server performance. Have you measured this effect somehow?

abg: This is a fair point and can also apply to the Galera/SST case.   Here, the flush is not strictly required but rather a convenience to avoid a case where the `mysql.gtid_executed` table is inconsistent after restore.

As far as performance impact, it is not hard to contrive a case with a large amount of "expired" binlogs that are purged on flush that can take many seconds or even minutes in extreme cases.   Based on that alone, I understand if this PR is rejected.

PR #416, opened 2017-08-25
[LP-1689686] Cannot install percona-xtrabackup-24 with PXC 5.6.35

EvgeniyPatlan:
We are going to install percona-xtrabackup into /opt dir.
In /usr/bin/ we will create links i.e xtrabackup-24 which will
point to the binary in the /opt. Using update-alternatives we will
manage links to the correct binaries:

update-alternatives --display xtrabackup
xtrabackup - status is auto.
 link currently points to /usr/bin/xtrabackup-24
/usr/bin/xtrabackup-23 - priority 100
/usr/bin/xtrabackup-24 - priority 110

Once package percona-xtrabackup is removed link will be automatically
moved to /usr/bin/xtrabackup-23


gl-sergei (approved):
Approved based on discussion on Slack. Thank you!

PR #441, opened 2017-10-11
Bug 1717158: Safe-slave-backups stucks with long query

VasilyNemkov:
Not stopping slave SQL thread until long-running query is over.

Bug: https://bugs.launchpad.net/percona-xtrabackup/2.3/+bug/1717158
Param-builds: 
* http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-light/33/
* http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-medium/24/

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -1043,6 +1043,27 @@ get_open_temp_tables(MYSQL *connection)
 	return(result);
 }
 
+static
+char*
+get_slave_coordinates(MYSQL *connection)
+{
+	char *relay_log_file = NULL;
+	char *exec_log_pos = NULL;
+	char *result = NULL;
+
+	mysql_variable slave_coordinates[] = {
+		{"Relay_Master_Log_File", &relay_log_file},
+		{"Exec_Master_Log_Pos", &exec_log_pos},
+		{NULL, NULL}
+	};
+
+	read_mysql_variables(connection, "SHOW SLAVE STATUS",
+		slave_coordinates, false);
+	ut_a(asprintf(&result, "%s\\%s", relay_log_file, exec_log_pos));
+	free_mysql_variables(slave_coordinates);
+	return result;
+}
+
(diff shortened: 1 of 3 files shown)


gl-sergei:
I have couple of questions:

1. Does it work in GTID mode? Can test case be added?
2. If query takes 10 min to complete and doesn't close temp table and the next query doesn't close temp either, and we run 10 such queries in a row, xtrabackup will wait for 100 min even though it was run with timeout=30min. Can we improve this behaviour in this patch?

Please, be more verbose in commit message.

gl-sergei:
Correction regarding waiting time. Instead of 100 minutes it will be 30mins + time it takes to rollback queries. Which probably is not that bad. Can you please do some experiments to measure actual rollback time for large updates/inserts?

VasilyNemkov:
Added test case for GTID-slave with AUTO_POSITION=1;
Taking 'STOP SLAVE' time into account as part of total timeout.

VasilyNemkov:
Also, added an attempt to handle race condition when stopping slave after requesting  `Slave_open_temp_tables`.

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -1074,47 +1099,86 @@ wait_for_safe_slave(MYSQL *connection)
...
 
 		open_temp_tables = get_open_temp_tables(connection);
 		msg_ts("Slave open temp tables: %d\n", open_temp_tables);
 	}
 
-	/* Restart the slave if it was running at start */
 	if (open_temp_tables == 0) {
-		msg_ts("Slave is safe to backup\n");
+		xb_mysql_query(connection, "STOP SLAVE SQL_THREAD", false);
+		/* We are in a race here, slave might have opened another
```
VasilyNemkov: @gl-sergei could you please take a look, is it worth re-running a check in this case?

VasilyNemkov:
Updated once more: taking into account all time used on stopping slave SQL thread.

gl-sergei (approved):
LGTM

PR #443, opened 2017-10-27
Bug 1727922: Using non-extended RE syntax in order to make test more portable.

VasilyNemkov:
It looks like grep was downgraded at least on some of Jenkins slaves. Using more portable non-extended RE syntax in order to fix the test.

Bug:  https://bugs.launchpad.net/percona-xtrabackup/2.4/+bug/1727922
Jenkins param-build: http://jenkins.percona.com/job/percona-xtrabackup-2.4-param-light/17/


gl-sergei (approved):
Looks good to me. Thank you!

PR #445, opened 2017-10-31
Bug 1705383: xtrabackup 2.4.7 prepare hangs at log_reserve_and_open

gl-sergei:
As analysis shown there was not enough room in log buffer to accomodate
checkpoint information at the end of the crash recovery process. Further
on log buffer size would automatically grow, but not in this specific
place.

Fixed by aligning initial log buffer size with MySQL 5.7 default.


VasilyNemkov (approved):
Looks good, but could you please run a Jenkins param-build before merging? A light build would be enough: http://jenkins.percona.com/job/percona-xtrabackup-2.4-param-light/

PR #446, opened 2017-11-08
Bug 1729241: Build failure on Ubuntu artful

gl-sergei:
Addressed new warnings and errors reported by GCC 7

param build: http://jenkins.percona.com/job/percona-xtrabackup-2.3-param-medium/30/

Test failures are because of missing rsync and some perl packages for version_check


laurynas-biveinis (commented):
Server part LGTM

PR #467, opened 2017-12-09
Bug 1737179[pxb-2.4]: Added check if recv_needed_recovery is FALSE then print binlog pos

ShahriyarR:
Simple fix for:
https://bugs.launchpad.net/percona-xtrabackup/+bug/1737179

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -7790,10 +7790,10 @@ next_node:
 	}
 
 	/* print the binary log position  */
-	if (!recv_needed_recovery) {
-        trx_sys_print_mysql_binlog_offset();
-        msg("\n");
-    }
+//	if (!recv_needed_recovery) {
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7790,8 +7790,10 @@ xtrabackup_prepare_func(void)
 	}
 
 	/* print the binary log position  */
-	trx_sys_print_mysql_binlog_offset();
```
gl-sergei: Should these lines be completely removed instead?

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
ShahriyarR: Commented this line and tested several times.
No issues found. So yes, it can be removed instead.

ShahriyarR: Removed specified line and added test case

PR #469, opened 2017-12-13
Bug 1736119: xtrabackup 2.4.x doesn't compile on MacOS X anymore

gl-sergei:
MySQL bug #25048573

Problem: Some instantiations of std::map have discrepancies between the
value_type of the map and the value_type of the map's allocator. On
FreeBSD 11 this is detected by Clang, and an error is raised at
compilation time.

Fix: Specify the correct value_type for the allocators.


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/54/

PR #471, opened 2017-12-21
kill-long-query-type default value is worng

lipengchong:
The kill-long-query-type option default value  described is all，but actually is QUERY_TYPE_SELECT，so fix it to QUERY_TYPE_ALL.


altmannmarcelo:
Hi @lipengchong , Ideally we should not change the behaviour, instead, we should be updating the help text to state the default is SELECT.

Do you want to proceed with this PR by rebasing it to 2.4 and changing the helper text?

altmannmarcelo:
closing due to innactivity. Fixed at #1159 & #1160

PR #473, opened 2018-01-22
Disable insert buffer during recovery

efirs:
Insert buffer merge operations causing deadlock during recovery between
recovery thread and read IO threads.

The reason is write to redo log during recovery by insert buffer
merge operation, while free space in the log below the margin.

Recovery thread is holding recv_sys->apply_batch_on and waiting for async IOs to
complete and IO threads are waiting for recv_sys->apply_batch_on to ensure
recovery hash applied before freeing space in the log.

Relevant thread stacks:

    4 nanosleep(libpthread.so.0), os_thread_sleep(os0thread.cc:254), *recv_apply_hashed_log_recs*(log0recv.cc:2756),log_checkpoint(log0log.cc:1938),log_margin_checkpoint_age(log0log.cc:341),mtr_t::Command::prepare_write(mtr0mtr.cc:869),mtr_t::Command::execute(mtr0mtr.cc:949),mtr_t::commit(mtr0mtr.cc:576),ibuf_mtr_commit(ibuf0ibuf.ic:60),ibuf_merge_or_delete_for_page(ibuf0ibuf.ic:60),buf_page_io_complete(buf0buf.cc:5894),fil_aio_wait(fil0fil.cc:6128),io_handler_thread(srv0start.cc:309),start_thread(libpthread.so.0),clone(libc.so.6)
    1 nanosleep(libpthread.so.0),os_thread_sleep(os0thread.cc:254),buf_read_recv_pages(buf0rea.cc:850),recv_read_in_area(log0recv.cc:2722),*recv_apply_hashed_log_recs*(log0recv.cc:2722),innobase_start_or_create_for_mysql(srv0start.cc:2217),innodb_init(xtrabackup.cc:2004),xtrabackup_prepare_func(xtrabackup.cc:7585),main(xtrabackup.cc:8704)

The fix is to completely disable insert buffer operations while recovering logs.

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1575,7 +1575,7 @@ log_preflush_pool_modified_pages(
 		and we could not make a new checkpoint on the basis of the
 		info on the buffer pool only. */
 
-		recv_apply_hashed_log_recs(FALSE);
+		recv_apply_hashed_log_recs(TRUE);
 	}
 
 	if (new_oldest == LSN_MAX
@@ -1929,7 +1929,7 @@ log_checkpoint(
 	ut_ad(!srv_read_only_mode);
 
 	if (recv_recovery_is_on()) {
-		recv_apply_hashed_log_recs(FALSE);
+		recv_apply_hashed_log_recs(TRUE);
 	}
 
 #ifndef _WIN32
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1575,7 +1575,7 @@ log_preflush_pool_modified_pages(
 		and we could not make a new checkpoint on the basis of the
 		info on the buffer pool only. */
 
-		recv_apply_hashed_log_recs(TRUE);
+		recv_apply_hashed_log_recs(FALSE);
 	}
 
 	if (new_oldest == LSN_MAX
@@ -1929,7 +1929,7 @@ log_checkpoint(
 	ut_ad(!srv_read_only_mode);
 
 	if (recv_recovery_is_on()) {
-		recv_apply_hashed_log_recs(TRUE);
+		recv_apply_hashed_log_recs(FALSE);
 	}
 
 #ifndef _WIN32
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -2284,7 +2284,7 @@ files_checked:
 			respective file pages, for the last batch of
 			recv_group_scan_log_recs(). */
 
-			recv_apply_hashed_log_recs(TRUE);
+			recv_apply_hashed_log_recs(FALSE);
 			DBUG_PRINT("ib_log", ("apply completed"));
 
 			if (recv_needed_recovery) {
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -2284,7 +2284,9 @@ files_checked:
 			respective file pages, for the last batch of
 			recv_group_scan_log_recs(). */
 
+			log_mutex_enter();
 			recv_apply_hashed_log_recs(FALSE);
+			log_mutex_exit();
 			DBUG_PRINT("ib_log", ("apply completed"));
 
 			if (recv_needed_recovery) {


efirs:
This  can be a fix for https://jira.percona.com/browse/PXB-1704 too.

gl-sergei:
Hi @efirs ,

I think this is the right way to go. It looks like that the actual root cause for the deadlock is that ibuf merge is trying to make extra log record and we don't have enough room in the `xtrabackup_logfiles` for it. However, I think that disabling ibuf merge in the `innobase_start_or_create_for_mysql` will be sufficient. Other two places are actually called from `log_checkpoint`, which is called later from `recv_recovery_from_checkpoint_finish`, by that time we should have enough room for transaction rollbacks and ibuf merges.

ovaistariq:
@gl-sergei yes this seems like the right fix. How do we move forward with it?

gl-sergei:
@efirs , @ovaistariq Does the change in `srv0start.cc` alone resolve the issue? If yes, I'll ask @efirs to revert changes to the `log0log.cc` and will merge this PR. If not, I would like to know why...

efirs:
> @efirs , @ovaistariq Does the change in `srv0start.cc` alone resolve the issue? If yes, I'll ask @efirs to revert changes to the `log0log.cc` and will merge this PR. If not, I would like to know why...

Actually, the fix requires more thoughts. Whenever we pass FALSE to `recv_apply_hashed_log_recs`, it expects caller to hold `log_mutex`, which is not the case in `innobase_start_or_create_for_mysql`.

Disabling of insert buffer merges in `log0log.cc` is required, I think, because `recv_apply_hashed_log_recs` releases `log_mutex`, during buffer pool invalidation, while `recv_recovery_is_on()` is still true.

gl-sergei:
You are right regarding the the `log_mutex_own` requirement.

This requirement comes from the fact that `recv_apply_hashed_log_recs(FALSE)` is only called from `recv_group_scan_log_recs` which is holding the `log_mutex` and `recv_apply_hashed_log_recs` needs to unlock it in order to let other threads to make progress.

During crash recovery, only the last batch is ran with `recv_apply_hashed_log_recs(TRUE)`, the one in the `innobase_start_or_create_for_mysql` and according to the provided stack traces this last one is causing the hang.

Simple
```
			log_mutex_enter();
			recv_apply_hashed_log_recs(FALSE);
			log_mutex_exit();
```
will resolve the issue with `log_mutex`. Alternatively, we can introduce  additional boolean argument for `recv_apply_hashed_log_recs` to indicate whether the `log_mutex` is locked.

I assume that here we are waiting for IO completion: 

```
rtr_page_split_and_insert
buf_read_recv_pages
recv_apply_hashed_log_recs (TRUE)
innobase_start_or_create_for_mysql
```

IO completion threads are calling `ibuf_merge_or_delete_for_page`, because `recv_no_ibuf_operations == FALSE`. There are some IBUF changes to merge and `ibuf_merge_or_delete_for_page` started mtr transaction and is trying to commit.

```
recv_apply_hashed_log_recs ()
log_checkpoint ()  <---- this might be unusual in running server (see below)
log_margin_checkpoint_age ()
...
mtr_t::commit ()
```

There is a comment in `log_margin_checkpoint_age`:

> Our margin check should ensure that we never reach this condition.
> Try to do checkpoint once. We cannot keep waiting here as it might
> result in hang in case the current mtr has latch on oldest lsn

and xtrabackup actually reached this point, it is trying to make a checkpoint to free up some space in
the redo log and because crash recovery is ON, it calls `recv_apply_hashed_log_recs`:

```
log_checkpoint(
	bool	sync,
	bool	write_always)
{
	lsn_t	oldest_lsn;

	ut_ad(!srv_read_only_mode);

	if (recv_recovery_is_on()) {
		recv_apply_hashed_log_recs(TRUE);
	}
```

This is a deadlock, because we have two `recv_apply_hashed_log_recs` invoked recursively. Recursive call of `recv_apply_hashed_log_records` deadlocks because of:

```
loop:
	mutex_enter(&(recv_sys->mutex));

	if (recv_sys->apply_batch_on) {

		mutex_exit(&(recv_sys->mutex));

		os_thread_sleep(500000);

		goto loop;
	}

...

	recv_sys->apply_batch_on = TRUE;

... do the work ... ... if we call "recv_apply_hashed_log_records" now, it will deadlock ...

	recv_sys->apply_batch_on = FALSE;
```

My theory is that if we disallow ibuf (i.e. pass FALSE to `recv_apply_hashed_log_recs`) in the
`innobase_start_or_create_for_mysql`, we will not need to make checkpoints and actually pass safely
to the `recv_recovery_from_checkpoint_finish`, where IBUF merge will be OK. Thus, changes in `log0log.cc` are not necessary.

efirs:
Thank you for detailed analysis. 
I think grabbing and releasing log_mutex can be good quick workaround. I've added it to the patch.

gl-sergei:
I created new PR https://github.com/percona/percona-xtrabackup/pull/577 which supersedes this one, it also takes care of the crash in the `xtrabackup --stats`.

PR #474, opened 2018-02-13
Bug PXB-1497: Xtrabackup not returning Error when there is failure

gl-sergei:
XtraBackup skipping a tablespace when it's first page is corrupted, it
was done because newly created tablespace can have invalid first page
(not flushed yet to disk).

Instead of blindly skipping tablespace when it's first page is
corrupted, check whether it is blank and skip the tablespace only in
this case.


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/70/

PR #476, opened 2018-03-09
PXB-1467: XtraBackup stuck in prepare stage waiting for pending reads

gl-sergei:
By code review following was discovered:

1. xtrabackup set maximum number of pending reads 4 times lower than it
   supposed to due to regression
2. there was a deadlock in `recv_apply_hashed_log_recs' and
   `buf_page_io_complete'.

Here is an example deadlock scenario:

Thread 1 in `recv_apply_hashed_log_recs' is waiting when
`buf_pool->n_pend_reads' become not too high to make a progress. It is
`apply_batch_on=TRUE' and will change it to be `FALSE' once apply batch
completed. Note that `buf_pool->n_pend_reads' is already high.

Now, one of the pending reads completes and `buf_page_io_complete'
invoked. It should decrement `buf_pool->n_pend_reads' and let current
apply batch to make progress.

But before decrementing `buf_pool->n_pend_reads', `buf_page_io_complete'
invoked `ibuf_merge_or_delete_for_page' which in turn triggered one more
`recv_apply_hashed_log_recs'. This new `recv_apply_hashed_log_recs'
cannot make progress because `apply_batch_on' is `TRUE', it is waiting
for thread 1. We are in the deadlock now.

Lets imagine that all IO handlers (`buf_page_io_complete') stuck in the
`recv_apply_hashed_log_recs', here is what we see in this case.

Proposed fix is to decrement `buf_pool->n_pend_reads' before invoking
`ibuf_merge_or_delete_for_page'.

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -5376,6 +5376,7 @@ buf_page_init_for_read(
 		mutex_exit(&buf_pool->zip_mutex);
 	}
 
+	ut_ad(mutex_own(&buf_pool->mutex));
 	buf_pool->n_pend_reads++;
 func_exit:
 	buf_pool_mutex_exit(buf_pool);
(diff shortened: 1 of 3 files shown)


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/75/

PR #485, opened 2018-04-16
PXB-1525: Support keyring_vault plugin

gl-sergei:
This patch includes following:

1. `keyring_vault' and `keyring_file' are built as plugins, new option
   `--xtrabackup-plugin-dir' should be used to specify plugins location.
   PS plugins and PXB plugins are built with different options and are
   not binary compatible.
2. When doing a backup, xtrabackup will fetch keyring plugin setup from
   live server and load appropriate plugin.
3. There are three backup/prepare options:
   - To dump tablespace keys encrypted with transition key, transition
     key specified as command line option. Transition key then must be
     passed to prepare and to copy-back (if needed).
   - To dump tablespace keys encrypted with transition key, transition
     key is automatically generated and stored in the keyring. Keyring
     then must be accessible during prepare and copy-back (if needed).
   - Don't dump tablespace keys at all. Keyring then must be accessible
     during prepare and copy-back (if needed). Keys must not be purged.
4. copy-back has an option to generate new master key and put it into
   keyring. Tablespace headers will then be encrypted with this new
   master key. Suitable for spinning up new servers or switching to
   different keyring plugin.


berlincount:
It seems like this change breaks support for non-GTID backups (or rather, restores):

`[error] unknown variable 'server_uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'`

😢

gl-sergei:
Can you please clarify when do you see this message? Or better submit bug report - http://jira.percona.com/projects/PXB

berlincount:
We're getting it on MySQL server startup while doing a backup verification / restore against percona server 5.7.21-21 after the log apply was done. This takes a few hours every time, so I'm still collecting data.

gl-sergei:
@berlincount I think I have an idea. Do you start MySQL using the `backup-my.cnf` from the backup dir? If so, it isn't supposed to work. `backup-my.cnf` is needed for `xtrabackup --prepare` and it contains only bare minimum of InnoDB settings and it can contain options unknown to MySQL.

berlincount:
@gl-sergei yes, that might be related. our verification code constructs the final configuration file used from both a template and this file .. and I guess it just didn't contain unknown options before 😞

PR #486, opened 2018-04-18
PXB-1526: Test kill_long_selects.sh failing with MySQL 5.7.21

gl-sergei:
At some point behaviour of MySQL 5.7 has changed and now it requires
--secure-file-priv to explicitly allow using the directiry as a target
for SELECT INTO OUTFILE


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/102/

PR #492, opened 2018-05-14
PXB-202: Do not encrypt xtrabackup_checkpoints

dutow:


gl-sergei:
The changes look good. I think that instead of adding two tests based on `ib_incremental_common` which itself is quite heavy, maybe we could check whether the metadata file is compressed/encrypted in the `xb_encrypt.sh` / `xb_compess_encrypt.sh` ?

dutow:
Jenkins: https://jenkins.percona.com/job/percona-xtrabackup-2.4-param-medium/110/

gl-sergei (approved):
Looks good to me. Thank you!

PR #493, opened 2018-05-17
PXB-1540: XB removes and recreate keyring file of 0 size

gl-sergei:
This bug was introduced with the keyring_file plugin update in the
xtrabackup source tree. Keyring plugin removes empty keyring file. It
doesn't corrupt existing non-empty keyring file. The issue is that
server detects file modification time change and refuses to create
encrypted tables afterward.


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/111/

dutow (approved):

PR #494, opened 2018-05-25
PXB-1943: Add warnings for unknown client arguments

dutow:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/116/

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -8624,16 +8624,17 @@ handle_options(int argc, char **argv, int *argc_client, char ***argv_client,
 	for (int i = 0 ; i < *argc_client ; i++) {
 		const char * const opt = (*argv_client)[i];
 
+		bool server_option = true;
+
+		for (int j = 0; j < *argc_server; j++) {
+			if (opt == (*argv_server)[j]) {
+				server_option = false;
+				break;
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8599,23 +8599,37 @@ handle_options(int argc, char **argv, int *argc_client, char ***argv_client,
...
 			if (!server_option) {
 				msg("xtrabackup: Error:"
 				    " unknown argument: '%s'\n", opt);
 				exit(EXIT_FAILURE);
 			}
 		}
+
+			bool client_option = true;
+			for (unsigned j = 0; j < xb_client_options_count; j++) {
+				if (strcmp(opt, (xb_client_options)[j].name)) {
```
gl-sergei: Same questions, I would like to see example output and the test case. Also, can we abort with error if incorrect argument has been provided via command line?

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
dutow: For the error message: the ticket asked for a warning, to avoid issues if we do not recognize a valid command line option. I could change it to an error if that's better, but in that case, we probably need a more detailed testing for it to avoid any issues.

For the test case: I'll rebase this on #495 after that's approved, to use the same testcase.

PR #495, opened 2018-05-25
PXB-1494: Print out the used arguments

dutow:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/117/

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -335,6 +335,7 @@ static longlong	innobase_log_file_size_save;
 /* String buffer used by --print-param to accumulate server options as they are
 parsed from the defaults file */
 static std::ostringstream print_param_str;
+static std::ostringstream param_str;
 
 /* Set of specified parameters */
 std::set<std::string> param_set;
@@ -1360,9 +1361,27 @@ check_if_param_set(const char *param)
 
 my_bool
 xb_get_one_option(int optid,
-		  const struct my_option *opt __attribute__((unused)),
+		  const struct my_option *opt,
 		  char *argument)
 {
+  static const char* hide_value[]=
+    { "password", "encrypt-key", "transition-key" };
+
+  param_str << "--" << opt->name;
+  if (argument) {
+    bool param_handled = false;
+    for (unsigned i=0; i<sizeof(hide_value)/sizeof(char*); ++i) {
+      if (strcmp(opt->name, hide_value[i]) == 0) {
+        param_handled = true;
+        param_str << "=*";
+        break;
+      }
+    }
+    if(!param_handled) {
+      param_str << "=" << argument;
+    }
+  }
+  param_str << " ";
   switch(optid) {
   case 'h':
     strmake(mysql_real_data_home,argument, FN_REFLEN - 1);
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8622,6 +8622,24 @@ handle_options(int argc, char **argv, int *argc_client, char ***argv_client,
 		opt_password = get_tty_password(NullS);
 	}
 
+	msg("xtrabackup: Provided arguments:");
```
gl-sergei: Would be nice to add basic test case for this feature. Can you please also provide example output?
I have a feeling that this feature actually won't print used arguments, but rather it will print unused arguments... There is another feature -- `--print-param`, maybe we could use the same idea here? Also, do we need this for 2.3 as well?

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
dutow: * The jira ticket only mentioned 2.4 - but I can open a PR for 2.3 too
* It prints out every argument, even the program name, and then every argument it detects as a client or server argument - for those, it even prints out the arguments it added automatically.
* I added a testcase

An example output:
```
./xtrabackup --backup --target-dir=/backups/ --some-arg --port 4000
xtrabackup: Provided arguments: ./xtrabackup --backup --target-dir=/backups/ --some-arg --port 4000
xtrabackup: Client arguments: --some-arg
xtrabackup: Server arguments: --user=mysql --pid-file=/var/run/mysqld/mysqld.pid --socket=/var/run/mysqld/mysqld.sock --port=3306 --basedir=/usr --lc-messages-dir=/usr/share/mysql --skip-external-locking --bind-address=127.0.0.1 --key_buffer_size=16M --max_allowed_packet=16M --thread_stack=192K --thread_cache_size=8 --myisam-recover-options=BACKUP --max_connections=10000 --query_cache_limit=1M --query_cache_size=16M --log_error=/var/log/mysql/error.log --expire_logs_days=10 --max_binlog_size=100M --backup --target-dir=/backups/ --some-arg --port 4000
```

Edit: changed the example to include a server agument too

gl-sergei:
This is what I meant actually by "unused":
```
--pid-file=/var/run/mysqld/mysqld.pid
--basedir=/usr
--lc-messages-dir=/usr/share/mysql
--skip-external-locking
--bind-address=127.0.0.1
--key_buffer_size=16M
```
etc were found in `my.cnf` but were not used. `handle_options` changes `argv_client` and `argv_server` leaving only those which did not match with any known ones. I think this patch doesn't do what was asked for.

dutow:
PR updated, jenkins: https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/121/

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1480,9 +1481,18 @@ check_if_param_set(const char *param)
 
 my_bool
 xb_get_one_option(int optid,
-		  const struct my_option *opt __attribute__((unused)),
+		  const struct my_option *opt,
 		  char *argument)
 {
+  param_str << "--" << opt->name;
+  if(argument) {
+    if(strcmp(opt->name, "password")==0) {
```
gl-sergei: Please also consider hiding `encrypt-key` and `transition-key` option values. With that fixed, this PR looks good.

gl-sergei (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
gl-sergei: ah... also formatting looks a bit odd. There should be spaces around `==` and between `if` and `(`...

dutow:
Updated, new jenkins: https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/122/

gl-sergei (approved):
LGTM

PR #496, opened 2018-06-11
PXB-1544: xtrabackup --copy-back must take plugin-dir setting from my…

gl-sergei:
….cnf

Change copy-back to make it to look into my.cnf for plugins.

### DIFF
--- a/storage/innobase/xtrabackup/src/keyring_plugins.cc
+++ b/storage/innobase/xtrabackup/src/keyring_plugins.cc
@@ -46,6 +46,30 @@ const char *XTRABACKUP_KEYS_FILE = "xtrabackup_keys";
 const char *XTRABACKUP_KEYS_MAGIC = "KEYSV01";
 const size_t XTRABACKUP_KEYS_MAGIC_SIZE = 7;
 
+/** Load plugins and keep argc and argv untouched */
+static void
+init_plugins(int argc, char **argv)
+{
+	int t_argc = argc;
+	char **t_argv = new char *[t_argc + 1];
+
+	memset(t_argv, 0, sizeof(char *) * (t_argc + 1));
+	memcpy(t_argv, argv, sizeof(char *) * t_argc);
+
+	mysql_optional_plugins[0] = 0;
+	mysql_mandatory_plugins[0] = 0;
+
+	if (opt_xtra_plugin_dir != NULL) {
+		strncpy(opt_plugin_dir, opt_xtra_plugin_dir, FN_REFLEN);
+	} else {
+		strcpy(opt_plugin_dir, PLUGINDIR);
+	}
+
+	plugin_init(&t_argc, t_argv, PLUGIN_INIT_SKIP_PLUGIN_TABLE);
+
+	delete [] t_argv;
+}
+
 /** Fetch tablespace key from "xtrabackup_keys".
 @param[in]	space_id	tablespace id
 @param[out]	key		fetched tablespace key
@@ -203,16 +227,7 @@ xb_keyring_init_for_backup(MYSQL *connection)
 		t_argv[i] = (char *) keyring_plugin_args[i - 1].c_str();
 	}
 
-	mysql_optional_plugins[0] = 0;
-	mysql_mandatory_plugins[0] = 0;
-
-	if (opt_xtra_plugin_dir != NULL) {
-		strcpy(opt_plugin_dir, opt_xtra_plugin_dir);
-	} else {
-		strcpy(opt_plugin_dir, PLUGINDIR);
-	}
-
-	plugin_init(&t_argc, t_argv, PLUGIN_INIT_SKIP_PLUGIN_TABLE);
+	init_plugins(t_argc, t_argv);
 
 	delete [] t_argv;
 
@@ -225,7 +240,7 @@ get_one_option(int optid MY_ATTRIBUTE((unused)),
 	       const struct my_option *opt MY_ATTRIBUTE((unused)),
 	       char *argument MY_ATTRIBUTE((unused)))
 {
-	return 0;
+	return(FALSE);
 }
 
 /** Initialize keyring plugin for stats mode. Configuration is read from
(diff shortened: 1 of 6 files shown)


gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/125/

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -46,6 +46,30 @@ const char *XTRABACKUP_KEYS_FILE = "xtrabackup_keys";
...
 
+/** Load plugins and keep argc and argv untouched */
+static void
+init_plugins(int argc, char **argv)
+{
+	int t_argc = argc;
+	char **t_argv = new char *[t_argc + 1];
+
+	memset(t_argv, 0, sizeof(char *) * (t_argc + 1));
+	memcpy(t_argv, argv, sizeof(char *) * t_argc);
```
dutow, line 57: Shouldn't just `memcpy(t_argv, argv, sizeof(char*)*(t_argc+1));` be enough? I think argv is required to end with a null pointer

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -46,6 +46,30 @@ const char *XTRABACKUP_KEYS_FILE = "xtrabackup_keys";
...
+
+	mysql_optional_plugins[0] = 0;
+	mysql_mandatory_plugins[0] = 0;
+
+	if (opt_xtra_plugin_dir != NULL) {
+		strcpy(opt_plugin_dir, opt_xtra_plugin_dir);
```
dutow: While this code is old, just moved, I don't see why we know that the buffer `opt_plugin_dir` points to is enough for these strings - shouldn't this be `strdup` / freeing the current buffer?

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -225,7 +240,7 @@ get_one_option(int optid MY_ATTRIBUTE((unused)),
 	       const struct my_option *opt MY_ATTRIBUTE((unused)),
 	       char *argument MY_ATTRIBUTE((unused)))
 {
-	return 0;
+	return FALSE;
```
dutow: `return(FALSE);`? (Or maybe we could change the my_bool to bool too?)

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -328,28 +402,53 @@ xb_keyring_init_for_prepare(int argc, char **argv)
...
+		{"plugin-load-add", 3, "", 0, 0, 0, GET_STR, REQUIRED_ARG,
+		 0, 0, 0, 0, 0, 0},
+		{0, 0, 0, 0, 0, 0, GET_NO_ARG, NO_ARG, 0, 0, 0, 0, 0, 0}
+	};
 
-	plugin_init(&t_argc, t_argv, PLUGIN_INIT_SKIP_PLUGIN_TABLE);
+	char **old_t_argv = t_argv;
+	if (handle_options(&t_argc, &t_argv, plugin_load_options,
+			   get_plugin_load_option)) {
+		return(false);
```
dutow: Shouldn't we delete `old_t_argv` here to be consistent?

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -236,24 +251,69 @@ argc and argv.
...
 	my_option keyring_options[] = {
+		{"server-uuid", 0, "", &uuid, &uuid, 0, GET_STR,
+		 REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 		{"plugin-load", 0, "", &plugin_load, &plugin_load, 0, GET_STR,
 		 REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 		{0, 0, 0, 0, 0, 0, GET_NO_ARG, NO_ARG, 0, 0, 0, 0, 0, 0}
 	};
 
-	if (handle_options(&argc, &argv, keyring_options, get_one_option)) {
+	char *exename = (char *) "xtrabackup";
```
dutow: Does it really have to be a non-const string? This is definitely at least a warning, and dangerous - maybe create a modifiable instance with `strdup`?

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -264,26 +324,43 @@ xb_keyring_init_for_stats(int argc, char **argv)
...
+	       const struct my_option *opt MY_ATTRIBUTE((unused)),
+	       char *argument)
+{
+	std::string token;
+	std::istringstream token_stream(argument);
+	while (std::getline(token_stream, token, ';'))
+	{
+		plugin_load_list.push_back(token);
+	}
+	return FALSE;
```
dutow: `return(FALSE)`?

storage/innobase/xtrabackup/src/keyring_plugins.cc
gl-sergei: It might be enough, I have not check. argv comes here after couple modifications, so I tried to play safe.

storage/innobase/xtrabackup/src/keyring_plugins.cc
gl-sergei: it is a buffer of FN_REFLEN size, will change to strncpy

storage/innobase/xtrabackup/src/keyring_plugins.cc
gl-sergei: I guess we should :)

storage/innobase/xtrabackup/src/keyring_plugins.cc
gl-sergei: `handle_options` should not really modify argv[0]

gl-sergei:
repushed

gl-sergei:
https://jenkins.percona.com/view/PXB%202.4/job/percona-xtrabackup-2.4-param-medium/129/

dutow (approved):

PR #497, opened 2018-06-12
PXB-1542: xbstream exit with error code 0 when it failed to create on…

gl-sergei:
…e or more target files

Make xbstream to exit with non-zero error code when it fails to create
target files.


gl-sergei:
https://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param-light/43/

dutow (approved):

PR #508, opened 2018-07-17
Fix DOC-739 documenting keyring_vault support in PXB

fiowro:


storage/innobase/xtrabackup/doc/source/advanced/encrypted_innodb_tablespace_backups.rst
```diff
@@ -182,3 +185,146 @@ Preparing the second incremental backup is a similar process: apply the deltas t
 
 Backup is now prepared and can be restored with :option:`xtrabackup --copy-back` option. In case the keyring has been rotated you'll need to restore the keyring which was used to take and prepare the backup.
 
+Backups with ``keyring_vault`` plugin
```
gl-sergei: This entire section really work for server with both plugins, i.e. it is perfectly fine to create a backup with command
`xtrabackup --backup --user=root -p --target-dir=/data/backup --transition-key=MySecetKey` when server is using `keyring_file` plugin. What really has changed is
1. Now there is more than one plugin, this part is clear from the "Two keyring plugins supported by |xtrabackup| are ``keyring_file`` and ``keyring_vault``" and is fine by me
2. Previously, the only supported case was the one described below in "Creating Backup", i.e. xtrabackup required access to the keyring during the backup and prepare, and it was on the user to make sure that keyring file is accessible (now that we added keyring_vault, it is on user to make sure vault server is accessible). The cases described in "backups with passphrase" and "with stored passphrase" also apply to keyring file.

Instead of sections
1. keyring file
2. keyrig vault
  - Creating Backup
  - backups with passphrase
  - backups with stored passphrase

There should be sections:
1. Creating Backup
2. backups with passphrase
3. backups with stored passphrase

With some specifics explained in each regarding how to deal with different plugins. Like when do we need to copy keyring file and when do we need to make sure that xtrabackup is able to connect to keyring vault server.

storage/innobase/xtrabackup/doc/source/advanced/encrypted_innodb_tablespace_backups.rst
```diff
@@ -63,13 +69,68 @@ After |xtrabackup| is finished preparing the backup you should see the following
...
+
+Preparing the Backup
+********************
+
+In order to prepare the backup |xtrabackup| will need an access to the keyring.
+Since |xtrabackup| doesn't talk to MySQL server and doesn't read default
+``my.cnf`` configuration file during prepare, user will need to specify keyring
+settings via the command line::
+
+.. code-block:: bash
```
gl-sergei: Formatting is broken as I see in the test page

gl-sergei (approved):
LGTM. Thank you! Please fix formatting of the code block and then it is good to merge.

PR #509, opened 2018-07-25
PXB-1569: --history does not work when autocommit is disabled

gl-sergei:
Fix is to enable session autocommit when connected.


gl-sergei:
https://jenkins.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param-medium/51/

PR #512, opened 2018-08-15
Fixed pxb-1614 by updating description of version-check option

borys-belinsky:


storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
```diff
@@ -270,4 +270,15 @@ Options
 
 .. option:: --version-check
```
gl-sergei: Is there PR for 2.3? The option name is `--no-version-check` and it is OFF by default, it means that data is sent by default. xtrabackup has the same option, please also add it to xtrabackup option reference.

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
```diff
@@ -270,4 +270,15 @@ Options
...
 
-   When this option is specified, innobackupex will perform a version check against the server on the backup stage after creating a server connection.
+   When this option is specified, innobackupex will perform a version check
```
gl-sergei: When `--no-version-check` is specified, innobackupex/xtrabackup will **skip** a version check. When it is not specified (default) innobackupex/xtrabackup will perform version check.

gl-sergei (changes requested):

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
borys-belinsky: Do I understand you correctly that 2.3 should also be updated?

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
dbadapt: does innobackupex/xtrabackup do the version check with every execution?  even with --help  ?   Or only before backup, restore, and prepare?

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
```diff
@@ -270,4 +270,15 @@ Options
...
+   against the server on the backup stage after creating a server
+   connection. |innobackupex| sends the following information to the server:
+
+   - MySQL flavour and version
+   - Operation system name
```
dbadapt: should be "Operating system"

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
borys-belinsky: Also, what happens if the user specifies both no-version-check and version-check?

gl-sergei: there is no `version-check`, just `no-version-check`

gl-sergei: 2.3 behaviour is the same, it would be logical to update both

storage/innobase/xtrabackup/doc/source/innobackupex/innobackupex_option_reference.rst
gl-sergei: Only before backup. Not before prepare, not for other modes.

gl-sergei:
Is it ready for review? Please update xtrabackup option reference too. As per my comment "xtrabackup has the same option, please also add it to xtrabackup option reference."

gl-sergei (approved):
LGTM. Thanks!

PR #521, opened 2018-08-29
PXB-1637: Readding O_DIRECT messages

dutow:


gl-sergei (approved):

gl-sergei:
On second thought, why not just print out `innodb_flush_method_names[innodb_flush_method]` ?

dutow:
It's based on the 2.4 implementation, which only printed out if some kind of O_DIRECT was used - if you think its better, I could change it to always print the correct name.

gl-sergei:
I don't have strong opinion on that. Lets keep previous behaviour. This PR is good to merge

dutow:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/9/

PR #522, opened 2018-08-30
PXB-1634: CREATE TABLE fails with DUPLICATE KEY error

gl-sergei:
CREATE TABLE reported DUPLICATE KEY error after a database restore. The
error was easy to reproduce with release builds of xtrabackup, but not
with debug builds.

Starting with MySQL 8 InnoDB does not use SELECT MAX(ai_col) to
initialize auto increment value on server restart. Instead, it relies on
two sources:

- mysql.dynamic_table_metadata table contents are used after normal
  server shutdown
- MLOG_TABLE_DYNAMIC_META redo log records are used to update counters
  which haven't been flushed yet

It turned out that debug builds of xtrabackup backed up the data a bit
slower than release builds and InnoDB had time to flush auto increment
counters to dynamic_table_metadata table. During prepare, xtrabackup was
able to restore this table from usual INSERT and UPDATE log records.

In the case when InnoDB had no time to flush auto-increment data,
dynamic_table_metadata contained outdated counters after prepare, which
caused DUPLICATE KEY errors.

Fix is to make sure that xtrabackup is flushing auto increment counters
it has read from the redo log during prepare.

### DIFF
--- a/storage/innobase/include/dict0mem.h
+++ b/storage/innobase/include/dict0mem.h
@@ -1941,13 +1941,13 @@ detect this and will eventually quit sooner. */
     return (s);
   }
 
-  int field_number(const char* field_name) const {
-    for(int i = 0; i < n_def; ++i) {
-      if(strcmp(field_name, get_col_name(i)) == 0) {
+  int field_number(const char *field_name) const {
+    for (int i = 0; i < n_def; ++i) {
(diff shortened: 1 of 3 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/10/

dutow (approved):

PR #523, opened 2018-08-30
PXB-1639: Partition tests failures

gl-sergei:
Don't attempt to create mysql directory at destination when restoring
partial backup if directory already exists.


dutow:
Uh, why didn't this cause errors for me? Nice catch

dutow (approved):

PR #537, opened 2018-09-18
PXB-1627: Support obtaining binary log coordinates from performance_s…

gl-sergei:
…chema.log_status

Implementation details:

At the beginning of the backup PXB detecting is there any MyISAM
tables to backup and whether the server is a GTID slave with auto
positioning.

Privileges needed to query `p_s.log_status` table are checked if
privilege check was requested.

Log copying thread has changed to stop at specific LSN. Log recovery
modified to take a final LSN to recover up to.

`--binlog-info` option has been removed, binary log info always
stored during the backup. Code related to LOCK BACKUP FOR BINLOG
removed because this feature won't be ported to PS 8.

Final stage of the backup now looks like following:

1.  Issue FTWRL is needed. Conditions for FTWRL are:
    -   MyISAM tables
    -   `-slave-info` is specified and server is not a GTID slave with
        auto positioning
2.  Copy MyISAM tables
3.  Query `p_s.log_status`, save LSN and binary log position
4.  Query SHOW SLAVE STATUS if `-slave-info` is specified
5.  Stop log copying thread at the LSN obtained from `log_status`
6.  Unlock tables if they were locked at step 1

Tests changed:

`binlog_info.sh`: test that binlog info is always created during the
backup stage

`bug1277403.sh`: create MyISAM table to force FTWRL

`bug1343722.sh`: produce the workload on the server we backing up

`bug1630841.sh`: privileges adjusted

`ib_slave_info.sh`: check whether FTWRL was used

`kill_long_selects.sh`: create MyISAM table to force FTWRL

`pxb-1569.sh`: privileges adjusted

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -4410,7 +4410,7 @@ static int innobase_init_files(dict_init_mode_t dict_init_mode,
   directories.push_back(FIL_PATH_SEPARATOR);
   directories.append(MySQL_datadir_path.path());
 
-  err = srv_start(create, directories);
+  err = srv_start(create, directories, ULLONG_MAX);
 
   if (err != DB_SUCCESS) {
     DBUG_RETURN(innodb_init_abort());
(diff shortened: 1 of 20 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/42/

PR #538, opened 2018-09-20
PXB-1658: Import keyring vault plugin from PS 8

gl-sergei:
keyring_vault plugin imported from PS 8


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/47/

PR #543, opened 2018-09-25
PXB-1660: InnoDB: Log block N at lsn M has valid header, but checksum…

gl-sergei:
… field contains Q, should be P

xtrabackup was run as follows:

xtrabackup --prepare --transition-key --innodb-redo-log-encrypt

In this mode, xtrabackup created fresh redo log and encrypted it. It did
not save the key however. During subsequent restart of InnoDB engine,
the key was not found and this error message was displayed telling that
xtrabackup is not able to properly decrypt and validate redo log block.

Fix is to save encryption key between restarts of InnoDB.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/51/

PR #545, opened 2018-09-26
PXB-1665: Fix two issues reported by ubsan

dutow:


storage/innobase/buf/buf0dblwr.cc
```diff
@@ -454,6 +453,8 @@ dberr_t buf_dblwr_init_or_load_pages(pfs_os_file_t file, const char *path,
       }
 
     } else if (load_corrupt_pages) {
+      ut_ad(recv_sys != nullptr);
```
gl-sergei: better to replace this with `ut_error` ? as per comment in `fsp0sysspace.cc` "XtraBackup never loads corrupted pages from the doublewrite buffer"

gl-sergei (changes requested):

dutow:
Updated.

gl-sergei (approved):

PR #546, opened 2018-09-26
PXB-1666: Fix memory leaks reported by asan

dutow:


dutow:
Looks like this wasn't reported by jenkins because the test is skipped there:

> kill_long_selects                        w4	[skipped]   Incompatible with AddressSanitizer

gl-sergei:
Can we enable this test now or there are more issues with it?

dutow:
Yes, I've tested that on jenkins too, it works: https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/56/

I removed the disabling rule from 3 tests - kept it there for one, as it still crashes with gcc. (but passes without any leaks with clang)

gl-sergei (approved):

PR #547, opened 2018-09-26
PXB-1653: Error messages specific to PXB 8.0

gl-sergei:
Changed wording of error messages for unsupported server versions.
Changed formatting of some error messages.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -350,6 +350,8 @@ static bool check_server_version(unsigned long version_number,
                                  const char *version_comment,
                                  const char *innodb_version) {
   bool version_supported = false;
+  bool mysql51 = false;
+  bool pxb24 = false;
 
   mysql_server_version = version_number;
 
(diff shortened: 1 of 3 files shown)


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -367,9 +367,10 @@ static bool check_server_version(unsigned long version_number,
       version_supported || (version_number > 80000 && version_number < 90000);
 
   if (!version_supported) {
-    msg("Error: Unsupported server version: '%s'. Please "
-        "report a bug at "
-        "https://jira.percona.com/projects/PXB\n",
+    msg("Error: Unsupported server version: '%s'.\n"
+        "This version of Percona XtraBackup can only perform backups and "
+        "restores against MySQL 8.0 and Percona Server 8.0, please use Percona "
+        "XtraBackup 2.4 for this database.\n",
```
dutow: This condition only checks that the database is unsupported by pxb8, it doesn't guarantee that it is supported by 2.4 - for example, I removed the mariadb version checks in my previous update to this function, and now with this, we would claim that the user should be able to backup/restore even the newest mariadb database with 2.4.

We could reintroduce those checks here, or be more specific in the last part, e.g. "please use Percona XtraBackup 2.4 for databases on server X or Y"?

gl-sergei: Good point. Lets print print "please use xtrabackup 2.4" only for versions which are supported by 2.4

PR #553, opened 2018-10-02
PXB-1669: Preparation of incremental backup displays database corrupted

gl-sergei:
The part of the PXB-1627 patch was to stop redo log recovery at the LSN
stored in the xtrabackup_checkpoints file. The issue is that when
xtrabackup is called on already prepared backup, it has nothing in redo
logs to apply, instead it creates new empty redo logs. This moves LSN a
bit further and at the end the last checkpoint LSN becomes greater than
the last applied LSN. It triggers the error message.

The fix is to recognise the case when log is already applied and avoid
setting the hard limit on the LSN.

Another issue this patch is addressing is that in some cases redo log
recovery would not stop at intended position and it could lead to
assertion failures.

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3173,12 +3173,13 @@ bool meb_scan_log_recs(
       recv_sys->recovered_lsn = recv_sys->parse_start_lsn;
     }
 
-    scanned_lsn += data_len;
-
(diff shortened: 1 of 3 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/72/

PR #555, opened 2018-10-03
PXB-1655: Support ddl locking with mysql 8.0

dutow:


dutow:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/73/

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -960,9 +960,11 @@ bool lock_tables_for_backup(MYSQL *connection, int timeout) {
     return (true);
   }
 
-  msg_ts("Error: LOCK TABLES FOR BACKUP is not supported.\n");
+  tables_locked = true;
```
gl-sergei: by setting this, will we avoid acquiring FTWRL later on, even when MyISAM tables are present?

dutow: No - the condition to avoid that requires PS. With this, the lock-ddl option is supported when backing up mysql, and the test case I modified passes because of this. But we will still execute FTWRL when we did before, as LIFB only blocks ddl operations, and using only it would break some use cases.

gl-sergei: OK. It is fine than. One more thing - I don't see UNLOCK INSTANCE.

gl-sergei: I guess sequence will be
LOCK INSTANCE FOR BACKUP
backup 
UNLOCK TABLES (instead of UNLOCK INSTANCE)

dutow: Fixed - now we execute unlock instance too.

gl-sergei: But now UNLOCK TABLES will be unnecessarily invoked when backup was taken with --lock-ddl and without FTWRL. What is the reason for setting `tables_locked` here? If there are places where xtrabackup relies on the knowledge that LTFB or LIFB was invoked, we can check for `tables_locked || instance_locked`...

dutow: I kept that there to avoid accidentally not unlocking tables we should unlock - as the unlocking part had no conditions depending on `tables_locked`. I've added a condition to it now, with this, we only unlock if tables_locked is set, and looks like that didn't decrese the unlock numbers in the tests.

gl-sergei (approved):

PR #556, opened 2018-10-10
PXB-1672: Allow MTS slave without GTID to be backed up with --safe-sl…

gl-sergei:
…ave-backup

- Only fail when --slave-info is used with MTS without GTID and without
  --safe-slave-backup.
- Debug assertion added to write_slave_info to make sure that slave SQL
  thread is stopped if slave is multi-threaded.
- Error message changed to reflect code changes.
- Test case modified to check --slave-info works in conjunction with
  --safe-slave-backup.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param-medium/15/

PR #559, opened 2018-10-16
PXB-1678: Incremental backup prepare with apply-log-only rolls back u…

gl-sergei:
…ncommitted transactions

The fix is to explicitly skip rollback, IBUF merge and update of
persistent dict stats when prepare was started with --apply-log-only as
it was done in previous xtrabackup versions.

Special care is taken to free update_undo_list and insert_undo_list on
shutdown.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/76/

Single test failure is not related to this PR. I am investigating it.

PR #560, opened 2018-10-22
Update documentation for PXB 8.0

borys-belinsky:


gl-sergei:
Hi! I guess 8.0 dosc need massive update because `innobackupex` which has been deprecated since 2.3 has finally been removed and most of our docs and examples still describe `innobackupex`. Also, stream format `tar` is no longer supported in favour of `xbstream`. Is this something we want to address in this PR, or there will be separate ones?

PR #562, opened 2018-10-24
PXB-1690: Test ib_slave_info is failing

gl-sergei:
Test case was expecting UNLOCK TABLES to be issued when it was not
needed.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/81/

PR #568, opened 2018-11-06
PXB-1668: Off by one error for --throttle option

gl-sergei:
Fix off-by-one error in counting io tickets.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param-medium/17/

PR #572, opened 2018-11-07
PXB-1679: Crash after truncating table or partition when the backup i…

gl-sergei:
…s taken

Behaviour change between MySQL 5.6 and 5.7 is that TRUNCATE TABLE no
longer translates into DROP TABLE and CREATE TABLE. File is actually
truncated in 5.7.

In order to handle this, following has been done:

- os_file_read replaced with os_file_read_no_error_handling to give more
  control over error handling to caller in the fil_cur
- read_filter interface changed in order to handle partial reads
- partial read is no longer a fatal error when EOF is reached
  unexpectedly


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/50/

PR #575, opened 2018-11-08
PXB-1674: xbcloud, xbcrypt and xbstream should have the same version …

gl-sergei:
…as xtrabackup

1. Make xbstream, xbcrypt and xbcloud report xtrabackup version
2. Update year in copyright messages
3. Don't print grcypt version when encryption is not needed


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/94/

PR #576, opened 2018-11-12
PXB-1700: PXB8 crashes with ssl-mode option

gl-sergei:
Server and client versions of SSL variables were mixed in xtrabackup.
This commit makes xtrabackup to recognize client SSL variables and avoid
linking against server SSL variables.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/95/

PR #588, opened 2018-12-03
PXB-1714: Backup of server with open trx is not restored correctly

gl-sergei:
MySQL 8.0.12 introduced instant add column, which not only changed row
format, but also added new properties to data dictionary object and
internal InnoDB objects.

xtrabackup did not set number of instant columns properly when converted
from DD into dict_table_t.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/116/

PR #591, opened 2018-12-06
PXB-1721: Issues with partitioned tables

gl-sergei:
Call to deserialize_tablespace_ref always returned error, which stopped
deserialization of partition indexes at first entry. Since xtrabackup
does not load and use dd::Tablespace objects, it is safe to skip
resolving of tablespace_ref.

DD object for individual partition was named "schema/table.partition".
Fix changes naming to "schema/table#P#partition#SP#subpartition" to
allow lookup partitions by tablespace file name.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/120/

PR #594, opened 2018-12-13
PXB-1691: OOM error when backing up using --lock-ddl-per-table

gl-sergei:
PXB-1698: Percona Xtrabackup scanning all rows of partitioned tables
when using --lock-ddl-per-table

The bug itself has been fixed along with PXB-1539. The root cause was
that

SELECT * FROM t#p1 LIMIT 1

is interpreted my MySQL as "SELECT * FROM t" and "#p1 LIMIT 1" is
considered as comment which leads to all rows being fetched from
partitioned tables.

This patch adds couple of tests for partitioned tables.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/12/

PR #596, opened 2018-12-16
PXB-1616: Invalid tablespaces are not cleaned properly for SYS_TABLES…

gl-sergei:
…PACES and SYS_DATAFILES during recover

As described in the bug report, delete from SYS_TABLESPACES statement is
looking for space_id by querying SYS_TABLES, however it never able to
find the space_id because the table has already been deleted from
SYS_TABLES.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/10/

PR #602, opened 2018-12-20
PXB-1632: Corrupted double-linked list

gl-sergei:
When data file cannot be copied, data copying thread reports an error,
cleans up some of it's resources and invokes exit immediately. In case
of 'BROKEN PIPE' error and parallel backup all threads will do the same
at approximately same time. This causing exit being invoked by multiple
threads, while other threads still cleaning up their memory.

Fix allows all threads to shutdown gracefully before exiting xtrabackup.
It fixes double free errors and corrupted linked list errors reported by
system memory allocator. It doesn't make xtrabackup to shutdown cleanly
and properly close all files, so ASAN will still report memory leaks.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -488,6 +488,7 @@ typedef struct {
 	uint			num;
 	uint			*count;
 	os_ib_mutex_t		count_mutex;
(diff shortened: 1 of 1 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%202.3/job/percona-xtrabackup-2.3-param/12/

PR #607, opened 2019-01-08
xbstream: Built-in decompression support

rikonen:
Allow direct quicklz decompression as part of stream processing to
make restoration process more streamlined and symmetric with the backup
creation command.

Previously extracting xbstream package did not remove compression from
files and running separate command was required to uncompress them.
That separate command also required installing another external
dependency, which is not generally available from standard sources on
many distributions. Also, having to have both the compressed and
uncompressed files on disk simultaneously increased disk space
requirement during restoration.


storage/innobase/xtrabackup/src/ds_decompress.cc
```diff
@@ -0,0 +1,600 @@
...
+  decomp_file = (ds_decompress_file_t *)file->ptr;
+  dest_file = decomp_file->dest_file;
+
+  rc = ds_close(dest_file);
+
+  my_free(decomp_file->buffer);
+  decomp_file->buffer = NULL;
+  my_free(file);
+
+  if (decomp_file->state != STATE_REACHED_EOF) {
```
gl-sergei: Memory use after free

storage/innobase/xtrabackup/src/ds_decompress.cc
```diff
@@ -0,0 +1,600 @@
...
+      decompress code expects size bytes to be part of the data buffer */
+      if (decomp_file->nbytes_read == 0 &&
+          nbytes_available == nbytes_missing) {
+        decomp_file->nbytes_expected = qlz_size_compressed(*ptr);
+        if ((*len) >= decomp_file->nbytes_expected) {
+          decomp_file->state = STATE_PROCESS_DIRECT_BLOCK_DATA;
+          return 0;
+        } else {
+          decomp_file->state = STATE_BUFFER_BLOCK_DATA;
+          decomp_file->buffer = (char *)my_malloc(
```
gl-sergei, line 385: It looks like `decomp_file->buffer` leaks.

gl-sergei (changes requested):
Hi @rikonen ,
This feature is indeed very useful and we have long standing request to implement it. Thank you for looking into it!

I haven't done full review yet, but I spotted couple of memory issues (see comments). Please consider fixing them.

storage/innobase/xtrabackup/src/ds_decompress.cc
rikonen: Oops. Fixed

storage/innobase/xtrabackup/src/ds_decompress.cc
rikonen: It shouldn't. This code is ever reached after previously allocated buffer has been moved to `decomp_thread_ctxt_t` of one of the threads and that buffer is freed separately after the thread finishes processing the data.

gl-sergei: I see. I missed that bit.

rikonen:
@gl-sergei any idea when you might have time to have a more thorough look at this?

gl-sergei:
@rikonen , I will do my best to complete full review in the beginning of the next week.

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -62,14 +63,24 @@ static ulong opt_encrypt_algo;
...
      0, 0, 0, 0, 0},
     {"create", 'c', "Stream the specified files to the standard output.", 0, 0,
      0, GET_NO_ARG, NO_ARG, 0, 0, 0, 0, 0, 0},
+    {"decompress", OPT_DECOMPRESS, "Decompress individual backup files.",
+     &opt_decompress, &opt_decompress, 0,
+     GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
+    {"decompress-threads", OPT_DECOMPRESS_THREADS,
+     "Number of threads for parallel data decompression. The default value is 1.",
+     &opt_decompress_threads,
+     &opt_decompress_threads, 0, GET_UINT, REQUIRED_ARG, 1, 1,
```
gl-sergei: `opt_decompress_threads` is int, but `GET_UINT` is used here. `ds_compress_decompress_threads` is `uint`. they should be of the same type and use appropriate `var_type`

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -62,14 +63,24 @@ static ulong opt_encrypt_algo;
 static char *opt_encrypt_key_file = NULL;
 static char *opt_encrypt_key = NULL;
 static int opt_encrypt_threads = 1;
+static bool opt_decompress = 0;
+static int opt_decompress_threads = 1;
 
-enum { OPT_ENCRYPT_THREADS = 256 };
+enum { OPT_DECOMPRESS = 258, OPT_DECOMPRESS_THREADS = 257, OPT_ENCRYPT_THREADS = 256 };
```
gl-sergei: please only assign starting value, all subsequent will be automatically numbered started with 256. Something like:

```
enum { OPT_DECOMPRESS = 256, OPT_DECOMPRESS_THREADS, OPT_ENCRYPT_THREADS };
```

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -355,6 +367,8 @@ static file_entry_t *file_entry_new(extract_ctxt_t *ctxt, const char *path,
 
   if (ctxt->ds_decrypt_ctxt && ends_with(path, ".xbcrypt")) {
     file = ds_open(ctxt->ds_decrypt_ctxt, path, NULL);
+  } else if (ctxt->ds_decompress_ctxt && ends_with(path, ".qp")) {
```
gl-sergei, line 369: I wonder what will happen if archive would contain a mix of `.qp.xbcrypt`, `.qp` and `.xbcrypt` files

gl-sergei (changes requested):
The patch looks good. I have just a few comments. Please also format the patch with clang-format 5.0

storage/innobase/xtrabackup/src/xbstream.cc
rikonen: Changed `opt_decompress_threads` to `uint`

storage/innobase/xtrabackup/src/xbstream.cc
rikonen: Changed as suggested

storage/innobase/xtrabackup/src/xbstream.cc
rikonen: All those combinations should work. Anything ending with `.xbcrypt` goes to decrypt sink which, when `--decompress` is given, passes the file next to decompress sink. That'll check the file extension and if it is `.qp` it'll decompress but otherwise it'll just pass through to next sink without decompressing.

rikonen:
@gl-sergei Formatted with clang-format and addressed other comments as well.

gl-sergei (approved):

gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/144/

PR #610, opened 2019-01-17
Make --safe-slave-backup support multiple replication channels (Was: Avoid FTWRL + parallel replication deadlock)

ejortegau:
We have observed that sometimes, taking backups against setups with
parallel replication ends up in a deadlock between FLUSH TABLES WITH
READ LOCK and one of the running applier threads. Whether this is a
server issue or not is open to discussion, but a workaround while
taking backups seems appropriate to be able to run them against
older MySQL versions.

At a high level. the workaround is stopping the SQL thread before
doing FTWRD, and starting them afterwards. In practice, we need to
make sure we only start SQL threads for replication channels /
connections that were running before we stopped them, and we need
to support both vanilla MySQL dialect as well as MariaDB dialect
(Eg. STOP SLAVE SQL_THREAD vs STOP ALL SLAVES SQL_THREAD; START
SLAVE SQL_THREAD FOR CHANNEL 'channel_name' vs START SLAVE
'channel_name' SQL_THREAD). This is what this pull request implements.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -1043,8 +1043,8 @@ bool build_channel_name_status_set(MYSQL_RES *res, short channel_name_position, 
   sql_threads_running = false;
   short n_running_sql_threads = 0;
   while ((row = mysql_fetch_row(res))) {
-    if (std::__cxx11::string(row[sql_thread_status_position]) == std::__cxx11::string("Yes")) {
-      sql_thread_running_set.insert(std::__cxx11::string(row[channel_name_position]));
+    if (std::string(row[sql_thread_status_position]) == std::string("Yes")) {
+      sql_thread_running_set.insert(std::string(row[channel_name_position]));
       sql_threads_running = true;
       n_running_sql_threads++;
     }
@@ -1065,12 +1065,12 @@ void get_channel_name_and_status_position(MYSQL_RES *res, const std::string &sql
   bool found_thread_status = false;
   MYSQL_FIELD *field;  // Helper for query results' rows
   while ((field = mysql_fetch_field(res))) {
-    if (std::__cxx11::string(field->name) == channel_field_name) {
+    if (std::string(field->name) == channel_field_name) {
       channel_name_position = i;
       found_name = true;
       continue;
     }
-    if (std::__cxx11::string(field->name) == sql_thread_status_field_name) {
+    if (std::string(field->name) == sql_thread_status_field_name) {
       sql_thread_status_position = i;
       found_thread_status = true;
       continue;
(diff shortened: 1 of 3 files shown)


dveeden:
This is related to:
* https://bugs.mysql.com/bug.php?id=93089
* https://jira.mariadb.org/browse/MDEV-7818

We have seen this happen on 5.7.24

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -1017,8 +1023,51 @@ bool lock_tables_ftwrl(MYSQL *connection) {
     xb_mysql_query(connection, "SET SESSION wsrep_causal_reads=0", false);
   }
 
+  // Find out whether the server is a slave and, if so, stop SQL threads before flushing tables.
+  mysql_result = xb_mysql_query(connection, "SHOW SLAVE STATUS", true);  // Is there any replica configuration?
+  // Which DB vendor are we talking to?
```
dveeden: ?

dveeden:
Does this deal with older MySQL and MariaDB versions which don't support multiple replication channels/connections?

gl-sergei:
Bug 93089 is against MySQL 5.7, but the patch is for xtrabackup 8.0 which does not support MySQL 5.7. Is this bug present in MySQL 8.0?
I don't like the idea of stopping sql slave thread unconditionally. Besides we have very similar thing called `--safe-slave-backup`

ejortegau:
@dveeden:  I can change it to cover versions that don't support multiple replication channels. Hoping to push a change with that soon.
@gl-sergei: We are not currently running 8.0+ backup slaves, so I am not sure whether the bug is still present or not. However, since it is marked as "could not reproduce", I can only assume it has not been fixed.
Before I go ahead and start making changes to the code, I'd like to get some agreement on what good would like. 

I think that the way this pull request stops SQL threads is better than `--safe-slave-backup`, as it considers multi-channel replication. As it is now, `wait_for_safe_slave()` will do `STOP SLAVE SQL THREAD`, which in MariaDB will only stop the default replication connection. It will also do `START SLAVE SQL_THREAD` which in MySQL will start SQL threads for all channels (regardless of whether they were running before), and in MariaDB will only start it for default connection. Or, at least, that's my current understanding.

I am tempted to go like this:

* Condition the stopping and starting of SQL threads to the usage of the `--safe-slave-backup` is specified flag.

Does that seem OK?

ejortegau:
@gl-sergei : 

I have changed this to be part of --safe-slave-backup, supporting multiple replication channels. This addresses your concern about stopping the SQL thread unconditionally, and improves the scenario that --safe-slave-backup covers. Would you consider it for merging?

ejortegau:
BTW, I have renamed the PR since, as it stands, it is not aimed at solving the deadlock by stopping the SQL thread, but instead, at ensuring that existing option `--safe-slave-backup` supports setups with multiple replication channels/connections.

ejortegau:
Hi, @gl-sergei : is there any feedback on this?

altmannmarcelo:
@ejortegau picking up on stalled PR.

First of all thanks for your contribution.

Can you please let us know if this is still an issue? If so, is this patch still valid?

I see there is some logic behind finding the appropriate server flavor (Mariadb, ...).  Versions 2.4 and 8.0 are no longer 100% compatible with MariaDB, thus to make new code more efficient, we can assume MySQL/PS only.

2.4 patch -  #616

altmannmarcelo:
Closing this as no feedback has been provided. We can revisit in the future if the issue persists.

PR #619, opened 2019-01-29
PXB-1771: Copy-back thread reports an error when is has not done

gl-sergei:
anything

Return value in copy_back_thread_func is initialized to false (means
error). Inside of the loop through the files it's value gets assigned to
the result of the file copy operation. However if there was not a single
file to copy by this thread, ret remains false and the thread reports
back an error.

Fix is to initialize ret as true at the beginning.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/23/

PR #621, opened 2019-01-30
PXB-1775: PXB displays path error for undo directory when undo logs

gl-sergei:
exist outside data directory

- change default for innodb_undo_tablespaces to 2
- avoid using Fil_path constructor because it prints out error message
  if file does not exist. Use static methods of Fil_path instead


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/157/

PR #623, opened 2019-02-05
PXB-1770: Issues with prepared XA transactions

gl-sergei:
Xtrabackup behaviour prior to 8.0 was to always rollback prepared XA
transactions in the InnoDB prepare. It was fine since InnoDB was the
only transactional storage engine.

Now that we ship Percona Server with RocksDB, it is better to let server
to decide whether to rollback or commit prepared transactions at
startup. For this purpose we need also copy the binary log, because it
is eventually the source of truth for the server to decide whether to
commit or rollback (in combination with --tc-heuristic-recover).

On the other hand, PXC is the another example of the case where two
transnational storage engines are enabled and it always relied on old
xtrabackup behaviour to rollback transactions. For PXC we provide an
option --rollback-prepared-trx.

### DIFF
--- a/storage/innobase/include/srv0srv.h
+++ b/storage/innobase/include/srv0srv.h
@@ -551,6 +551,7 @@ extern bool srv_apply_log_only;
 
 extern bool srv_backup_mode;
 extern bool srv_close_files;
+extern bool srv_rollback_prepared_trx;
 
 /*-------------------------------------------*/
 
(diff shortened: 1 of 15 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/165/

gl-sergei:
Alternatively, we can allocate single large buffer instead of one per thread and share it between all threads, but it will affect scalability
Also we can release large buffer as soon as we logged the record. I am not sure if it will increase memory fragmentation significantly.

PR #624, opened 2019-02-06
PXB-1779: xtrabackup fails to apply deltas to the tablespaces created

gl-sergei:
between full and incremental backups

xb_delta_open_matching_space did not expect undo tablespaces to be
created between full and incremental backups resulting in the error
while applying the delta.

Fix is to create undo tablespace if it is missing in the full backup but
is present in the incremental one. Since undo tablespaces cannot be
renamed, new xb_delta_open_matching_space does not attempt to find a
renamed version.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/166/

PR #628, opened 2019-02-17
PXB-1784: ALTER UNDO TABLESPACE fails after restore when undo tablespace

gl-sergei:
state is changed during backup

This bug is a result of unfortunate combination of two conditions:

1. xb_page_bitmap_init creates a bitmap object when incremental_lsn
   equals to last checkpoint lsn
2. LOCK INSTANCE FOR BACKUP paused ALTER UNDO TABLESPACE at the moment
   when it has flushed all changes

As a result even though vanilla mysql does not support changed page
bitmaps, xtrabackup opted to use changed page bitmaps read filter and
produced zero sized .delta file for undo tablespace (which is the only
one changed between full and incremental backups).

The fix is to check if server indeed support changed page bitmaps when
choosing the read filter.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -4401,7 +4401,7 @@ void xtrabackup_backup_func(void) {
     debug_sync_point("xtrabackup_suspend_at_start");
 
     if (xtrabackup_incremental) {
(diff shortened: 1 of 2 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/175/

PR #629, opened 2019-02-18
PXB-1785: PXB crashes in incremental backup prepare when MS is running

gl-sergei:
with innodb-rollback-segments

Previous MySQL versions even though had the innodb-rollback-segments
setting did not actually physically change the default number of
rollback segments (128) in the undo logs. It has been changed in MySQL
8.0. When first started with smaller number or rollback segments, server
creates exactly this number of rollback segments in the undo files. When
later started with greater number it adjusts the number of rollback
segments in the undo files.

This new behavior is not desirable for xtrabackup since it breaks
preparation of incremental backups by making undo logs physically
different for the ones on the server.

The fix is to avoid creating of missing rollback segments when run with
--apply-log-only.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/178/

PR #640, opened 2019-03-05
PXB-1801: Xtrabackup fails if log_bin_index is defined in my.cnf

gl-sergei:
xtrabackup did not append .index when log-bin-index value was not ending
with .index and was read from my.cnf.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/182/

PR #654, opened 2019-04-22
PXB-1824: PXB crashes in incremental backup prepare when row format is

gl-sergei:
changed

This patch fixes several issues.

First one is a 0-day bug in the incremental backups logic. Pages are not
included into incremental backup if page's LSN is less than or equal to
the incremental LSN. Page's LSN points to the end of the last log record
applied to the page. It means that this last record will be missing from
the incremental backup in the case when checkpoint LSN did not advance
between full and incremental backups, causing data corruption.

The fix is to change the condition in the incremental filter.

Second one is related to online alter skipping redo logs. Xtrabackup is
using LOCK TABLES FOR BACKUP in order to stop online DDL and rely on
InnoDB to flush the changes on disk. There is a caveat though,
xtrabackup copies the redo starting with checkpoint LSN which is earlier
than the last flushed LSN. Some DDL may rebuild table using following
sequence:

Lets say we rebuilding the table t.

1. Create the table tmp1
2. Build the new table in the tmp1 using fast index creation and
   skipping redo logs
3. Rename the table t into tmp2
4. Rename the tmp1 into t
5. Drop the tmp2

Backup has been started right after 5. Now if the checkpoint was after
5, xtrabackup is good. If checkpoint was before 1, xtrabackup will
delete the good table t flushed by server and replace it with the table
with some of the rows (or all of them) missing.

The fix is to remember the last flushed LSN at the beginning of the
backup and do not replay file operations for LSNs older than that.

Third one is a memory leak. Resurrected transactions were not rolled
back (this is a feature) and memory was not freed at shutdown (this is a
bug). Fix is to avoid undo log scan altogether when prepare is run with
apply-log-only.

Also backported fix for PXB-1773.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -2512,8 +2512,9 @@ fil_op_log_parse_or_replay(
 	ulint	space_id,	/*!< in: the space id of the tablespace in
 				question, or 0 if the log record should
 				only be parsed but not replayed */
-	ulint	log_flags)	/*!< in: redo log flags
+	ulint	log_flags,	/*!< in: redo log flags
 				(stored in the page number parameter) */
+	bool	apply)		/*!< in: whether to apply the record */
 {
 	ulint		name_len;
 	ulint		new_name_len;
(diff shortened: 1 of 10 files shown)


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/43/

PR #658, opened 2019-04-24
PXB-1810: PXB on move-back of binlog doesn't change the name of the

gl-sergei:
binlog as per apply node configuration

xtrabackup --copy-back and --move-back restored the binary log with the
same name/location it had on the original server which is not always
desired.

The fix is to respect the --log-bin and --log-bin-index options for
--copy-back and --move-back.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/203/

PR #661, opened 2019-04-25
PXB-1837: various issues with xbcloud

gl-sergei:
- Memory allocated on stack by xblcoud_<put/delete/download> may have
  been accessed by the callbacks after the function has exited. Fixed by
  making xblcoud_<put/delete/download> to wait for completion of all
  asynchronous requests.

- xbcloud/xbstream on centos 7 did not preserve alphabetical
  order. Fixed by passing ls output through the sort command.

- Older libcurl used on Centos 6 freed internal structures of an
  easy_handle when curl_multi_remove_handle was called. These structures
  were than accessed in Http_connection::finalize. Fix swaps two calls.

- Older libcurl version used in Centos 6 could fail to return valid
  value for CURLINFO_EFFECTIVE_URL. Error code check was added.

- Older libcurl version used in Centos 6 does not understand
  Content-Length header passed manually and ended up sending two
  Content-Length headers. Which broke S3 authentication.

- Older libcurl version used in Centos 6 can return
  CURLM_CALL_MULTI_PERFORM in response to curl_multi_socket_action. This
  error code was not handled by mcode_or_die.

- Class Object_store did not have virtual destructor which caused memory
  leaks of small amount of resources.

- Because of the bad timings, event loop could have started after we
  pushed some work into the queue which caused missed async events and
  timeouts. Fixed by adding a kickoff timer who's only goal is to
  indicate that event loop has started.

- Even though libcurl claims to be thread-safe, this is not always true
  when it comes to using it with OpenSSL. xbcloud makes all HTTP
  requests using curl_multi in single thread. There is another thread
  which can call curl_easy_escape. curl_easy handle was created for each
  of such calls. Crash was observed on Debian 9 and the stack trace
  included curl_easy_cleanup followed by OpenSSL cleanup code. It is
  known that in order to use older versions of OpenSSL in multi-threaded
  environment, one needed to provide locking callbacks. Debian 9,
  however, comes with OpenSSL 1.1.0j, which has built-in support for
  multi-threading which should just work. Fix (more a workaround) is to
  use single curl_easy for all curl_easy_escape calls, guard the calls
  with mutex (which is not really needed at the moment, because all url
  escape calls come from one thread).

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/http.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/http.cc
@@ -31,9 +31,47 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 
 namespace xbcloud {
 
+class Global_curl {
+ private:
+  CURL *curl{nullptr};
+  std::mutex mutex;
+
+ public:
+  Global_curl() {}
+
+  bool create() {
+    curl = curl_easy_init();
+    return (curl != nullptr);
+  }
+
+  void cleanup() {
+    curl_easy_cleanup(curl);
+    curl = nullptr;
+  }
+
+  void lock() { mutex.lock(); }
+
+  CURL *get() const { return curl; }
+
+  void unlock() { mutex.unlock(); }
+};
+
+static Global_curl global_curl;
+
+bool http_init() {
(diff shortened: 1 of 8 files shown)


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/57/

PR #664, opened 2019-04-29
PXB-1601: compact option - remove

borys-belinsky:


gl-sergei (approved):

gl-sergei:
please also make sure it merges cleanly into 8.0 and create PR for if it doesn't

PR #665, opened 2019-04-30
Update version

EvgeniyPatlan:


EvgeniyPatlan:
Self-approve as it is just version update

PR #673, opened 2019-05-06
PXB-1850: SIGPIPE handling in xbcloud

gl-sergei:
When xbcloud output is piped to another command and this command is
terminated for some reason, xbcloud normally gets killed by SIGPIPE. It
is observed that in some environments xbcloud is not killed and hangs.

Fix makes xbcloud to always ignore SIGPIPE and exit with error when it
cannot write to the output file.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/60/

PR #678, opened 2019-05-08
Test pxb-1536.sh is unstable with MySQL 8.0.16

gl-sergei:
Default behavior in xtrabackup since 8.0 is to skip rollback of prepared
XA transactions. This is done to let MySQL decide on startup whether to
rollback or commit these transactions.

Combined with partial backups this may crash the server on startup if XA
transactions are touching missing tables. To avoid this, one must pass
the --rollback-prepared-trx option when preparing partial backups.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/218/

PR #686, opened 2019-05-17
Fix PXB-1867 in 8.0

gl-sergei:
Besides fixing PXB-1867 itself this patch also fixes an issue when ROW_FORMAT=COMPRESSED was not recognized by xtrabackup during prepare.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/230/

PR #689, opened 2019-05-17
PXB-1864: External XA-prepared transactions are rolled back after pre…

fungo:
…pare

Desctription
============
Starting MySQL 5.7, the new XA_prepare_log_event binlog event type is
introduced to persist XA-prepared transactions into binary log, so that
prepared XA-transaction will survive both a client reconnects or the server
restarts (see WL#6860).

But currently PXB will always rollback prepared transactions, this strategy
is based on 2PC implementation prior to MySQL 5.7, and is not suitable now.

Suppose an external xa transaction record in binlog like bellow, the 1st
column is position:

10 | GTID:1
20 | XA START 'test1'
30 | Update_rows
40 | XA END 'test1'
50 | XA PREPARE 'test1'

---- Backup finished here ----

60 | GTID:2
70 | XA COMMIT 'test1'

And suppose the backup finished after XA PREARE but before XA COMMIT, when
the backup result is prepared, the XA-repared transaction 'test1' is rolled
back.

If we setup the replication from the new instance prepared from backup, to
the source instnace, and start replication from position 60 (the position
recorded in backup result), the replication will broke with error.

Solution
========
From MySQL 5.7, external XA-prepared is designed to survive both client
reconnects or the server restarts, so PXB backup and restore should follow
this design.

When prepare a backup, only rollback the internal XA-prepared transactions,
but keep the external XA-prepared transactions in prepared state.

### DIFF
--- a/storage/innobase/trx/trx0trx.cc
+++ b/storage/innobase/trx/trx0trx.cc
@@ -842,9 +842,40 @@ trx_resurrect_insert(
 
 			if (srv_force_recovery == 0) {
 
-				/* XtraBackup should rollback prepared XA
-				transactions */
-				trx->state = TRX_STATE_ACTIVE;
+				/* XtraBackup should rollback internal or wsrep
+				prepared XA transactions, but leave external
+				prepared XA transactions untouched. */
+				if (trx->xid->get_my_xid()) {
+					/* internal XA transaction */
+					trx->state = TRX_STATE_ACTIVE;
+
+					ib::info() << "Transaction "
+						   << trx_get_id_for_print(trx)
+						   << " was an internal "
+						"prepared XA transaction, and "
+						"will be rolled back.";
+				} else if (wsrep_is_wsrep_xid(trx->xid)) {
+					/* wsrep XA transaction */
+					trx->state = TRX_STATE_ACTIVE;
+
+					ib::info() << "Transaction "
+						   << trx_get_id_for_print(trx)
+						   << " was a wsrep "
+						"prepared XA transaction, and "
+						"will be rolled back.";
+				} else {
+					/* external XA transaction */
+					trx->state = TRX_STATE_PREPARED;
+					++trx_sys->n_prepared_trx;
+					++trx_sys->n_prepared_recovered_trx;
+
+					ib::info() << "Transaction "
+						   << trx_get_id_for_print(trx)
+						   << " was an external "
+						"prepared XA transaction, and "
+						"will be kept in prepared "
+						"state.";
+				}
 			} else {
 
 				ib::info() << "Since innodb_force_recovery"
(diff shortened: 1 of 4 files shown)


storage/innobase/trx/trx0trx.cc
```diff
@@ -842,9 +842,31 @@ trx_resurrect_insert(
 
 			if (srv_force_recovery == 0) {
 
-				/* XtraBackup should rollback prepared XA
-				transactions */
-				trx->state = TRX_STATE_ACTIVE;
+				/* XtraBackup should rollback internal prepared
+				XA transactions, but leave external prepared
+				XA transactions untouched. */
+				if (trx->xid->get_my_xid()) {
```
gl-sergei, line 848: Thank you for contribution. This change looks reasonable, but it will also do not rollback WSREP XIDs (prefixed with "WSREPXid"). However, Galera Cluster rely on fact that XIDs are rolled back by xtrabackup.

fungo: Thanks for the information, I'm not familiar with Galera or mysql-wsrep :)
If the XID prefix is fixed for WSREP, I can add another condition check here.
I'll update the PR shortly.

fungo: I have pushed the updates
1. handle wsrep xid case
2. rebase to latest 2.4

storage/innobase/xtrabackup/test/t/ib_external_xa_prepared.sh
```diff
@@ -0,0 +1,42 @@
...
+# Test external XA prepare transactions in backup result are not rolled back
+# after --prepare
+############################################################################
+
+. inc/common.sh
+
+vlog "start server with gtid enabled"
+start_server --gtid-mode=ON --enforce-gtid-consistency=ON --log-bin --log-slave-updates
+
+run_cmd $MYSQL $MYSQL_ARGS test -e "create table t1(id int, primary key(id)) engine=innodb"
```
gl-sergei: can you please use `mysql` instead of `run_cmd $MYSQL $MYSQL_ARGS`

storage/innobase/xtrabackup/test/t/ib_external_xa_prepared.sh
```diff
@@ -0,0 +1,42 @@
...
+run_cmd $MYSQL $MYSQL_ARGS test -e "xa start 'test1'; insert t1 values (1); xa end 'test1'; xa prepare 'test1'"
+run_cmd $MYSQL $MYSQL_ARGS test -e "xa start 'test2'; insert t1 values (2); update t1 set id = 3 where id = 2; xa end 'test2'; xa prepare 'test2'"
+
+run_cmd $MYSQL $MYSQL_ARGS test -e "xa recover" | grep "test1"
+run_cmd $MYSQL $MYSQL_ARGS test -e "xa recover" | grep "test2"
+
+vlog "Make full backup"
+mkdir $topdir/backup
+innobackupex --slave-info --no-timestamp $topdir/backup
```
gl-sergei: Because `innobackupex` is removed in 8.0, it is better to use `xtrabackup` in new test cases. I assume this test case is passing on 8.0?

gl-sergei (changes requested):

storage/innobase/xtrabackup/test/t/ib_external_xa_prepared.sh
fungo: I have not tried 8.0, and I guess it will fail.
I'll change to use `xtrabackup`, and try 8.0.

storage/innobase/xtrabackup/test/t/ib_external_xa_prepared.sh
fungo: Sure.

fungo:
Update as suggested.
1. the name of test case is changed to `xb_external_xa_prepared.sh`, as it's intended for `xtrabackup` now
2. I tried the new test case agains 8.0, it works (for the `xtrabackup` command ) and fails as expected (for the log msg grep). BTW, I noticed there is difference for handling prepared trx in 8.0, which is keep prepared by default. This is controlled by `--rollback-prepared-trx` option.

altmannmarcelo:
Hi @fungo ,

First of all, thanks for your contribution. 

Is this PR still relevant nowadays?
If so, could you please raise a JIRA ticket describing the issue? I will move forward if the analyzes.

altmannmarcelo:
Closing this as no feedback has been provided. We can revisit in the future if the issue persists.

PR #692, opened 2019-05-18
add Hygon CPU support for innobase ut

hygonsoc:
Add Hygon CPU Vendor ID ("HygonGenuine") checking to get correct family/model, as hygon Dhyana(family 18h) share similiar arch with AMD family 17h.


gl-sergei:
Hi @hygonsoc , thank you for contribution. This code is inherited from MySQL - https://github.com/mysql/mysql-server, please also submit bug report and your patch there.

altmannmarcelo:
Hi @hygonsoc .

Thanks for your PR.

We kindly ask you to raise this with MySQL upstream as we aim to keep compatible with them. Once they claim compatibility with this CPU and merge your changes to their code, we will inherit it.

PR #693, opened 2019-05-20
PXB-1862: Misleading error after preparing an incremental backup for

gl-sergei:
second time

xtrabackup gave a notice when it detected that xtrabackup_logfile was
already used to prepare. Now it gives an error and aborts immediately.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/64/

PR #695, opened 2019-05-20
PXB-1649: System tablespace encryption

gl-sergei:
- added tests for innodb system (mysql.ibd) and shared system (ibdata1)
  tablespace encryption
- removed assertions that system shared tablespace cannot be encrypted
- disabled use of doublewrite buffer on prepare
- added the code to read encryption information from the system shared
  tablespace at startup (both for backup and prepare)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/250/

PR #708, opened 2019-06-14
PXB-1493: XtraBackup does not check validity on arguments used.

gl-sergei:
Added validation for options passed via command line and stored in the
[xtrabackup] section of defaults file.

Because xtrabackup may load keyring plugins which have their own set of
options, validation is done after plugins are loaded.

xtrabackup now initializes log_builtins to log mysql error log messages.

Changes:

- my_handle_options: added option to perform only perform validation
  without changing current option values.

- derror.cc: do not log error message about missing lc_messages

- xtrabackup.cc: validate_options called immediately after keyring
  plugin is initialized. init_error_log is called to enable output mysql
  server and plugins error messages. destroy_error_log is set as exit
  hook to make sure that log if flushed at exit.


gl-sergei:
https://jira.percona.com/browse/PXB-1493

gl-sergei:
Added boolean --strict option (OFF by default). When strict is ON
    unknown options cause error, when strict is OFF unkown options cause
    warning.

gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/280/

PR #709, opened 2019-06-21
PXB-1874: xbcloud doesn't update Date header when retrying

gl-sergei:
PXB-1875: xbcloud doesn't handle Google's HTTP 408 error correctly

Patch does following:

- recalculate AWS signature when retry
- don't try to S3 parse response as XML unless content type is
  application/xml
- set buffer size equal to read_buffer_size for buffered datasink
- retry http error 408


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/66/

PR #711, opened 2019-06-25
PXB-1879: xtrabackup does not take FS block size into account for

gl-sergei:
compressed and encrypted tables

Fix is to save block_size into file cursor and use it when
decompressing/decryting pages.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/71/

PR #713, opened 2019-06-26
PXB-1880: Test PXB-1870 is failing on ubuntu xenial

gl-sergei:
Fil_system is trying to limit the number of open files. Limit is 300 by
default. If InnoDB needs to open next file, it first checks if limit is
reached. When limit is reached InnoDB is trying to close some
files. This is where xtrabackup hangs with following symptoms:

- opened files have pending IO requests and they cannot be closed
  because of that
- os_aio_simulated_handler threads are sleeping and not doing any IO
- the thread which is trying to open new file hangs in the loop trying
  to close some files

Fix is to detect that we are not making progress for some time (1s) and
try to wake IO threads and flush changes so that files can be closed.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -2798,6 +2798,7 @@ bool Fil_shard::mutex_acquire_and_get_space(space_id_t space_id,
 
   auto begin_time = ut_time();
   auto start_time = begin_time;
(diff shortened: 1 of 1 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/278/

PR #714, opened 2019-06-26
incorrect arguments for memset_s

luqun:
The definition for memset_s(https://github.com/percona/percona-xtrabackup/blob/b04b2c4e0691054195642148fda9334fadf7ed52/include/my_sys.h#L616):

The 2nd argument is size of buffer.
The 3rd argument is character to fill.


gl-sergei:
Thank you for contribution @luqun ! Because we want to fix both 2.4 and 8.0 versions, I created these two PRs: https://github.com/percona/percona-xtrabackup/pull/720, https://github.com/percona/percona-xtrabackup/pull/721. Upstream (percona server) will also take this fix: https://jira.percona.com/browse/PS-5741

PR #715, opened 2019-06-26
use double instead ulong for variable innodb_max_dirty_pages_pct

luqun:
In InnoDB srv_max_buf_pool_modified_pct is a double. But the XtraBackup initialization code was setting it as a ulong.

https://github.com/percona/percona-xtrabackup/blob/e0f4232770fc3ac20af2bb7e075bacaa5006abeb/storage/innobase/handler/ha_innodb.cc#L20705

so if you try to set innodb_max_dirty_pages_pct  as 75.000000 in your cnf, xtrabackup will complain:
xtrabackup: [ERROR] Unknown suffix '.' used for variable 'innodb_max_dirty_pages_pct' (value '75.000000').
xtrabackup: [ERROR] xtrabackup: Error while setting value '75.000000' to 'innodb_max_dirty_pages_pct'.

The change is to use type double for innodb_max_dirty_pages_pct.


gl-sergei:
Hi. Thank you for contribution. This PS is superseded by https://github.com/percona/percona-xtrabackup/pull/718 and https://github.com/percona/percona-xtrabackup/pull/719

PR #716, opened 2019-06-27
PXB-1884: Xbstream does not decompress lz4 when decryption is also be…

gl-sergei:
…ing done

xbstream did not decompress .lz4.xbstream files.

Fix is to initialize lz4 decompression datasink at startup, it was not
initialized by mistake.

Patch also renames *compress* typedefs to *compress_lz4* to avoid
linker conflicts.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/279/

PR #717, opened 2019-06-28
PXB-1557: xbstream creates non-sparse ibd files and breaks

gl-sergei:
COMPRESSION='zlib'

Following implemented:

- added new datasink API call `ds_write_sparse' which takes a buffer and
  a sparse map and writes it preserving its sparseness.
- `ds_write_sparse' implemented in `ds_local' datasink using a
  combination of write and seek system calls
- `ds_write_sparse' implemented in `ds_xbstream' datasink. New sparse
  chunk type introduced which contains the data and the sparse map
- sparse chunk type support added to `xbstream' utility
- backup and copy back modified to detect sparse tablespaces and use
  `ds_write_sparse' to copy them
- xtrabackup will avoid compression of already compressed tablespaces
- xtrabackup when applying incremental deltas will use punch hole
  operation to preserve page sparseness

Limitations:

- `ds_write_sparse' is not supported by `ds_encrypt', encrypted backup
  will not be sparse. However, sparseness will be restored during
  prepare and copy-back.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -97,15 +97,6 @@ static std::mutex rsync_mutex;
 /* skip these files on copy-back */
 static std::set<std::string> skip_copy_back_list;
 
-/* the purpose of file copied */
-enum file_purpose_t {
-  FILE_PURPOSE_DATAFILE,
-  FILE_PURPOSE_REDO_LOG,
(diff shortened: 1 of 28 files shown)


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/287/

gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/290/

PR #722, opened 2019-07-03
PXB-1888: kill_long_selects.sh fails with ASAN

gl-sergei:
Problem:

MySQL result set is not freed by have_queries_to_wait_for and
kill_long_queries, also kill_query_thread even though it is using MySQL
APIs does not call my_thread_init.

Fix:

Release MySQL result set and call my_thread_init/my_thread_end.


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/77/

PR #725, opened 2019-07-05
PXB-1890: Update error message in 2.4 to point that MySQL 8.0 can be …

gl-sergei:
…backed up with PXB 8.0

Fixed by printing specific error message for MySQL/PS 8.0


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/79/

PR #730, opened 2019-07-12
PXB-1894: PXB crashes during incremental backup prepare for tablespace

gl-sergei:
encryption

Problem:

When encrypted table have just been created it already has encryption
flag in its tablespace header, but may not have encryption information
yet on its first page. This may happen because tablespace flags are
written synchronously and encryption information goes through the buffer
pool and will be flushed later.

Xtrabackup will skip such table in hope that it will be created on
prepare from the redo log. The problem is that crash recovery does not
actually save encryption information when it see it in the redo
log. Instead crash recovery only updates in-memory data structures. It
works in most cases but one, when both full and incremental backups
happened before InnoDB flushed the first page.

Fix:

Make crash recovery to actually apply log record to the first page when
encryption information is saved.

Because xtrabackup parsing redo log as it does the backup, it can get
encryption information for newly created encrypted table from the redo
log and include its .ibd file into a backup.

### DIFF
--- a/storage/innobase/fsp/fsp0file.cc
+++ b/storage/innobase/fsp/fsp0file.cc
@@ -658,13 +658,36 @@ Datafile::validate_first_page(lsn_t*	flush_lsn,
 				<< " datafile: " << m_filepath
 				<< " can't be decrypted.";
 
-			m_is_valid = false;
-			free_first_page();
-			ut_free(m_encryption_key);
-			ut_free(m_encryption_iv);
-			m_encryption_key = NULL;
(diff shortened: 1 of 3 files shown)


gl-sergei:
MySQL server itself is not affected, because MySQL calls `fil_encryption_rotate` at startup. PXB when run with --prepare --apply-log-only of course cannot run key rotation.

gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/81/

PR #733, opened 2019-07-31
PXB-1901: Compressed lz4 files are also copied in restored data dir

gl-sergei:
Problem:

LZ4 compressed files were copied back to the datadir with

xtrabackup --copy-back

Fix:

Add .lz4 to the skip list for copy-back


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/22/

PR #736, opened 2019-08-01
PXB-1903: PXB is stuck with redo log corruption when master key is

gl-sergei:
rotated

Problem:

While copying the datafiles, xtrabackup also parsing redo logs. When
master key is rotated the record about reeencrypting the file header
goes into redo log. Xtrabackup is trying to parse it, but cannot find
new master key because the contents of the keyring file is cached by the
keyring plugin. xtrabackup prints corruption message and stops parsing
the redo log.

Fix:

We actually can skip this error silently because tablespace key has not
changed and there are two possible cases:

1 we have already decrypted the key for this tablespace, then backup
  will proceed without issues 2 we have not yet decrypted the kye for
  this tablespace, then backup will fail when trying to copy the
  tablespace

There is a small room for race condition between update of the keyring
file and the tablespace header which will lead to second case, but there
is no way to reliably fix this race condition. Aborted backup is the
best we can do.

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -1692,10 +1692,7 @@ fil_write_encryption_parse(
 		if (!fsp_header_decode_encryption_info(key,
 						       iv,
 						       ptr)) {
-			recv_sys->found_corrupt_log = TRUE;
-			ib::warn() << "Encryption information"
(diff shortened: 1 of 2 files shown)


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/81/

PR #750, opened 2019-09-10
PXB-1912: Support log archiving feature in PXB 8.0

gl-sergei:
When innodb_redo_log_archive_dirs is set, xtrabackup will try to
enable redo log archiving at the beginning of the backup. Once archived
log is cought up with the current LSN, xtrabackup will switch to
copying/streaming redo log records from the archive instead of the live
redo log. This will improve the case when backup destination is slower
than redo log storage.

If xtrabackup is unable to create or read redo log archive, it will log
error message and continue the backup using live redo logs.

Redo copying refactored:

- redo_log_data_manager to coordinate redo log copying
- redo_log_reader to read redo logs
- redo_log_writer to write redo logs
- redo_log_parser to parse redo logs

Log archiving:

- archived_redo_log_monitor to create and open redo log archive file
- archived_redo_log_read to read redo log from archive file

Added -X parameter to run.sh to pass additional options for MySQL server
when running the test suite

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -3678,7 +3678,6 @@ dberr_t Fil_shard::iterate_spaces(bool include_log,
   return (DB_SUCCESS);
 }
 
-
 /** Iterate through all persistent tablespace files (FIL_TYPE_TABLESPACE)
 returning the nodes via callback function cbk.
 @param[in]	include_log	Include log files, if true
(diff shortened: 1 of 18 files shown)


gl-sergei:
test run with log archiving enabled by default https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/56/

test run with log archiving disabled by default https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/55/

PR #751, opened 2019-09-11
PXB-1913: PXB crashes in incremental backup prepare when a multi

gl-sergei:
valued index is added/dropped for json data

Problem:

xtrabackup did not convert multi value indexes correctly into
InnoDB data dictionary structures. It caused issues on rollback.

Fix:

1. Set correct prtype for multi value index
2. Pick correct column for indexes on virtual columns


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/57/

PR #753, opened 2019-09-17
PXB-1905: Encrypted table is not restored when ADD/DROP INDEX is run …

gl-sergei:
…on the table

Problem:

ALTER TABLE ... , ALGORITHM=COPY works as follows:

1. Create temp table
2. Copy data from original table into the temp table
3. Rename original table and temp table

Issue comes when full backup contains an version of the table and
.delta contains new version of the table. Special condition is
that the first page of the new table has not been flushed to the
dist yet.

In this case LSN of the first page in .delta will be 0 and the
page won't be included into .delta. Xtrabackup based in space_id
(stored in .meta) will guess that the table was altered, rename
old table into #xtrabackup_tmp_old_id.ibd and create tablespace
with new ID.

Because .meta does not contain space flags, they will be set as
default, then during apply log phase assertion will fail that
space flags indicate that space is not encrypted while it supposed
to be (encryption info record found).

Fix:

Store space flags into .meta

### DIFF
--- a/storage/innobase/xtrabackup/src/fil_cur.cc
+++ b/storage/innobase/xtrabackup/src/fil_cur.cc
@@ -147,6 +147,7 @@ xb_fil_cur_open(
 	cursor->node = NULL;
 
 	cursor->space_id = node->space->id;
+	cursor->space_flags = node->space->flags;
 	cursor->is_system = !fil_is_user_tablespace_id(node->space->id);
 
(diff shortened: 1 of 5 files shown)


gl-sergei:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/89/

PR #756, opened 2019-10-03
PXB-1915: MyRocks backups feature - renewing checkpoints

gl-sergei:
When backing up MyRocks datadir, xtrabackup creates checkpoint, captures
binlog coordinates and then copies the .SST files into the
backup. Copying may take a while. Then, if one want to setup new slave
or just apply binary logs on top of the backup, the size of binary logs
may become large.

This feature provides a way to minimize the size of binary logs and to
move binary coordinates closer to the moment when file copying
completes.

The idea is to create the checkpoint and copy .SST files. If copy took a
while, create next checkpoint and copy only new .SST files, second copy
should take less time. Repeat the process until copy time fall within
acceptable range.

Two xtrabackup parameters added:

- rocksdb-checkpoint-max-age - checkpoint shall not be older than this
  number of seconds when backup completes

- rocksdb-checkpoint-max-count - complete backup even if checkpoint age
  requirement has not been met after this number of checkpoints

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1135,16 +1135,20 @@ void Myrocks_datadir::scan_dir(const std::string &dir,
         }
         char buf[FN_REFLEN];
         fn_format(buf, name, path, "", MY_UNPACK_FILENAME | MY_SAFE_PATH);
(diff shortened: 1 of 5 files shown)


george-lorch (commented):
I can't comment much on the actual implementation details, but the idea sounds solid to me.

PR #757, opened 2019-10-08
PXB-1937: xtrabackup --move-back is failing due to encrypted binlog file

gl-sergei:
Problem:

When log-bin variable was set to default value, xtrabackup did not
properly expand the destination directory for the binary log file.

Fix:

1. Apply target datadir for the binary log path when performing
   copy-back/move-back and log-bin is not set
2. Check the return value of the binlog_header_read and exit gracefully
   in case of error


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/62/

PR #758, opened 2019-10-11
PXB-1928: PXB does not use the redo log archiving feature if label is…

gl-sergei:
… not specified

Problem:

XtraBackup does not use redo log archiving when label is empty.

Fix:

Remove the check for empty label.

PXB-1938: PXB leaves empty directories in redo log archive directory after backup.

Problem:

When xtrabackup is trying to remove the subdir it uses relative name
which leads to rmdir failure.

Fix:

Call rmdir with absolute path.


gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/64/

PR #772, opened 2019-11-15
PXB-1954: PXB gets stuck in prepare of full backup for encrypted tabl…

gl-sergei:
…espace

Problem:

When tablespace has just been created its encryption info is not yet written
to the first page. In this case page will be still considered valid when
tablespace is opened for recovery. Encryption key will be read and set from
the redo log using fil_tablespace_encryption_init.

It is possible however that during apply log phase tablespace with the same id
is opened multiple times. Problem is that fil_tablespace_encryption_init erases
the key which it used once and next time tablespace is opened, the key is no
longer in the recv_sys->keys list.

Fix:

Do not erase the key from recv_sys->keys list after it has been used in
fil_tablespace_encryption_init.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -9317,14 +9317,6 @@ static void fil_tablespace_encryption_init(const fil_space_t *space) {
       ib::error(ER_IB_MSG_343) << "Can't set encryption information"
                                << " for tablespace" << space->name << "!";
     }
-
-    ut_free(key.iv);
-    ut_free(key.ptr);
-
-    key.iv = nullptr;
-    key.ptr = nullptr;
-
-    key.space_id = std::numeric_limits<space_id_t>::max();
   }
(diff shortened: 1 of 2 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtrabackup?pullRequest=772) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/percona-xtrabackup?pullRequest=772) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/percona-xtrabackup?pullRequest=772) it.</sub>

gl-sergei:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/83/

PR #784, opened 2019-12-13
PXB-1978 xbcloud fails to obtain data from S3 bucket

rahulmalik87:
Problem
-----

xbcloud fail to probe s3 bucket probe-bucket.

Analysis
----

xbcloud is using a hardcoded bucket probe-bucket in s3
which it not able to access.

Fix
---
For probe use bucket provided by option s3-bucket

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/s3.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/s3.cc
@@ -423,7 +423,7 @@ bool S3_client::create_bucket(const std::string &name) {
   return true;
 }
 
-bool S3_client::probe_api_version_and_lookup() {
+bool S3_client::probe_api_version_and_lookup(const std::string &bucket) {
   for (auto lookup : {LOOKUP_DNS, LOOKUP_PATH}) {
     if (bucket_lookup != LOOKUP_AUTO && bucket_lookup != lookup) {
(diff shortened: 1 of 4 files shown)


satya-bodapati (approved):
The test code is still using hardcoded name. Is this unit test? Can you check how to make it work post push of this. May be create a separate JIRA ticket.

Ok to push after QA confirms backup/restore from S3 works.

igorsol (commented):
This fix should help to alleviate the issue since instead of testing hard-coded "probe-bucket" which might not exist now the test is executed with the bucket name provided by the user.
OTOH it looks very suspicious that `bucket_exists` could return true for non-existent "probe-bucket". Does it mean it still can return true for some missing bucket? This can be another issue.
Another question is about calls to `probe_api_version_and_lookup` in the test (xbcloud-t.cc). Looks like return value is ignored there so this test calls are useless if error is not propagated in som eother way.

rahulmalik87:
yes, if it can create a valid connection, `bucket_exists` returns true even for non-existent "bucket".
 for a non-existent bucket, It tries to create a bucket before starting back up. 
I have opened the JIRA to fix xbcloud-t.cc PXB-1980 [https://jira.percona.com/browse/PXB-1980]

PR #801, opened 2020-01-10
PXB-1982 history table showing wrong value for lock_time

rahulmalik87:
PXB-1982 history table showing wrong value for lock_time

    In the backup history table, lock_time is shown as 0 when the database
    is locked and junk value when the database is not locked.

    Fix:
    Update lock_time to 0 if the database is not locked and if the database
    is locked then, time in seconds.
    With upstream, lock_time is the time between FTWRL and UNLOCK TABLES.
    With PS, lock_time is the time between LOCK BINLOG FOR BACKUP and UNLOCK TABLES

### DIFF
--- a/storage/innobase/xtrabackup/test/t/history_on_server.sh
+++ b/storage/innobase/xtrabackup/test/t/history_on_server.sh
@@ -85,6 +85,15 @@ do
     check_for_value "$column" "N"
 done
 
+get_one_value "lock_time"
+lock_time_without_lock=$val
+
+if [ $val -lt 0 ];
+then	
+    vlog "Error: lock_time in history record invalid, expected > 0, got \"$val\""
+    exit 1
+fi
+
 check_for_value "format" "tar"
 
 # saving for later
@@ -102,7 +111,7 @@ vlog "Testing incremental based on history name"
 multi_row_insert incremental_sample.test \({101..200},100\)
 
 innobackupex --history=test1 --incremental \
---incremental-history-name=test1 $backup_dir > /dev/null
+--incremental-history-name=test1 --no-lock $backup_dir > /dev/null
 
 # saving for later
 get_one_value "uuid"
@@ -111,6 +120,8 @@ get_one_value "innodb_from_lsn"
 second_from_lsn=$val
 get_one_value "innodb_to_lsn"
 second_to_lsn=$val
+
+check_for_value "lock_time" "0"
 
 check_for_value "format" "file"
 check_for_value "incremental" "Y"
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1574,9 +1574,9 @@ backup_finish()
 	/* release all locks */
 	if (!opt_no_lock) {
 		unlock_all(mysql_connection);
-		history_lock_time = 0;
-	} else {
 		history_lock_time = time(NULL) - history_lock_time;
+	} else {
+		history_lock_time = 0;
 	}
 
 	if (opt_safe_slave_backup && sql_thread_started) {


storage/innobase/xtrabackup/test/t/history_on_server.sh
```diff
@@ -85,6 +85,15 @@ do
...
 
+get_one_value "lock_time"
+lock_time_without_lock=$val
+
+if [ $val -lt 0 ];
+then	
+    vlog "Error: lock_time in history record invalid, expected > 0, got \"$val\""
+    exit 1
+fi
+
```
satya-bodapati, line 96: Are we testing xtrabckup binary or innobakupex? whats the difference between two

storage/innobase/xtrabackup/test/t/history_on_server.sh
```diff
@@ -102,7 +111,7 @@ vlog "Testing incremental based on history name"
 multi_row_insert incremental_sample.test \({101..200},100\)
 
 innobackupex --history=test1 --incremental \
---incremental-history-name=test1 $backup_dir > /dev/null
+--incremental-history-name=test1 --no-lock $backup_dir > /dev/null
```
satya-bodapati, line 114: better to add separate command which is full backup?

satya-bodapati (changes requested):
Description and/or Commit message is poor.

Can you please improve it?

storage/innobase/xtrabackup/test/t/history_on_server.sh
```diff
@@ -85,6 +85,15 @@ do
     check_for_value "$column" "N"
 done
 
+get_one_value "lock_time"
+lock_time_without_lock=$val
+
+if [ $val -lt 0 ];
```
satya-bodapati, line 91: Less than or equal to 0 is also wrong. Please update check

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/history_on_server.sh
rahulmalik87: innobakupex is a wrapper on the xtrabackup binary.  which is available in 2.4x
 for 8.0, we are using directly xtrabackup

storage/innobase/xtrabackup/test/t/history_on_server.sh
rahulmalik87: below are the results of manual testing with lock options. 

mysql> select * from PERCONA_SCHEMA.xtrabackup_history\G                                 *************************** 1. row ***************************
            uuid: 6e31a43a-3dbb-11ea-9398-002590d40aec
            name: NULL
       tool_name: xtrabackup
    tool_command: --target-dir=/home/rahul.malik/MySQL/data/xo7 --core-file --user=root --socket /tmp/socket_1320.sock --keyring-file-data=/home/rahul.malik/MySQL/data/o7/key.key --xtrabackup-plugin-dir=/home/rahul.malik/MySQL/src/x7/bld/storage/innobase/xtrabackup/src/keyring --history --backup
    tool_version: 2.4.18
ibbackup_version: 2.4.18
  server_version: 5.7.29-debug
      start_time: 2020-01-23 08:34:02
        end_time: 2020-01-23 08:36:20
       lock_time: 136
      binlog_pos: NULL
 innodb_from_lsn: 0
   innodb_to_lsn: 172601168
         partial: N
     incremental: N
          format: file
         compact: N
      compressed: N
       encrypted: N
1 row in set (0.00 sec)

storage/innobase/xtrabackup/test/t/history_on_server.sh
rahulmalik87: there is some values like lsn, UUID which use need initial backup value. and results are same with increment backup or with full backup.

storage/innobase/xtrabackup/test/t/history_on_server.sh
satya-bodapati: ideally mtr testcase would be nice. but since it cannot be tested easily ie getting non zero lock time is not easy, manual result proof is fine.

satya-bodapati (approved):

PR #812, opened 2020-03-06
PXB-2118 Externally created tablespaces are renamed by xtrabackup.

rahulmalik87:
PXB-2118 Externally created tablespaces are getting renamed by xtrabackup.
Analysis:
During backup, Xtrabackup renames the undo tablespace file with undo name
to handle duplicate undo tablespace name.
Fix:
With the new design as part of PXB-2099, PXB restores undo tablespace only
in undo directory or data directory.  It can’t have duplicate tablespace
name since there is a single directory where all undo tablespaces are moved.
Do not rename the undo tablespace during backup and fail if any duplicate file is present.

It will also fix PXB-2107 since now we are no more renaming the
tablespace. Just direct copy to data-dir

### DIFF
--- a/storage/innobase/xtrabackup/test/t/undo_tablespaces.sh
+++ b/storage/innobase/xtrabackup/test/t/undo_tablespaces.sh
@@ -36,9 +36,7 @@ start_server
 
 # create some undo tablespaces
 mysql -e "CREATE UNDO TABLESPACE undo1 ADD DATAFILE '$undo_directory_ext/undo1.ibu'"
-mysql -e "CREATE UNDO TABLESPACE undo2 ADD DATAFILE '$undo_directory_ext/undo2.ibu'"
-mysql -e "CREATE UNDO TABLESPACE undo3 ADD DATAFILE 'undo1.ibu'"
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/test/t/undo_tablespaces.sh
```diff
@@ -36,9 +36,7 @@ start_server
 
 # create some undo tablespaces
 mysql -e "CREATE UNDO TABLESPACE undo1 ADD DATAFILE '$undo_directory_ext/undo1.ibu'"
-mysql -e "CREATE UNDO TABLESPACE undo2 ADD DATAFILE '$undo_directory_ext/undo2.ibu'"
-mysql -e "CREATE UNDO TABLESPACE undo3 ADD DATAFILE 'undo1.ibu'"
-mysql -e "CREATE UNDO TABLESPACE undo4 ADD DATAFILE 'undo2.ibu'"
```
satya-bodapati, line 41: Please keep this scenario as upstream allows it (until it gets fixed) and show that backup detects the conflict and aborts the backup

satya-bodapati (changes requested):

satya-bodapati (approved):
LGTM!

PR #826, opened 2020-03-23
PXB-1818 Provide a new flag --max-memory to improve performance by au…

rahulmalik87:
…to-tuning the value for --use-memory

Performance in PXB is heavily impacted by the memory provided to InnoDB during
"prepare" stages. If memory is less backup may have to swap pages multiple
 times between buffer pool and disk which increases io and slow down prepare.

The new flag "--max-free-memory-per" would allow backup to take up to that
percentage of free memory and allocate it to XtraBackup.
By default, it will auto-scale up to 50% of available memory.
Xtrabackup will start with default memory and based on the processing
it will increase the memory. Maximum free memory used depends on
"--max-free-memory-per" and server available memory.
If the user wants a fixed amount of memory it can be done by providing
"use-memory"

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -85,6 +85,8 @@ bool meb_replay_file_ops = true;
 #endif /* !UNIV_HOTBACKUP */
 
 std::list<space_id_t> recv_encr_ts_list;
+extern longlong xtrabackup_use_memory;
+extern longlong xtrabackup_free_memory_per;
 
 /** Log records are stored in the hash table in chunks at most of this size;
 this must be less than UNIV_PAGE_SIZE as it is stored in the buffer pool */
@@ -3221,6 +3223,7 @@ automatically when the hash table becomes full.
                                 contiguous log data up to this lsn
 @param[out]	read_upto_lsn	scanning succeeded up to this lsn
 @param[in]  to_lsn LSN to stop scanning at
+@param[in]  true if the data is flushed to disk
 @return true if not able to scan any more in this log */
 #ifndef UNIV_HOTBACKUP
 static bool recv_scan_log_recs(log_t &log,
@@ -3230,7 +3233,7 @@ bool meb_scan_log_recs(
                                ulint max_memory, const byte *buf, ulint len,
                                lsn_t checkpoint_lsn, lsn_t start_lsn,
                                lsn_t *contiguous_lsn, lsn_t *read_upto_lsn,
-                               lsn_t to_lsn) {
+                               lsn_t to_lsn, bool *flushed) {
   const byte *log_block = buf;
   lsn_t scanned_lsn = start_lsn;
   bool finished = false;
@@ -3424,7 +3427,6 @@ bool meb_scan_log_recs(
     if (finished || (recv_scan_print_counter % 80) == 0) {
       ib::info(ER_IB_MSG_725, ulonglong{scanned_lsn});
     }
-  }
 
   if (more_data && !recv_sys->found_corrupt_log) {
     /* Try to parse more log records */
@@ -3432,9 +3434,13 @@ bool meb_scan_log_recs(
     recv_parse_log_recs(checkpoint_lsn);
 
 #ifndef UNIV_HOTBACKUP
-    if (recv_heap_used() > max_memory) {
-      recv_apply_hashed_log_recs(log, false);
-    }
+  }
+
+  if (recv_heap_used() > max_memory) {
+    recv_apply_hashed_log_recs(log, false);
+    ut_ad(*flushed == false);
+    *flushed = true;
+  }
 #endif /* !UNIV_HOTBACKUP */
 
     if (recv_sys->recovered_offset > recv_sys->buf_len / 4) {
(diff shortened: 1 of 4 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
     finished = recv_scan_log_recs(log, max_mem, log.buf, RECV_SCAN_SIZE,
                                   checkpoint_lsn, start_lsn, contiguous_lsn,
-                                  &log.scanned_lsn, to_lsn);
+                                  &log.scanned_lsn, to_lsn, &flushed);
+
+    redo_applied += RECV_SCAN_SIZE;
+    if (flushed) {
+      redo_applied_total += redo_applied;
+
+      auto per = ((double)redo_applied / (double)srv_log_file_size) * 100;
```
satya-bodapati, line 3655: s/per/ pct everywhere

storage/innobase/log/log0recv.cc
```diff
@@ -3221,6 +3223,7 @@ automatically when the hash table becomes full.
                                 contiguous log data up to this lsn
 @param[out]	read_upto_lsn	scanning succeeded up to this lsn
 @param[in]  to_lsn LSN to stop scanning at
+@param[in]  true if the data is flushed to disk
```
satya-bodapati, line 3226: in,out and  description as "true if redo records are applied from hash table"

storage/innobase/log/log0recv.cc
```diff
@@ -3424,17 +3427,20 @@ bool meb_scan_log_recs(
...
-      recv_apply_hashed_log_recs(log, false);
-    }
+  }
+
+  if (recv_heap_used() > max_memory) {
+    recv_apply_hashed_log_recs(log, false);
+    ut_ad(*flushed == false);
+    *flushed = true;
+  }
 #endif /* !UNIV_HOTBACKUP */
```
satya-bodapati, line 3444: why this block has to be moved? Is it now now out of more_data && !corrupt check?

storage/innobase/log/log0recv.cc
satya-bodapati: No need to put double in both places.

storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+      redo_applied_total += redo_applied;
+
+      auto per = ((double)redo_applied / (double)srv_log_file_size) * 100;
+      auto total_applied_per =
+          ((double)redo_applied_total / (double)srv_log_file_size) * 100;
 
+      ib::info() << "redo applied in last iteration:" << per
+                 << "%\n total redo applied:" << total_applied_per << "%";
+
+      /* find the availabe memory */
```
satya-bodapati, line 3662: find available free memory

storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+      auto xtrabackup_max_memory =
```
satya-bodapati, line 3663: auto free_mem =

storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+          (ulint)(double(sysconf(_SC_AVPHYS_PAGES) * sysconf(_SC_PAGESIZE) *
+                         xtrabackup_free_memory_per) /
+                  100);
```
satya-bodapati, line 3666: is pxb used and tested only on linux platforms? What about other free memory calcuation for windows, osx etc?

storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+
+      ib::info() << "Maximum memory for buffer pool:" << xtrabackup_max_memory;
+
+      /* gradually increase the buffer pool until it can processed redo log in
+       * one iteration or it hit threshold */
```
satya-bodapati, line 3671: this style of comment not allowed/ recommened

remove the the * on next line and align to the top comment

storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,63 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+      if (xtrabackup_use_memory == 100 * 1024 * 1024L &&
```
satya-bodapati, line 3672: Why this enhancement only if --use-memory is not used?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -156,7 +156,10 @@ bool xtrabackup_print_param = FALSE;
 bool xtrabackup_export = FALSE;
 bool xtrabackup_apply_log_only = FALSE;
 
+
```
satya-bodapati, line 159: remove this

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -156,7 +156,10 @@ bool xtrabackup_print_param = FALSE;
...
 longlong xtrabackup_use_memory = 100 * 1024 * 1024L;
+/* default value max memory is 50% of availabe memory */
+longlong xtrabackup_free_memory_per = 50;
```
satya-bodapati, line 162: /*  Default free memory percentage used is 50 % of available free memory */

satya-bodapati: xtrabackup_ prefix can be skipped IMHO..

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -731,6 +735,10 @@ struct my_option xb_client_options[] = {
      (G_PTR *)&xtrabackup_use_memory, (G_PTR *)&xtrabackup_use_memory, 0,
      GET_LL, REQUIRED_ARG, 100 * 1024 * 1024L, 1024 * 1024L, LLONG_MAX, 0,
      1024 * 1024L, 0},
+    {"max-free-memory-per", OPT_XTRA_MAX_MEMORY,
```
satya-bodapati, line 738: use-free-memory-pct= (max is not required)

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
rahulmalik87: pxb doesn't support windows.

storage/innobase/log/log0recv.cc
rahulmalik87: I had a discussion with Tyler in PXB team meeting. He said we have to do optimization only if       --use memory option is not provided.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: xtrabackup_prefix is used for al other variables.

PR #833, opened 2020-04-06
PXB-2142 transition key information is not masked

rahulmalik87:
PXB-2142 transition key information is not masked

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -2096,6 +2096,14 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
 				strlen("--encrypt_key")) == 0) {
 			arg = "--encrypt_key=...";
 		}
+		if (strncmp(*argv, "--transition-key",
+				strlen("--transition-key")) == 0) {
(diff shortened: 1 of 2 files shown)


satya-bodapati:
> _No description provided._

Please provide description message

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2096,6 +2096,14 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
 				strlen("--encrypt_key")) == 0) {
 			arg = "--encrypt_key=...";
 		}
+		if (strncmp(*argv, "--transition-key",
+				strlen("--transition-key")) == 0) {
```
satya-bodapati, line 2100: strlen() should ideally be aligned to *argv.. but it looks like old code is doing wrong alignment too..

So either fix all strlen alignment or leave it as it is.. 
choice is yours :)

satya-bodapati (approved):
Please fix description in github page

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: I will leave it as it for now.

PR #834, opened 2020-04-06
PXB-2142 transition key information is not masked

rahulmalik87:
PXB-2142 transition key information is not masked


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2096,6 +2096,14 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
...
 			arg = "--encrypt_key=...";
 		}
+		if (strncmp(*argv, "--transition-key",
+				strlen("--transition-key")) == 0) {
+			arg = "--transition-key=...";
+		}
+		if (strncmp(*argv, "--transition_key",
+				strlen("--transition_key")) == 0) {
+			arg = "--transition_key=...";
+		}
```
satya-bodapati: Hmm, PXB 8.0 doesn't follow clang-formatting? If not, please create as separate JIRA ticket for this issue. With clang-formatting we can introduce format check in github.

satya-bodapati (approved):

PR #838, opened 2020-04-08
PXB-2152 new master key is written in standard error output

rahulmalik87:
Problem:
with generate-new-master-key, copy-back is printing the new
encryption-key

Fix
remove the printing of key in error messages


peterschwaller:
Reviewed.  Okay to merge.

PR #851, opened 2020-04-16
PXB-1818 Provide a new flag --max-memory to improve performance by

rahulmalik87:
auto-tunning value for --use-memory

Performance in PXB is heavily impacted by the memory provided to InnoDB during
"prepare" stages. If memory is less backup may have to swap pages multiple
 times between buffer pool and disk which increases io and slow down prepare.

The new flag "--use-free-memory-pct" would allow backup to take up to that
percentage of free memory and allocate it to XtraBackup.
By default, it will auto-scale up to 50% of available memory.
Xtrabackup will start with default memory and based on the processing
it will increase the memory. Maximum free memory used depends on
"--use-free-memory-per" and server available memory.
If the user wants a fixed amount of memory it can be done by providing
"use-memory"

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2055,7 +2055,6 @@ static void buf_pool_resize() {
     mutex_exit(&buf_pool->free_list_mutex);
 
     buf_flush_list_mutex_enter(buf_pool);
-    ut_ad(buf_pool->flush_rbt == NULL);
     buf_flush_list_mutex_exit(buf_pool);
 #endif
 
@@ -2066,25 +2065,6 @@ static void buf_pool_resize() {
         new_instance_size * UNIV_PAGE_SIZE / srv_buf_pool_chunk_unit;
 
     os_wmb;
-  }
-
-  /* disable AHI if needed */
-  bool btr_search_disabled = false;
-
-  buf_resize_status("Disabling adaptive hash index.");
-
-  btr_search_s_lock_all();
-  if (btr_search_enabled) {
-    btr_search_s_unlock_all();
-    btr_search_disabled = true;
-  } else {
-    btr_search_s_unlock_all();
-  }
-
-  btr_search_disable(true);
-
-  if (btr_search_disabled) {
-    ib::info(ER_IB_MSG_60) << "disabled adaptive hash index.";
   }
 
   /* set withdraw target */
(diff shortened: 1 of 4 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -3627,15 +3633,62 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
...
+          ((double)redo_applied_total / (double)srv_log_file_size) * 100;
+
+      ib::info() << "redo applied in last iteration:" << pct
+                 << "%\n total redo applied:" << total_applied_per << "%";
+
+      /* find availabe free memory */
+      auto free_memory =
+          (ulint)(double(sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGESIZE) *
+                         xtrabackup_free_memory_per) /
+                  100);
```
satya-bodapati, line 3667: Please create a separate function for this.

Add code for macos, with ifdefs.
See this for free memory calculation on macos:
https://stackoverflow.com/questions/6094444/how-can-i-programmatically-check-free-system-memory-on-mac-like-the-activity-mon/6095158

storage/innobase/buf/buf0buf.cc
```diff
@@ -2068,25 +2067,6 @@ static void buf_pool_resize() {
...
-    btr_search_disabled = true;
-  } else {
-    btr_search_s_unlock_all();
-  }
-
-  btr_search_disable(true);
-
-  if (btr_search_disabled) {
-    ib::info(ER_IB_MSG_60) << "disabled adaptive hash index.";
-  }
```
satya-bodapati, line 2088: Why is this removed? Is AHI not initialized?

storage/innobase/log/log0recv.cc
```diff
@@ -85,6 +85,8 @@ bool meb_replay_file_ops = true;
 #endif /* !UNIV_HOTBACKUP */
 
 std::list<space_id_t> recv_encr_ts_list;
+extern longlong xtrabackup_use_memory;
+extern longlong xtrabackup_free_memory_per;
```
satya-bodapati, line 89: remove "per" everywhere and  "pct" instead

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -731,6 +734,10 @@ struct my_option xb_client_options[] = {
      (G_PTR *)&xtrabackup_use_memory, (G_PTR *)&xtrabackup_use_memory, 0,
      GET_LL, REQUIRED_ARG, 100 * 1024 * 1024L, 1024 * 1024L, LLONG_MAX, 0,
      1024 * 1024L, 0},
+    {"use-free-memory-pct", OPT_XTRA_MAX_MEMORY,
+     "Maximum free memory percentage of server can be used during prepare ",
```
satya-bodapati, line 738:
```suggestion
     "Percentage of system's free memory for prepare. Default 50%",
```

satya-bodapati (changes requested):

satya-bodapati:
Is it possible to get testcase? Show that the time use auto-scaling with default (50%) is better than auto-scaling off (0%) ?

satya-bodapati:
Also please provide improvement statistics here.

satya-bodapati:
When you create pull request please add your name in 'assignee'. You will get notifications

PR #853, opened 2020-04-24
PXB-1961 Prepare of full backup fails for encrypted tables with Generic

rahulmalik87:
Problem:
PXB is not handling PS encryption types.

Analysis:
PS support addition encrytion types like keyring which is not handled.
This bugs only adresses encryption='n'

Fix:
Handle and process ENCRYPTION_KEY_MAGIC_PS_V2 in PXB


storage/innobase/log/log0recv.cc
```diff
@@ -1727,9 +1727,20 @@ static byte *recv_parse_or_apply_log_rec_body(
...
+          ptr_copy += 2;  // skip offset
+          ulint len = mach_read_from_2(ptr_copy);
+          ptr_copy += 2;
+          if (end_ptr < ptr_copy + len) return NULL;
+          if (memcmp(ptr_copy, ENCRYPTION_KEY_MAGIC_V3,
+                     ENCRYPTION_MAGIC_SIZE) == 0) {
+            return (fil_tablespace_redo_encryption(ptr, end_ptr, space_id));
+          } else if (memcmp(ptr_copy, ENCRYPTION_KEY_MAGIC_PS_V2,
+                            ENCRYPTION_MAGIC_SIZE) == 0) {
+            break;
```
satya-bodapati, line 1697: this works? surprising... the v2 writer is only writing KERYING_ENCRYPTION_INFO_MAX_SIZE bytes.

mlog_write_string(page + offset, encrypt_info,
                    KERYING_ENCRYPTION_INFO_MAX_SIZE, mtr);

Also you need to handle V1 KERYING_ENCRYPTION_INFO_MAX_SIZE_V1 bytes.

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
satya-bodapati: Also there should be ptr null check

rahulmalik87:
update the test case @satya-bodapati. it passed locally. 

Using /home/rahul.malik/MySQL/src/x8/bld/storage/innobase/xtrabackup/test/var as test root
Detecting server version...
Running against MySQL 8.0.21-debug (InnoDB 8.0.21)
Using 8 parallel workers

xbcloud                                  w1	[passed]    106
Spent 106 of 124 seconds executing testcases

{smm-12-d40aec x8 o8 test}$

PR #857, opened 2020-04-29
Added support for session tokens

mikeamzn:
https://jira.percona.com/browse/PXB-2146

I also tested these changes locally with a real s3 bucket/account.


mikeamzn:
Can I get a review here?

PR #862, opened 2020-05-19
PXB-2181: Update info about supported backup targets

borys-belinsky:
This added text will appear in two contexts: the start page and the FAQ page (at the very top)


george-lorch (approved):
Perfect, thank you @borys-belinsky

kathy-williamson (approved):

PR #867, opened 2020-06-02
PXB-2155 : If redo logs are encrypted, xtrabackup --backup --compress=lz4 creates corrupt redo logs

rahulmalik87:
Problem:
-------
If redo logs are encrypted, xtrabackup --backup --compress=lz4  creates
 corrupt redo logs

Analysis:
--------
when we compress the encrypted block, the size of the compressed block can be
greater than the original block. In such cases, we should skip the compression
of the block and write an uncompressed block to the stream otherwise
decompression would complain about this bigger size.

Fix:
---
It involves two fixes.
Disable the compression for the encrypted redo logs.
If the size of the lz4 compressed block is greater than the uncompressed,
skip the compression for that block and write an uncompressed block

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_compress_lz4.cc
+++ b/storage/innobase/xtrabackup/src/ds_compress_lz4.cc
@@ -221,7 +221,10 @@ static int compress_write(ds_file_t *file, const void *buf, size_t len) {
 
     if (error) continue;
 
-    if (thd.to_len > 0) {
+    /* Compressing encrypted or already compressed
+    data the length of compression should exceed, in such case
+    skip the compression */
+    if (thd.to_len > 0 && thd.to_len < COMPRESS_CHUNK_SIZE) {
       /* compressed block length */
       if (write_uint32_le(dest_file, thd.to_len)) {
         error = true;
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -3108,17 +3108,23 @@ static void xtrabackup_init_datasinks(void) {
                        : DS_TYPE_COMPRESS_QUICKLZ);
     xtrabackup_add_datasink(ds);
     ds_set_pipe(ds, ds_data);
-    if (ds_data != ds_redo) {
+
+    /* disable redo compression if redo log is encrypt */
+    if (srv_redo_log_encrypt) {
       ds_data = ds;
-      ds = ds_create(xtrabackup_target_dir,
-                     xtrabackup_compress == XTRABACKUP_COMPRESS_LZ4
-                         ? DS_TYPE_COMPRESS_LZ4
-                         : DS_TYPE_COMPRESS_QUICKLZ);
-      xtrabackup_add_datasink(ds);
-      ds_set_pipe(ds, ds_redo);
-      ds_redo = ds;
     } else {
-      ds_redo = ds_data = ds;
+      if (ds_data != ds_redo) {
+        ds_data = ds;
+        ds = ds_create(xtrabackup_target_dir,
+                       xtrabackup_compress == XTRABACKUP_COMPRESS_LZ4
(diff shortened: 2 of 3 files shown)


satya-bodapati:
> disable compression for encypted redo logs

Although Kathy and PeterS like your JIRA or Github problem "summary", I do not.

Please keep the summary for the problem statement, not a solution.

Also in the description, please explain what caused the lz4 issue.

Fix also is doing two things. Please explain that.

Aslo there is typo in "disable compression for encypted redo logs"  use "encrypted"

storage/innobase/xtrabackup/src/ds_compress_lz4.cc
```diff
@@ -221,7 +221,7 @@ static int compress_write(ds_file_t *file, const void *buf, size_t len) {
 
     if (error) continue;
 
-    if (thd.to_len > 0) {
+    if (thd.to_len > 0 && thd.to_len < COMPRESS_CHUNK_SIZE) {
```
satya-bodapati, line 227: Please explain the code change with a comment

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3108,17 +3108,22 @@ static void xtrabackup_init_datasinks(void) {
                        : DS_TYPE_COMPRESS_QUICKLZ);
     xtrabackup_add_datasink(ds);
     ds_set_pipe(ds, ds_data);
-    if (ds_data != ds_redo) {
+
```
satya-bodapati, line 3111: Add a comment here too

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3108,17 +3108,22 @@ static void xtrabackup_init_datasinks(void) {
...
+    if (srv_redo_log_encrypt)
```
satya-bodapati: What if the variable is turned off after encryption is turned on for a while? Shouldn't we check the redo log header to be sure?

satya-bodapati: use this style for consistency reasons:
```
if (x) {
 blah;
} else {
 blah2;
}
```

storage/innobase/xtrabackup/test/t/innodb_redo_log_encrypt.sh
```diff
@@ -26,6 +26,7 @@ function start_workload()
 	done ) | mysql test
 }
 
+
```
satya-bodapati: remove spurious extra line

storage/innobase/xtrabackup/test/t/innodb_redo_log_encrypt.sh
```diff
@@ -35,6 +36,14 @@ function test_do() {
 	job=$!
 
 	xtrabackup --backup --target-dir=$topdir/backup ${xtra_backup_args}
+
+	# for compresss backup xtrabackup_logfile should not be compress since redo log is encrypt
```
satya-bodapati:
```suggestion
	# for compressed backup, xtrabackup_logfile should not be compressed since redo log is encrypted
```

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: it is an overhead. we have fixed the code of compression.  even if it compressing an encrypted log file it won't cause any trouble.

satya-bodapati: I think, its the other way round. Wrong detection of encrypted redo log would lead un-necessary attempts to compress it.

Since we parse redo log header, we should detect if redo is encrypted or not, based on it. This is accurate.

satya-bodapati (approved):

PR #868, opened 2020-06-02
PXB-2165 new high preference environment variables for xbcloud access…

rahulmalik87:
… to AWS S3

Problem:
AWS_* could be used for some  other general purpose use _*

Fix:
Introduce new Envioment variables with high preference than AWS
S3_ACCESS_KEY_ID,
S3_SECRET_ACCESS_KEY
S3_SESSION_TOKEN
S3_DEFAULT_REGION
S3_CA_BUNDLE
S3_ENDPOINT
like S3_ACCESS_KEY_ID preference would be above AWS_ACCESS_KEY_ID and ACCESS_KEY_ID

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -409,6 +409,16 @@ static void get_env_args() {
   get_env_value(opt_swift_storage_url, "OS_STORAGE_URL");
   get_env_value(opt_cacert, "OS_CACERT");
 
+  /* Below block should always be above AWS_* and should not be moved because
+  the order of prefrence are like S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID and
+  ACCESS_KEY  */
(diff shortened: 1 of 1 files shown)


satya-bodapati:
1. Can you please explain the problem better
2. In the fix, please mention the final order of prefercnes of multiple ways/variables (total 4 I believe)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -409,6 +409,13 @@ static void get_env_args() {
   get_env_value(opt_swift_storage_url, "OS_STORAGE_URL");
   get_env_value(opt_cacert, "OS_CACERT");
 
+  get_env_value(opt_s3_access_key, "S3_ACCESS_KEY_ID");
```
satya-bodapati: add a comment that these should be always above the "AWS_*" and shouldn't be moved because the order of perference 1->2->3.. (subsitute numbers with actual access types)

satya-bodapati (changes requested):

PR #869, opened 2020-06-04
PXB-2164 xbcloud doesn't return the error if the backup doesn't exist…

rahulmalik87:
… in s3 bucket

Probelm:
xbcloud doesn't error out if backup doesn't exist

Fix:
check if backup exists else error out

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -774,7 +774,8 @@ bool xbcloud_download(Object_store *store, const std::string &container,
                       const std::string &backup_name) {
   std::vector<std::string> object_list;
 
-  if (!store->list_objects_in_directory(container, backup_name, object_list)) {
+  if (!store->list_objects_in_directory(container, backup_name, object_list) ||
+      object_list.size() == 0) {
     msg_ts("%s: Download failed. Cannot list %s.\n", my_progname,
            backup_name.c_str());
     return false;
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -774,7 +774,8 @@ bool xbcloud_download(Object_store *store, const std::string &container,
                       const std::string &backup_name) {
   std::vector<std::string> object_list;
 
-  if (!store->list_objects_in_directory(container, backup_name, object_list)) {
+  if (!store->list_objects_in_directory(container, backup_name, object_list) ||
```
satya-bodapati, line 777: so with invalid backupname, list_objects_in_directory doesn't fail (ie doesn't return false?)

rahulmalik87: no it won't return false as long as it has proper access on bucket.

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -101,12 +101,19 @@ sakila/payment.ibd
...
     die 'xbcloud did not exit with error'
 fi
 
-if [ ! grep failed 2>$topdir/pxb-1832.log ] ; then
+if ! grep -q failed $topdir/pxb-1832.log ; then
+    die 'xbcloud did not exit with error'
+fi
+
+#PXB-2164 xbcloud doesn't return the error if the backup doesn't exist in s3 bucket
+xbcloud --defaults-file=$topdir/xbcloud.cnf get somedummyjunkbackup 2>$topdir/pxb-2164.log
```
satya-bodapati, line 114: do u remove these *.log files somewhere?

satya-bodapati (approved):

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: they are removed as part of topdir

PR #874, opened 2020-06-09
PXB-2165 : xbcloud cannot store backups using s3 access key parameters if AWS access key env variables are set

rahulmalik87:
Problem:
--------
If AWS prefix environment variables were set, then xbcloud didn't pick the
other environment variables. for example, If AWS_ACCESS_KEY_ID is set by
some application it did not pick set by ACCESS_KEY which would end up
using the wrong value for s3-access-key and unable to store the backup in
cloud

Analysis:
--------
These AWS prefix credentials are commonly used by other applications as well.
So If a user had set ACCESS_KEY_ID and another application had set
AWS_ACCESS_KEY_ID then xbcloud was pick AWS_ACCESS_KEY_ID and end up
using the wrong credentials.
PXB was giving preference to AWS prefix over other environment variables.

Fix:
---
Introduced new environment variables Prefix with S3 having high preference than AWS*
S3_ACCESS_KEY_ID,
S3_SECRET_ACCESS_KEY
S3_SESSION_TOKEN
S3_DEFAULT_REGION
S3_CA_BUNDLE
S3_ENDPOINT
like S3_ACCESS_KEY_ID preference would be above AWS_ACCESS_KEY_ID and
ACCESS_KEY_ID.So users can set S3 prefix variables and xbcloud would pick them
instead of the variable prefix with AWS.

order of preference:  higher(1) to lower (4)
1) s3 command line like --s3-access-key
2) S3 ENV
3) AWS ENV
4) ACCESS_KEY*

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -408,6 +408,16 @@ static void get_env_args() {
   get_env_value(opt_swift_storage_url, "OS_STORAGE_URL");
   get_env_value(opt_cacert, "OS_CACERT");
 
+  /* Below block should always be above AWS_* and should not be moved because
+  the order of prefrence are like S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID and
+  ACCESS_KEY  */
+  get_env_value(opt_s3_access_key, "S3_ACCESS_KEY_ID");
+  get_env_value(opt_s3_secret_key, "S3_SECRET_ACCESS_KEY");
+  get_env_value(opt_s3_session_token, "S3_SESSION_TOKEN");
+  get_env_value(opt_s3_region, "S3_DEFAULT_REGION");
+  get_env_value(opt_cacert, "S3_CA_BUNDLE");
+  get_env_value(opt_s3_endpoint, "S3_ENDPOINT");
+
   get_env_value(opt_s3_access_key, "AWS_ACCESS_KEY_ID");
   get_env_value(opt_s3_secret_key, "AWS_SECRET_ACCESS_KEY");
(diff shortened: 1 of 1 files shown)


satya-bodapati:
Where is the testcase? Are there any xbcloud testcases?

If it is not possible, I want atleast manual test output result here in the below format

Before fix:
-----------
command and output that indicates error/failure

Afte fix
---------
output that shows success

satya-bodapati:
Problem statement doesn't show the reality. What is the issue that user faces.

Also in fix: Please explain the order of perference for all the aws/s3/token related variables (total 4 I believe)

rahulmalik87:
> Where is the testcase? Are there any xbcloud testcases?
> 
> If it is not possible, I want atleast manual test output result here in the below format
> 
> ## Before fix:
> command and output that indicates error/failure
> 
> ## Afte fix
> output that shows success

before fix
ACCESS_KEY_ID=****
AWS_ACCESS_KEY_ID=wrong
AWS_SECRET_ACCESS_KEY=wrong
SECRET_ACCESS_KEY=***

failed to upload
xtrabackup: Generating a list of tablespaces
InnoDB: Allocated tablespace ID 21 for sys/sys_config, old maximum was 0
200609 07:47:37 [01] Streaming ./ibdata1
xtrabackup: Error writing file 'UNOPENED' (Errcode: 32 - Broken pipe)
xb_stream_write_data() failed.
xtrabackup: Error writing file 'UNOPENED' (Errcode: 32 - Broken pipe)
[01] xtrabackup: Error: xtrabackup_copy_datafile() failed.
[01] xtrabackup: Error: failed to copy datafile.

after fix
AWS_ACCESS_KEY_ID=wrong
AWS_SECRET_ACCESS_KEY=wrong
S3_ACCESS_KEY_ID=****
S3_SECRET_ACCESS_KEY=*****

backup was successful

satya-bodapati (approved):

PR #878, opened 2020-06-10
PXB-2127 xbcloud fails to upload backups with empty database to min.i…

rahulmalik87:
…o storage

Problem:
-----
xbcloud failed to upload backups with an empty database to min.io
It throws an error message: Object name contains unsupported characters.
Analysis:
Xbcloud is placing dummy file db.opt in an empty directory.
Path becomes /path/to//db.opt instead of /path/to/db.opt
minio assume name of the file as “/db.opt” and throws error
unspported file name.
Fix:
Make sure the complete path doesn’t contain // and contains only /.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1396,6 +1396,9 @@ backup_files(const char *from, bool prep_mode)
 		} else if (!prep_mode) {
 			/* backup fake file into empty directory */
 			char path[FN_REFLEN];
+			/* remove trailing / */
+			if(node.filepath[strlen(node.filepath) -1] == '/')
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1396,6 +1396,9 @@ backup_files(const char *from, bool prep_mode)
 		} else if (!prep_mode) {
 			/* backup fake file into empty directory */
 			char path[FN_REFLEN];
+			/* remove trailing / */
+			if(node.filepath[strlen(node.filepath) -1] == '/')
```
satya-bodapati, line 1400:
```suggestion
			if(node.filepath[strlen(node.filepath) - 1] == '/')
```

satya-bodapati (approved):
LGTM!

PR #882, opened 2020-06-11
Adding location constraint to the request body when creating a bucket

jzzaresk:
### Issue
When using a regional endpoint for S3 (specified with --s3-endpoint=...) and non-default region (--s3-region=...), the bucket will fail to be created due to AWS S3's location constraint missing from the request. See S3 API documents [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html#API_CreateBucket_RequestBody).

Example error output:

```
200610 23:55:43 xbcloud_real: Successfully connected. 
200610 23:55:43 xbcloud_real: Failed to create bucket. Error message: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.
```

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/s3.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/s3.cc
@@ -39,6 +39,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 
 namespace xbcloud {
 
+static char *default_s3_region = "us-east-1";
+
 static std::string canonicalize_http_header_value(const std::string &s) {
   std::string r = s;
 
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtrabackup?pullRequest=882) <br/>All committers have signed the CLA.

jzzaresk:
Tested changes with default region, us-east-1, as well as in ap-southeast-2.  Successfully created buckets while streaming backup to S3.

altmannmarcelo:
HI @jzzaresk 

Thanks for our contribution.

I would like to check a few things:

1) Is this still an issue with current versions of s3 api ?
2) If so, can you raise a JIRA isse for this at https://jira.percona.com/projects/PXB/issues/PXB-2534?filter=allopenissues
3) Can you create test case for this patch?

PR #926, opened 2020-08-11
PXB-2229 Error out backup if redo logging is disabled

rahulmalik87:
Problem:
--------
From MySQL 8.0.21 release, redo logs can be disabled and enabled dynamically
using ALTER INSTANCE [ENABLE/DISABLE] REDO_LOG. (This is meant only for initial
data loading stages)

PXB relies on replaying redo logs to get consistent backups.

Fix:
----
Detect whether REDOLOG is disabled or not and error out as follows

1. Redo is disabled before backup                              : Backup failure
2. Redo is disabled and enabled before backup                  : Successful backup
3. Redo is enabled at start of backup, but disabled before end
of backup (applies to both full and incremental backups)     : Backup failure
4. Redo is disabled and enabled again between Full and
Incremental backup                                            : Successful backup

Special case:
-------------

If redo is enabled at startup of backup but during the process of backup it is
disabled and enabled before the backup completes, it will result in inconsistent
backups. PXB may not error out in this scenario.

Users are advised to use --lock-ddl=ON with PXB to avoid users disabling and
enabling redo logs "during the backups"

### DIFF
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -10867,7 +10867,7 @@ ER_XPLUGIN_FAILED_TO_BIND_INTERFACE_ADDRESS
   eng "Value '%s' set to `Mysqlx_bind_address`, X Plugin can't bind to it. Skipping this value."
 
 ER_IB_ERR_RECOVERY_REDO_DISABLED
-  eng "Server was killed when InnoDB redo logging was disabled. Data files could be corrupt. You can try to restart the database with innodb_force_recovery=6"
+  eng "Redo logging is disabled, cannot take consistent backup"
 
 ER_IB_WRN_FAST_SHUTDOWN_REDO_DISABLED
   eng "InnoDB cannot do cold shutdown 'innodb_fast_shutdown = 2' and is forcing 'innodb_fast_shutdown = 1' as redo logging is disabled. InnoDB would flush all dirty pages to ensure physical data consistency."
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -1110,7 +1110,8 @@ bool Redo_Log_Data_Manager::is_error() const { return (error); }
 Redo_Log_Data_Manager::~Redo_Log_Data_Manager() { os_event_destroy(event); }
 
 bool Redo_Log_Data_Manager::stop_at(lsn_t lsn) {
-  if (reader.find_last_checkpoint_lsn(&last_checkpoint_lsn)) {
+  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
+  if (last_checkpoint) {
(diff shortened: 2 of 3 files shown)


storage/innobase/xtrabackup/test/t/alter_redo.sh
```diff
@@ -0,0 +1,147 @@
...
+grep "Redo logging is disabled, cannot take consistent backup" $topdir/inc.log || die "missing error message"
+rm -r $topdir/inc
+rm -r $topdir/full
+rm  $topdir/inc.log
+mysql -e "alter instance enable innodb redo_log"
+
+vlog "case #5 disable disable/enable before normal backup"
+mysql -e "alter instance disable innodb redo_log"
+load_sakila &
+sleep 2 #sleep so some of data is loaded by load_sakila
```
satya-bodapati: with sleep, testcase is not reliable. Just use load_sakila without the &

and add verify

storage/innobase/xtrabackup/test/t/alter_redo.sh
```diff
@@ -0,0 +1,147 @@
...
+rm -r $mysql_datadir
+rm -r $topdir/full
+start_server
+
+vlog "case #6 run normal backup and disable/enable before incremental and parallel dml"
+xtrabackup --backup --target-dir=$topdir/full
+mysql -e "alter instance disable innodb redo_log"
+load_sakila &
+sleep 2 # sleep so some of data is loaded by load_sakila
+mysql -e "alter instance enable innodb redo_log"
```
satya-bodapati: with sleep, testcase is not reliable. Just use load_sakila without the &

and add verify

storage/innobase/xtrabackup/test/t/alter_redo.sh
```diff
@@ -0,0 +1,147 @@
...
+rm -r $mysql_datadir
+rm -r $topdir/full
+rm -r $topdir/inc
+start_server
+
+vlog "case #7 run normal backup and disable/enable before incremental with verify"
+xtrabackup --backup --target-dir=$topdir/full
+mysql -e "alter instance disable innodb redo_log"
+load_sakila
+mysql -e "alter instance enable innodb redo_log"
```
satya-bodapati: This scenario is then duplicate.. may be remove

storage/innobase/xtrabackup/test/t/alter_redo.sh
```diff
@@ -0,0 +1,147 @@
+########################################################################
+#test redo enable/disable feature with backup and incremental backup	
+########################################################################
+. inc/common.sh
+require_server_version_higher_than 8.0.20
+start_server
```
satya-bodapati: rename test to alter_instance_redo.sh

satya-bodapati (changes requested):

rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/lastCompletedBuild/testReport/

satya-bodapati (approved):

PR #927, opened 2020-08-11
PXB-2249 : xtrabackup sh: perl: command not found

satya-bodapati:
Problem:
--------
xtrabackup does version check by default. To do this it uses perl. Can be
skipped with --no-version-check

On kubernetes pods, perl doesn't exist by default and it prints message
"xtrabackup sh: perl: command not found". Note that backup is successful.

Fix:
----
Check if perl binary exists or not before doing version check

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -2337,6 +2337,11 @@ decrypt_decompress()
 void
 version_check()
 {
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtrabackup?pullRequest=927) <br/>All committers have signed the CLA.

rahulmalik87 (approved):
LGTM

PR #928, opened 2020-08-12
PXB-2215 Importing tablespace where character set of table is specifi…

rahulmalik87:
…ed fails with (Column x precise type mismatch.)

Problem:
--------
PXB doesn't calculate pack length of the varchar columns correctly. This is
because it relies on TABLE*->field->field_length. PXB can rely only on dd::Table* 
(created by SDI to dd::Table* conversion) and doesn't have TABLE* initialized.
So TABLE*->field->field_length would give wrong values

Fix:
---
calculate the pack_length based on dd::Column* field length

Thanks to luqun@fb.com for providing this patch

### DIFF
--- a/storage/innobase/dict/dict0dd.cc
+++ b/storage/innobase/dict/dict0dd.cc
@@ -679,9 +679,10 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
 
     long_true_varchar = 0;
     if (field_type == MYSQL_TYPE_VARCHAR) {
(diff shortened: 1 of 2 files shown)


satya-bodapati (changes requested):
Please fix the typos in the commit message "lenght".

and Jenkins link

rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/146/

satya-bodapati (approved):
LGTM!

it-percona:
Comment to trigger webhook.

PR #934, opened 2020-08-19
2.4 Implemented PXB-2254 (Redesign --lock-ddl-per-table)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2254

The current implementation of --lock-ddl-per-table works as follow:

 - Creates a dedicated mysql connection for MDL and start a transaction.
 - Start to scan directories (databases) looking for .ibd files.
 - For each .ibd file it tries to transalate the space id recorded on
the file header to a table (or tables in case of shared tablespace).
 - If it finds a table associate with the .ibd (or tables) it executes a
SELECT * FROM table LIMIT 1 using the dedicated connection for MDL.
 - From this point forward, the table is locked and new DDL will be
blocked until the MDL connection commits.
 - With this, it has been assumed that if a MLOG_LOAD_INDEX event has
been generated (https://dev.mysql.com/worklog/task/?id=7277) it's safe
to continue since we have taken MDL on the table we are copying.

This design has a few flaws:

 1. FTS have their own tablespace, which is been copied without
MDL protection.
 2. Ongoing DDL that generate temporary table (#sql...) are copied
without MDL protection.
 3. When creating FTS for the first time, MySQL will have to create a
tmp table (Item 2) and FTS files (Item 1) are pointing to temp table.
 4. Table space IDs for FTS or regular tables will change if you
DROP and then CREATE the same FTS/TABLE. MDL lock query will fail to
find the corresponding table but file will exist, thus will be copied
without MDL protection.
 5. If a table is created after xtrabackup has gathered the list of
.ibd's to copy and before the backup has finished, this particular
table will be part of the backup as it will be contained within the
redo log. However, if a bulk index load operations have been performed
( MLOG_INDEX_LOAD has been written to the redo log), this change to the
data file will not be tracked and will be missed from the backup.

This commit changes to:

1. MDL is taken at the beginning of backup for all know tables at
that time, before we start to copy individual .ibd files.
2. A new flag controls whatever MDL step has been completed or not.
3. Before MDL from step 1 is completed, redo follow thread accept
parsing MLOG_INDEX_LOAD. At this point is still safe to accept it since
we have not started to copy .ibd files. After MDL part is completed,
mdl_taken changes to true and parsing a MLOG_INDEX_LOAD causes the
backup to abort.
4. Before we execute the MDL query for a specific entry from
INNODB_SYS_TABLES, we validate it against a regexp for a FTS file name,
if we are manipulating a FTS index, we translate it to the corresponding
table name.
5. If for any reason we cannot execute the MDL lock query (for example,
a DDL is ongoing and we are manipulating a file starting with #sql-)
we do not abort the backup. We will only abort if a subsequent
MLOG_INDEX_LOAD is found by the redo follow thread during the backup.
6. We identify temporary tables (#sql-) and skip trying to lock them.
No reason to attempt to query it as it will always result in an error.
6. MDL query changed to not bring any row back. This is enough for
taking a MDL while saving the server from bringing any row back.

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -1793,7 +1793,7 @@ recv_parse_or_apply_log_rec_body(
 		of offline backup and continue.
 		*/
 		if (!recv_recovery_on) {
-			if (!opt_lock_ddl_per_table) {
+			if (mdl_taken) {
 				if (backup_redo_log_flushed_lsn
 				    < recv_sys->recovered_lsn) {
 					ib::info() << "Last flushed lsn: "
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -54,6 +54,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "backup_copy.h"
 #include "backup_mysql.h"
 #include "mysqld.h"
+#include "xb_regex.h"
 
 
 char *tool_name;
@@ -1197,12 +1198,6 @@ the locking strategy being used */
 void
 unlock_all(MYSQL *connection)
 {
-	if (opt_debug_sleep_before_unlock) {
-		msg_ts("Debug sleep for %u seconds\n",
-		       opt_debug_sleep_before_unlock);
-		os_thread_sleep(opt_debug_sleep_before_unlock * 1000);
-	}
-
 	if (binlog_locked) {
 		msg_ts("Executing UNLOCK BINLOG\n");
 		xb_mysql_query(connection, "UNLOCK BINLOG", false);
@@ -2163,69 +2158,102 @@ backup_cleanup()
 	}
 }
 
-static ib_mutex_t mdl_lock_con_mutex;
 static MYSQL *mdl_con = NULL;
-
 void
-mdl_lock_init()
+mdl_lock_tables()
 {
-	mutex_create(LATCH_ID_XTRA_DATAFILES_ITER_MUTEX, &mdl_lock_con_mutex);
-
-	mdl_con = xb_mysql_connect();
-
-	if (mdl_con != NULL) {
-		xb_mysql_query(mdl_con, "BEGIN", false, true);
-	}
-}
-
-
-void
-mdl_lock_table(ulint space_id)
-{
+	msg_ts("Initializing MDL on all current tables.\n");
 	MYSQL_RES *mysql_result = NULL;
 	MYSQL_ROW row;
-	char *query;
+	mdl_con = xb_mysql_connect();
+	if (mdl_con != NULL) {
+		xb_mysql_query(mdl_con, "BEGIN", false, true);
+		mysql_result = xb_mysql_query(mdl_con, "SELECT NAME, SPACE FROM "
+									"INFORMATION_SCHEMA.INNODB_SYS_TABLES", true, true);
+		while ((row = mysql_fetch_row(mysql_result))) {
+			if (fil_is_user_tablespace_id(atoi(row[1]))) {
+				char full_table_name[MAX_FULL_NAME_LEN + 1];
+				innobase_format_name(full_table_name, sizeof(full_table_name), row[0]);
+				if (is_fts_index(full_table_name)) {
+					// We will eventually get to the row to lock the main table
+					msg_ts("%s is a Full-Text Index. Skipping\n", full_table_name);
+					continue;
+				} else if (is_tmp_table(full_table_name)) {
+					// We cannot run SELECT ... #sql-; Skipped to avoid invalid query.
+					msg_ts("%s is a temporary table. Skipping\n", full_table_name);
+					continue;
+				}
 
-	mutex_enter(&mdl_lock_con_mutex);
+				msg_ts("Locking MDL for %s\n", full_table_name);
+				char *lock_query;
+				xb_a(asprintf(&lock_query,
+						"SELECT 1 FROM %s LIMIT 0", full_table_name));
 
-	xb_a(asprintf(&query,
-		"SELECT NAME FROM INFORMATION_SCHEMA.INNODB_SYS_TABLES "
-		"WHERE SPACE = %lu", space_id));
+				xb_mysql_query(mdl_con, lock_query, false, false);
 
-	mysql_result = xb_mysql_query(mdl_con, query, true);
-
-	while ((row = mysql_fetch_row(mysql_result))) {
-		char *lock_query;
-		char table_name[MAX_FULL_NAME_LEN + 1];
-
-		innobase_format_name(table_name, sizeof(table_name), row[0]);
-
-		msg_ts("Locking MDL for %s\n", table_name);
-
-		xb_a(asprintf(&lock_query,
-			"SELECT * FROM %s LIMIT 1",
-			table_name));
-
-		xb_mysql_query(mdl_con, lock_query, false, false);
-
-		free(lock_query);
+				free(lock_query);
+			}
+		}
+		mysql_free_result(mysql_result);
 	}
-
-	mysql_free_result(mysql_result);
-	free(query);
-
-	mutex_exit(&mdl_lock_con_mutex);
 }
-
 
 void
 mdl_unlock_all()
 {
-	msg_ts("Unlocking MDL for all tables");
+	msg_ts("Unlocking MDL for all tables\n");
+	if (mdl_con != NULL) {
+		xb_mysql_query(mdl_con, "COMMIT", false, true);
+		mysql_close(mdl_con);
+	}
+}
+bool
+is_fts_index(const std::string &tablespace)
+{
+	const char *pattern =
+			"^(FTS|fts)_[0-9a-fA-f]{16}_(([0-9a-fA-]{16}_(INDEX|index)_[1-6])|"
+			"DELETED_CACHE|deleted_cache|DELETED|deleted|CONFIG|config|BEING_DELETED|"
+			"being_deleted|BEING_DELETED_CACHE|beign_deleted_cache)$";
+	const char *error_context = "is_fts_index";
+	return check_regexp_table_name(tablespace, error_context, pattern);
+}
 
-	xb_mysql_query(mdl_con, "COMMIT", false, true);
+bool
+is_tmp_table(const std::string &tablespace)
+{
+	const char* pattern = "^#sql-";
+	const char* error_context = "is_tmp_table";
+	return check_regexp_table_name(tablespace, error_context, pattern);
+}
 
-	mutex_free(&mdl_lock_con_mutex);
+bool
+check_regexp_table_name(
+	std::string tablespace,
+	const char* error_context,
+	const char* pattern
+)
+{
+	bool result = false;
+	get_table_name_from_tablespace(tablespace);
+	xb_regex_t preg;
+	size_t nmatch = 1;
+	xb_regmatch_t pmatch[1];
+	compile_regex(pattern, error_context, &preg);
+
+	if (xb_regexec(&preg, tablespace.c_str(), nmatch, pmatch, 0) != REG_NOMATCH) {
+		result = true;
+	}
+	xb_regfree(&preg);
+	return result;
+}
+void
+get_table_name_from_tablespace(std::string &tablespace)
+{
+	std::size_t pos = tablespace.find("`.`"); //`db`.`table` separator
+	if (pos != std::string::npos) {
+		tablespace = tablespace.substr(pos+3);
+		tablespace.erase(tablespace.size()-1); //remove leading `
+	}
 }
 
 bool
--- a/storage/innobase/xtrabackup/src/backup_mysql.h
+++ b/storage/innobase/xtrabackup/src/backup_mysql.h
@@ -104,11 +104,44 @@ write_slave_info(MYSQL *connection);
 void
 parse_show_engine_innodb_status(MYSQL *connection);
 
+/**************************************************************//**
+Acquire MDL lock on all tables*/
 void
-mdl_lock_init();
+mdl_lock_tables();
 
+/**************************************************************//**
+Identifies if tablespace is a Full Text Index
+@return TRUE if is a FTS, or FALSE otherwise */
+bool
+is_fts_index(
+	const std::string &tablespace 	/*!< in: tablespace */
+);
+
+/**************************************************************//**
+Identifies if tablespace is a temporary table (#sql-)
+@return TRUE if is a temporary table, or FALSE otherwise */
+bool
+is_tmp_table(
+	const std::string &tablespace 	/*!< in: tablespace */
+);
+
+/**************************************************************//**
+Runs a regexp against a table name
+@return TRUE if there is a match, or FALSE otherwise */
+bool
+check_regexp_table_name(
+	std::string tablespace,	/*!< in: tablespace */
+	const char* error_context,	/*!< in: error to be return in case of errors */
+	const char* pattern	/*!< in: regexp pattern to try a match */
+);
+
+/**************************************************************//**
+Extract the table name from a full qualified `db`.`table` string 
+removing escape string. Replace the name inplace `*/
 void
-mdl_lock_table(ulint space_id);
+get_table_name_from_tablespace(
+	std::string &tablespace 	/*!< in/out: tablespace */
+);
 
 void
 mdl_unlock_all();
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -471,6 +471,7 @@ ulint redo_log_version = REDO_LOG_V1;
 ulint opt_encrypt_server_id = 0;
 bool opt_encrypt_for_server_id_specified = false;
 
+bool mdl_taken = false;
 #define CLIENT_WARN_DEPRECATED(opt, new_opt) \
   msg("WARNING: " opt \
       " is deprecated and will be removed in a future version. " \
@@ -929,7 +930,7 @@ struct my_option xb_client_options[] =
    REQUIRED_ARG, 31536000, 1, 31536000, 0, 1, 0},
 
   {"lock-ddl-per-table", OPT_LOCK_DDL_PER_TABLE, "Lock DDL for each table "
-   "before xtrabackup starts to copy it and until the backup is completed.",
+   "before xtrabackup starts the copy phase and until the backup is completed.",
    (uchar*) &opt_lock_ddl_per_table, (uchar*) &opt_lock_ddl_per_table, 0,
    GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
@@ -2786,10 +2787,6 @@ xtrabackup_copy_datafile(fil_node_t* node, uint thread_n)
 
 	is_system = !fil_is_user_tablespace_id(node->space->id);
 
-	if (!is_system && opt_lock_ddl_per_table) {
-		mdl_lock_table(node->space->id);
-	}
-
 	if (!is_system && check_if_skip_table(node_name)) {
 		msg("[%02u] Skipping %s.\n", thread_n, node_name);
 		return(FALSE);
@@ -4183,7 +4180,6 @@ xb_register_table(
 	xb_register_include_filter_entry(name);
 }
 
-static
 bool compile_regex(
 	const char* regex_string,
 	const char* error_context,
@@ -4917,10 +4913,20 @@ reread_log_header:
 	xtrabackup_choose_lsn_offset(checkpoint_lsn_start);
 	mutex_exit(&log_sys->mutex);
 
+	if (opt_lock_ddl_per_table) {
+		mdl_lock_tables();
+	}
+
 	/* copy log file by current position */
 	if(xtrabackup_copy_logfile(checkpoint_lsn_start, FALSE))
 		exit(EXIT_FAILURE);
 
+	/*
+	* From this point forward, recv_parse_or_apply_log_rec_body should fail if
+	* MLOG_INDEX_LOAD event is parsed as its not safe to continue the backup
+	* in any situation (with or without --lock-ddl-per-table).
+	*/
+	mdl_taken = true;
 
 	log_copying_stop = os_event_create("log_copying_stop");
 	os_thread_create(log_copying_thread, NULL, &log_copying_thread_id);
@@ -4959,10 +4965,6 @@ reread_log_header:
 	if (xtrabackup_parallel > 1) {
 		msg("xtrabackup: Starting %u threads for parallel data "
 		    "files transfer\n", xtrabackup_parallel);
-	}
-
-	if (opt_lock_ddl_per_table) {
-		mdl_lock_init();
 	}
 
 	it = datafiles_iter_new(f_system);
@@ -5015,7 +5017,11 @@ reread_log_header:
 	if (!backup_start()) {
 		exit(EXIT_FAILURE);
 	}
-
+	if(opt_lock_ddl_per_table && opt_debug_sleep_before_unlock){
+		msg_ts("Debug sleep for %u seconds\n",
+		       opt_debug_sleep_before_unlock);
+		os_thread_sleep(opt_debug_sleep_before_unlock * 1000000);
+	}
 	/* read the latest checkpoint lsn */
 	latest_cp = 0;
 	{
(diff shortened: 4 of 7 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param/119/

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,137 @@ backup_cleanup()
...
+mdl_lock_tables(MYSQL *connection)
 {
-	mutex_create(LATCH_ID_XTRA_DATAFILES_ITER_MUTEX, &mdl_lock_con_mutex);
-
+	msg_ts("Initializing MDL on all current tables.\n");
+	MYSQL_RES *mysql_result = NULL;
+	MYSQL_ROW row;
 	mdl_con = xb_mysql_connect();
-
+	mdl_fts_con = xb_mysql_connect();
```
satya-bodapati: Do we need separate mdl_fts_conn?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,137 @@ backup_cleanup()
...
-	while ((row = mysql_fetch_row(mysql_result))) {
-		char *lock_query;
-		char table_name[MAX_FULL_NAME_LEN + 1];
-
-		innobase_format_name(table_name, sizeof(table_name), row[0]);
+		mysql_result = xb_mysql_query(mdl_con, "SELECT NAME, SPACE FROM "
+									"INFORMATION_SCHEMA.INNODB_SYS_TABLES", true, true);
+		while ((row = mysql_fetch_row(mysql_result))) {
+			if (fil_is_user_tablespace_id(atoi(row[1]))) {
+				char *lock_query;
```
satya-bodapati: move  declaration close to the place the first usage.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4918,6 +4915,7 @@ xtrabackup_backup_func(void)
 	mutex_exit(&log_sys->mutex);
 
 	/* copy log file by current position */
+	mutex_create(LATCH_ID_XTRA_MDL_TAKEN_MUTEX, &mdl_taken_mutex);
 	if(xtrabackup_copy_logfile(checkpoint_lsn_start, FALSE))
 		exit(EXIT_FAILURE);
```
satya-bodapati: May be it is better if we do MDL locking before we call xb_load_tablesapces() [which determines the list of files to copy]

This will also help to get rid of this mutex and syncronization with redo thread.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: @satya-bodapati when we try to translate FTS table, we still have a resultset open on mdl_con
```
		mysql_result = xb_mysql_query(mdl_con, "SELECT NAME, SPACE FROM "
									"INFORMATION_SCHEMA.INNODB_SYS_TABLES", true, true);
		while ((row = mysql_fetch_row(mysql_result))) { <-- result set iterator
			if (fil_is_user_tablespace_id(atoi(row[1]))) {
				char *lock_query;
				char full_table_name[MAX_FULL_NAME_LEN + 1];
				innobase_format_name(full_table_name, sizeof(full_table_name), row[0]);
				if (is_fts_index(full_table_name)) { <-- new query is executed while resultset still open
					msg_ts("%s is a Full-Text Index. Translating table name\n",
									full_table_name);
					fts_get_table_name(full_table_name);
				}
 . . .
		}
		mysql_free_result(mysql_result); <-- closing resultset

Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/backup_mysql.cc]: After offline discussion, we decided that there is no point of translating the FTS index to base table to lock the base table since we will eventually pick up the base table from INNODB_SYS_TABLES.
Will remove mdl_fts_con

**Comment by it-percona** [comment]
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtrabackup?pullRequest=934) <br/>All committers have signed the CLA.

File: storage/innobase/xtrabackup/src/xtrabackup.cc (hunk 3 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/xtrabackup.cc]: @satya-bodapati - Done. Now we:

1. Execute MDL on all tables via `mdl_lock_tables` 
2. Do the first redo log copy phase, so we are caught up with everything that happened before the backup start
3. Change `mdl_taken` to `true`
4. Start `log_copying_thread` - this thread should abort the backup if it finds `MLOG_INDEX_LOAD`

Let me know if this will be sufficient or if we want to go with the approach of validating each scan SN

File: storage/innobase/xtrabackup/src/backup_mysql.cc [hunk 4]
```diff
@@ -2163,69 +2164,99 @@ backup_cleanup()
...
+	msg_ts("Unlocking MDL for all tables\n");
+	if (mdl_con != NULL) {
+		xb_mysql_query(mdl_con, "COMMIT", false, true);
+		mysql_close(mdl_con);
+	}
+}
+bool
+is_fts_index(std::string tablespace)
+{
+	const char* pattern = "^FTS_[0-9a-fA-f]{16}_([0-9a-fA-f]{16}_INDEX_[1-6]|DELETED_CACHE|DELETED|CONFIG|BEING_DELETED|BEING_DELETED_CACHE)$";
```
satya-bodapati: will it do case insensitive search?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,99 @@ backup_cleanup()
...
 {
-	msg_ts("Unlocking MDL for all tables");
+	msg_ts("Unlocking MDL for all tables\n");
+	if (mdl_con != NULL) {
+		xb_mysql_query(mdl_con, "COMMIT", false, true);
+		mysql_close(mdl_con);
+	}
+}
+bool
+is_fts_index(std::string tablespace)
```
satya-bodapati: can we pass by reference? and also const if we don't change the string

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,99 @@ backup_cleanup()
...
 void
 mdl_unlock_all()
 {
-	msg_ts("Unlocking MDL for all tables");
+	msg_ts("Unlocking MDL for all tables\n");
+	if (mdl_con != NULL) {
+		xb_mysql_query(mdl_con, "COMMIT", false, true);
+		mysql_close(mdl_con);
+	}
+}
```
satya-bodapati, line 2209: what happens if the backup is killed just before it reaches here? (I know it is an existing issue)?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,99 @@ backup_cleanup()
...
-
-		msg_ts("Locking MDL for %s\n", table_name);
+		mysql_result = xb_mysql_query(mdl_con, "SELECT NAME, SPACE FROM "
+									"INFORMATION_SCHEMA.INNODB_SYS_TABLES", true, true);
+		while ((row = mysql_fetch_row(mysql_result))) {
+			if (fil_is_user_tablespace_id(atoi(row[1]))) {
+				char full_table_name[MAX_FULL_NAME_LEN + 1];
+				innobase_format_name(full_table_name, sizeof(full_table_name), row[0]);
+				if (is_fts_index(full_table_name)) {
+					//We will eventually get to the row to lock the main table
```
satya-bodapati: space after comment

satya-bodapati:
```suggestion
					// We will eventually get to the row to lock the main table
```

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2163,69 +2164,99 @@ backup_cleanup()
...
+					msg_ts("%s is a Full-Text Index. Skipping\n", full_table_name);
+					continue;
+				} else if (is_tmp_table(full_table_name)) {
+					//We cannot run SELECT ... #sql-; Skipped to avoid invalid query.
```
satya-bodapati:
```suggestion
					// We cannot run SELECT ... #sql-; Skipped to avoid invalid query.
```

storage/innobase/xtrabackup/src/backup_mysql.h
```diff
@@ -105,10 +105,23 @@ void
...
+
+bool
+is_tmp_table(std::string tablespace);
+
+bool
+check_regexp_table_name(
+	std::string tablespace,
+	const char* error_context,
+	const char* pattern
+);
```
satya-bodapati: Please add function comments. (atleast in .h, optional for .cc files)

satya-bodapati (changes requested):

satya-bodapati (commented):
Approach looks correct. Some minor comments posted.

Also please test with foreign keys on table. Specially if we lock child and then parent.. while server locks parent->child.

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: FTS|DELETED_CACHE|DELETED|CONFIG|BEING_DELETED|BEING_DELETED_CACHE us always upper case
for the hexa part, I've added both `[0-9a-fA-f]`

satya-bodapati: on lower case file systems, all these would be lower case

altmannmarcelo: good point, will fix

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: this is good enough

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: Connection is abruptly closed and rollback is performed by the server.

altmannmarcelo:
@satya-bodapati 
> Also please test with foreign keys on table. Specially if we lock child and then parent.. while server locks parent->child.

MDL its only taken on single table for FK, so there won't be any issue with cascade locking:

```
mysql> SHOW CREATE TABLE parent;
+--------+--------------------------------------------------------------------------------------------------------------+
| Table  | Create Table                                                                                                 |
+--------+--------------------------------------------------------------------------------------------------------------+
| parent | CREATE TABLE `parent` (
  `id` int(11) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
+--------+--------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> SHOW CREATE TABLE child;
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                 |
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| child | CREATE TABLE `child` (
  `id` int(11) DEFAULT NULL,
  `parent_id` int(11) DEFAULT NULL,
  KEY `my_fk` (`parent_id`),
  CONSTRAINT `child_ibfk_1` FOREIGN KEY (`parent_id`) REFERENCES `parent` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

# MDL check
START TRANSACTION; SELECT * FROM parent; select * from performance_schema.metadata_locks\G
Query OK, 0 rows affected (0.00 sec)

+----+
| id |
+----+
|  1 |
|  2 |
|  3 |
|  4 |
|  5 |
+----+
5 rows in set (0.00 sec)

*************************** 1. row ***************************
          OBJECT_TYPE: TABLE
        OBJECT_SCHEMA: test
          OBJECT_NAME: parent
OBJECT_INSTANCE_BEGIN: 139928894600224
            LOCK_TYPE: SHARED_READ
        LOCK_DURATION: TRANSACTION
          LOCK_STATUS: GRANTED
               SOURCE: 
      OWNER_THREAD_ID: 56
       OWNER_EVENT_ID: 46
*************************** 2. row ***************************
          OBJECT_TYPE: TABLE
        OBJECT_SCHEMA: performance_schema
          OBJECT_NAME: metadata_locks
OBJECT_INSTANCE_BEGIN: 139928894600128
            LOCK_TYPE: SHARED_READ
        LOCK_DURATION: TRANSACTION
          LOCK_STATUS: GRANTED
               SOURCE: 
      OWNER_THREAD_ID: 56
       OWNER_EVENT_ID: 47
2 rows in set (0.00 sec)

START TRANSACTION; SELECT * FROM child; select * from performance_schema.metadata_locks\G
Query OK, 0 rows affected (0.00 sec)

+------+-----------+
| id   | parent_id |
+------+-----------+
|    1 |         1 |
|    2 |         2 |
|    3 |         3 |
+------+-----------+
3 rows in set (0.00 sec)

*************************** 1. row ***************************
          OBJECT_TYPE: TABLE
        OBJECT_SCHEMA: test
          OBJECT_NAME: child
OBJECT_INSTANCE_BEGIN: 139928894600224
            LOCK_TYPE: SHARED_READ
        LOCK_DURATION: TRANSACTION
          LOCK_STATUS: GRANTED
               SOURCE: 
      OWNER_THREAD_ID: 56
       OWNER_EVENT_ID: 49
*************************** 2. row ***************************
          OBJECT_TYPE: TABLE
        OBJECT_SCHEMA: performance_schema
          OBJECT_NAME: metadata_locks
OBJECT_INSTANCE_BEGIN: 139928894883968
            LOCK_TYPE: SHARED_READ
        LOCK_DURATION: TRANSACTION
          LOCK_STATUS: GRANTED
               SOURCE: 
      OWNER_THREAD_ID: 56
       OWNER_EVENT_ID: 50
2 rows in set (0.00 sec)

#test 1- MDL child, try to add index on parent:
session 1> START TRANSACTION; SELECT * FROM child;
Query OK, 0 rows affected (0.00 sec)

+------+-----------+
| id   | parent_id |
+------+-----------+
|    1 |         1 |
|    2 |         2 |
|    3 |         3 |
+------+-----------+
3 rows in set (0.00 sec)

session 2> ALTER TABLE parent ADD KEY (id);
Query OK, 0 rows affected (0.01 sec)
Records: 0  Duplicates: 0  Warnings: 0

#Succeed even with MDL lock on child

# test 2 - MDL on parent, try to add index on child
session 1> START TRANSACTION; SELECT * FROM parent;
Query OK, 0 rows affected (0.00 sec)

+----+
| id |
+----+
|  1 |
|  2 |
|  3 |
|  4 |
|  5 |
+----+
5 rows in set (0.00 sec)

session 2> ALTER TABLE child ADD KEY (id);
Query OK, 0 rows affected (0.01 sec)
Records: 0  Duplicates: 0  Warnings: 0

#Succeed even with MDL lock on parent
```

satya-bodapati:
> @satya-bodapati
> 
> > Also please test with foreign keys on table. Specially if we lock child and then parent.. while server locks parent->child.
> 
> MDL its only taken on single table for FK, so there won't be any issue with cascade locking:
> 
> ```
> mysql> SHOW CREATE TABLE parent;
> +--------+--------------------------------------------------------------------------------------------------------------+
> | Table  | Create Table                                                                                                 |
> +--------+--------------------------------------------------------------------------------------------------------------+
> | parent | CREATE TABLE `parent` (
>   `id` int(11) NOT NULL,
>   PRIMARY KEY (`id`)
> ) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
> +--------+--------------------------------------------------------------------------------------------------------------+
> 1 row in set (0.00 sec)
> 
> mysql> SHOW CREATE TABLE child;
> +-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> | Table | Create Table                                                                                                                                                                                                                                 |
> +-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> | child | CREATE TABLE `child` (
>   `id` int(11) DEFAULT NULL,
>   `parent_id` int(11) DEFAULT NULL,
>   KEY `my_fk` (`parent_id`),
>   CONSTRAINT `child_ibfk_1` FOREIGN KEY (`parent_id`) REFERENCES `parent` (`id`)
> ) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
> +-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> 1 row in set (0.00 sec)
> 
> # MDL check
> START TRANSACTION; SELECT * FROM parent; select * from performance_schema.metadata_locks\G
> Query OK, 0 rows affected (0.00 sec)
> 
> +----+
> | id |
> +----+
> |  1 |
> |  2 |
> |  3 |
> |  4 |
> |  5 |
> +----+
> 5 rows in set (0.00 sec)
> 
> *************************** 1. row ***************************
>           OBJECT_TYPE: TABLE
>         OBJECT_SCHEMA: test
>           OBJECT_NAME: parent
> OBJECT_INSTANCE_BEGIN: 139928894600224
>             LOCK_TYPE: SHARED_READ
>         LOCK_DURATION: TRANSACTION
>           LOCK_STATUS: GRANTED
>                SOURCE: 
>       OWNER_THREAD_ID: 56
>        OWNER_EVENT_ID: 46
> *************************** 2. row ***************************
>           OBJECT_TYPE: TABLE
>         OBJECT_SCHEMA: performance_schema
>           OBJECT_NAME: metadata_locks
> OBJECT_INSTANCE_BEGIN: 139928894600128
>             LOCK_TYPE: SHARED_READ
>         LOCK_DURATION: TRANSACTION
>           LOCK_STATUS: GRANTED
>                SOURCE: 
>       OWNER_THREAD_ID: 56
>        OWNER_EVENT_ID: 47
> 2 rows in set (0.00 sec)
> 
> START TRANSACTION; SELECT * FROM child; select * from performance_schema.metadata_locks\G
> Query OK, 0 rows affected (0.00 sec)
> 
> +------+-----------+
> | id   | parent_id |
> +------+-----------+
> |    1 |         1 |
> |    2 |         2 |
> |    3 |         3 |
> +------+-----------+
> 3 rows in set (0.00 sec)
> 
> *************************** 1. row ***************************
>           OBJECT_TYPE: TABLE
>         OBJECT_SCHEMA: test
>           OBJECT_NAME: child
> OBJECT_INSTANCE_BEGIN: 139928894600224
>             LOCK_TYPE: SHARED_READ
>         LOCK_DURATION: TRANSACTION
>           LOCK_STATUS: GRANTED
>                SOURCE: 
>       OWNER_THREAD_ID: 56
>        OWNER_EVENT_ID: 49
> *************************** 2. row ***************************
>           OBJECT_TYPE: TABLE
>         OBJECT_SCHEMA: performance_schema
>           OBJECT_NAME: metadata_locks
> OBJECT_INSTANCE_BEGIN: 139928894883968
>             LOCK_TYPE: SHARED_READ
>         LOCK_DURATION: TRANSACTION
>           LOCK_STATUS: GRANTED
>                SOURCE: 
>       OWNER_THREAD_ID: 56
>        OWNER_EVENT_ID: 50
> 2 rows in set (0.00 sec)
> 
> #test 1- MDL child, try to add index on parent:
> session 1> START TRANSACTION; SELECT * FROM child;
> Query OK, 0 rows affected (0.00 sec)
> 
> +------+-----------+
> | id   | parent_id |
> +------+-----------+
> |    1 |         1 |
> |    2 |         2 |
> |    3 |         3 |
> +------+-----------+
> 3 rows in set (0.00 sec)
> 
> session 2> ALTER TABLE parent ADD KEY (id);
> Query OK, 0 rows affected (0.01 sec)
> Records: 0  Duplicates: 0  Warnings: 0
> 
> #Succeed even with MDL lock on child
> 
> # test 2 - MDL on parent, try to add index on child
> session 1> START TRANSACTION; SELECT * FROM parent;
> Query OK, 0 rows affected (0.00 sec)
> 
> +----+
> | id |
> +----+
> |  1 |
> |  2 |
> |  3 |
> |  4 |
> |  5 |
> +----+
> 5 rows in set (0.00 sec)
> 
> session 2> ALTER TABLE child ADD KEY (id);
> Query OK, 0 rows affected (0.01 sec)
> Records: 0  Duplicates: 0  Warnings: 0
> 
> #Succeed even with MDL lock on parent
> ```

Looks like in 5.7, Server isn't aware of FKs. All the locking is done internally by InnoDB.
From 8.0, I believe this WL changed the way MDL locking is done: https://dev.mysql.com/worklog/task/?id=6049

So inserts into child will take MDLs on parent as well. 
"If some statement is going to update or insert into a child
      table in a FK relationship, it should also acquire a SR lock
      on its parent as it is  going to be used for checks."

altmannmarcelo:
@satya-bodapati good point. 8.0 will indeed require to lock FK relationship. `collect_and_lock_fk_tables_for_complex_alter_table` is responsible for that and is called as part of either `mysql_inplace_alter_table` or `mysql_alter_table`. Dependent table lock is only taken when changing parent table of a FK relationship. So for example:

```
session 1> START TRANSACTION; SELECT * FROM child;
Query OK, 0 rows affected (0.00 sec)

+------+-----------+
| id   | parent_id |
+------+-----------+
|    1 |         1 |
|    2 |         2 |
|    3 |         3 |
+------+-----------+
3 rows in set (0.00 sec)

session 2> ALTER TABLE parent ADD KEY (id); -- This will wait
```
Session 2 will wait until session 1 release MDL lock on child. If session 1 now tries to acquire MDL on parent it will deadlock:
```
session 1> SELECT * FROM parent;
ERROR 1213 (40001): Deadlock found when trying to get lock; try restarting transaction

session 2> ALTER TABLE parent ADD KEY (id);
Query OK, 0 rows affected (15 min 10.76 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

So for context of this PR in case of conflicting MDL's, MySQL will handle them via deadlocks.

altmannmarcelo:
@satya-bodapati https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/101/ now tests are passing

satya-bodapati:
> @satya-bodapati good point. 8.0 will indeed require to lock FK relationship. `collect_and_lock_fk_tables_for_complex_alter_table` is responsible for that and is called as part of either `mysql_inplace_alter_table` or `mysql_alter_table`. Dependent table lock is only taken when changing parent table of a FK relationship. So for example:
> 
> ```
> session 1> START TRANSACTION; SELECT * FROM child;
> Query OK, 0 rows affected (0.00 sec)
> 
> +------+-----------+
> | id   | parent_id |
> +------+-----------+
> |    1 |         1 |
> |    2 |         2 |
> |    3 |         3 |
> +------+-----------+
> 3 rows in set (0.00 sec)
> 
> session 2> ALTER TABLE parent ADD KEY (id); -- This will wait
> ```
> 
> Session 2 will wait until session 1 release MDL lock on child. If session 1 now tries to acquire MDL on parent it will deadlock:
> 
> ```
> session 1> SELECT * FROM parent;
> ERROR 1213 (40001): Deadlock found when trying to get lock; try restarting transaction
> 
> session 2> ALTER TABLE parent ADD KEY (id);
> Query OK, 0 rows affected (15 min 10.76 sec)
> Records: 0  Duplicates: 0  Warnings: 0
> ```
> 
> So for context of this PR in case of conflicting MDL's, MySQL will handle them via deadlocks.

Thanks for the analysis

storage/innobase/log/log0recv.cc
```diff
@@ -1793,7 +1793,7 @@ recv_parse_or_apply_log_rec_body(
 		of offline backup and continue.
 		*/
 		if (!recv_recovery_on) {
-			if (!opt_lock_ddl_per_table) {
+			if (mdl_taken) {
 				if (backup_redo_log_flushed_lsn
```
satya-bodapati, line 1797: On more thought:

I think we can get rid of this variable altogether. Just start with if (backup_redo_flushed_lsn)..
(assuming backup_redo_flushed_lsn is noted always).

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
satya-bodapati: Please add the reason you mentioned on slack as code comment. Why a bool is better to have than removing it

satya-bodapati (approved):

storage/innobase/log/log0recv.cc
altmannmarcelo: I would prefer that we keep it for two reasons:

1. If you have a huge list of tables, querying INNODB_SYS_TABLES can take sometime, and there is no problem in accepting MLOG_INDEX_LOAD while we are not done with the MDL step.
2. In case PXB is waiting on MDL like an ALTER table is ongoing, it's also safe to keep accepting MLOG_INDEX_LOAD on tables that we have not yet acquired MDL

PR #939, opened 2020-08-25
8.0 Implemented PXB-2254 (Redesign --lock-ddl-per-table)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2254

The current implementation of --lock-ddl-per-table works as follow:

Creates a dedicated mysql connection for MDL and start a transaction.
Start to scan directories (databases) looking for .ibd files.
For each .ibd file it tries to transalate the space id recorded on
the file header to a table (or tables in case of shared tablespace).
If it finds a table associate with the .ibd (or tables) it executes a
SELECT * FROM table LIMIT 1 using the dedicated connection for MDL.
From this point forward, the table is locked and new DDL will be
blocked until the MDL connection commits.
With this, it has been assumed that if a MLOG_LOAD_INDEX event has
been generated (https://dev.mysql.com/worklog/task/?id=7277) it's safe
to continue since we have taken MDL on the table we are copying.
This design has a few flaws:

FTS have their own tablespace, which is been copied without
MDL protection.
Ongoing DDL that generate temporary table (#sql...) are copied
without MDL protection.
When creating FTS for the first time, MySQL will have to create a
tmp table (Item 2) and FTS files (Item 1) are pointing to temp table.
Table space IDs for FTS or regular tables will change if you
DROP and then CREATE the same FTS/TABLE. MDL lock query will fail to
find the corresponding table but file will exist, thus will be copied
without MDL protection.
If a table is created after xtrabackup has gathered the list of
.ibd's to copy and before the backup has finished, this particular
table will be part of the backup as it will be contained within the
redo log. However, if a bulk index load operations have been performed
( MLOG_INDEX_LOAD has been written to the redo log), this change to the
data file will not be tracked and will be missed from the backup.
This commit changes to:

MDL is taken at the beginning of backup for all know tables at
that time, before we start to copy individual .ibd files.
A new flag controls whatever MDL step has been completed or not.
Before MDL from step 1 is completed, redo follow thread accept
parsing MLOG_INDEX_LOAD. At this point is still safe to accept it since
we have not started to copy .ibd files. After MDL part is completed,
mdl_taken changes to true and parsing a MLOG_INDEX_LOAD causes the
backup to abort.
Before we execute the MDL query for a specific entry from
INNODB_SYS_TABLES, we validate it against a regexp for a FTS file name,
if we are manipulating a FTS index, we translate it to the corresponding
table name.
If for any reason we cannot execute the MDL lock query (for example,
a DDL is ongoing and we are manipulating a file starting with #sql-)
we do not abort the backup. We will only abort if a subsequent
MLOG_INDEX_LOAD is found by the redo follow thread during the backup.
We identify temporary tables (#sql-) and skip trying to lock them.
No reason to attempt to query it as it will always result in an error.
MDL query changed to not bring any row back. This is enough for
taking a MDL while saving the server from bringing any row back.

### DIFF
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -31,6 +31,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "backup_mysql.h"
 #include "common.h"
 #include "os0event.h"
+#include "xb0xb.h"
 #include "xtrabackup.h"
 
 extern ds_ctxt_t *ds_redo;
@@ -944,12 +945,23 @@ bool Redo_Log_Data_Manager::start() {
   last_checkpoint_lsn = start_checkpoint_lsn;
   stop_lsn = 0;
 
+  if (opt_lock_ddl_per_table) {
+    mdl_lock_tables();
+  }
+
   bool finished = false;
   while (!finished) {
     if (!copy_once(false, &finished)) {
       return (false);
     }
   }
+
+  /*
+   * From this point forward, recv_parse_or_apply_log_rec_body should fail if
+   * MLOG_INDEX_LOAD event is parsed as its not safe to continue the backup
+   * in any situation (with or without --lock-ddl-per-table).
+   */
+  mdl_taken = true;
 
   thread.start();
 
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -25,6 +25,7 @@ extern bool innodb_log_checksums_specified;
 extern bool innodb_checksum_algorithm_specified;
 
 extern bool opt_lock_ddl_per_table;
+extern bool mdl_taken;
 
 extern bool use_dumped_tablespace_keys;
 
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -35,6 +35,7 @@ extern bool innodb_log_checksum_algorithm_specified;
 extern bool innodb_checksum_algorithm_specified;
 
 extern my_bool opt_lock_ddl_per_table;
+extern bool mdl_taken;
 
 extern bool use_dumped_tablespace_keys;
 
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -1655,7 +1655,7 @@ static byte *recv_parse_or_apply_log_rec_body(
       Error out in case of online backup and emit a warning in case
       of offline backup and continue. */
       if (!recv_recovery_on) {
-        if (!opt_lock_ddl_per_table) {
+        if (mdl_taken) {
           if (backup_redo_log_flushed_lsn < recv_sys->recovered_lsn) {
             ib::info() << "Last flushed lsn: " << backup_redo_log_flushed_lsn
                        << " load_index lsn " << recv_sys->recovered_lsn;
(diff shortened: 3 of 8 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/317/

altmannmarcelo:
@satya-bodapati I spoke with @georgelorchpercona yesterday, we are good to go in terms of calling `--lock-ddl-per-table` deprecated on PXB 8.0

satya-bodapati:
> @satya-bodapati I spoke with @georgelorchpercona yesterday, we are good to go in terms of calling `--lock-ddl-per-table` deprecated on PXB 8.0

ok @altmannmarcelo  did you add the deprecation message ?

altmannmarcelo:
> > @satya-bodapati I spoke with @georgelorchpercona yesterday, we are good to go in terms of calling `--lock-ddl-per-table` deprecated on PXB 8.0
> 
> ok @altmannmarcelo did you add the deprecation message ?

@satya-bodapati yes - https://github.com/percona/percona-xtrabackup/pull/939/files#diff-29fd8beae6c2cfce456b51a213d966eeR6857

satya-bodapati:
> > > @satya-bodapati I spoke with @georgelorchpercona yesterday, we are good to go in terms of calling `--lock-ddl-per-table` deprecated on PXB 8.0
> > 
> > 
> > ok @altmannmarcelo did you add the deprecation message ?
> 
> @satya-bodapati yes - https://github.com/percona/percona-xtrabackup/pull/939/files#diff-29fd8beae6c2cfce456b51a213d966eeR6857

ok @altmannmarcelo, will check tomorrow

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -944,13 +945,24 @@ bool Redo_Log_Data_Manager::start() {
...
       return (false);
     }
   }
 
+  /*
+   * From this point forward, recv_parse_or_apply_log_rec_body should fail if
+   * MLOG_INDEX_LOAD event is parsed as its not safe to continue the backup
+   * in any situation (with or without --lock-ddl-per-table).
+   */
+  mdl_taken = true;
```
satya-bodapati, line 964: One doubt, if opt_lock_ddl_per_table is OFF (mdl is not taken), we still set mdl_taken= true. Is ithis intentional?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: @satya-bodapati yes, because regardless of MDL been taken or not, later at `log0recv.cc` https://github.com/percona/percona-xtrabackup/pull/939/files#diff-ec5cdb94d89780d34753579b799e83ebR1658 we cannot accept continuing the backup if we parse `MLOG_INDEX_LOAD`. Before we were only entering that condition if `lock-ddl-per-table` was off, assuming that if it was ON it was safe to continue the backup even though we parsed `MLOG_INDEX_LOAD`.

satya-bodapati: Ok, with LOCK INSTANCE/LOCK_TABLES_FOR_BACKUP taken earlier, we wouldn't see MLOG_INDEX_LOAD. So the setting mdl_taken=true is OK for these scenarios.

satya-bodapati (approved):

PR #955, opened 2020-09-21
PXB-2112 xbcloud: Add support for S3 storage classes

BenoitKnecht:
Add `--s3-storage-class` option to specify the S3 storage class object should
be uploaded in, instead of the `STANDARD` default.

Closes https://jira.percona.com/browse/PXB-2112

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>


BenoitKnecht:
#974 will be ported to 8.0.

PR #956, opened 2020-09-22
PXB-2242 Prevent back up if Source system (DB) version is higher the…

rahulmalik87:
…n current PXB version

Add a configuration parameter --no-server-version-check to allow override
of the check process.  The default is override is off (must turn on override) 
If same base upstream/Percona version and PXB version - allow back up to proceed
If base upsteam/Percona version is less then PXB version - allow back up
to proceed
If base upstream/Percona version is greater (newer) then PXB version and
no-server-version-check is  set to allow backup - allow back up to proceed 
If base upstream/Percona version is greater (newer) then PXB version and
no-server-version-check is NOT set to allow back up - prevent back from
occurring - provide compete notification

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -70,6 +70,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "xb_regex.h"
 
 extern uint opt_ssl_mode;
+extern bool opt_no_server_version_check;
 extern char *opt_ssl_ca;
 extern char *opt_ssl_capath;
 extern char *opt_ssl_cert;
@@ -363,6 +364,25 @@ void parse_show_engine_innodb_status(MYSQL *connection) {
   mysql_free_result(mysql_result);
 }
 
+/* find the pxb base version */
+static unsigned long pxb_base_version() {
+  std::string pxb_base = MYSQL_SERVER_VERSION;
+  unsigned long major = 0, minor = 0, version = 0;
+  std::size_t major_p = pxb_base.find(".");
+  if (major_p != std::string::npos) major = stoi(pxb_base.substr(0, major_p));
+
+  std::size_t minor_p = pxb_base.find(".", major_p + 1);
+  if (minor_p != std::string::npos)
+    minor = stoi(pxb_base.substr(major_p + 1, minor_p - major_p));
+
+  std::size_t version_p = pxb_base.find(".", minor_p + 1);
+  if (version_p != std::string::npos)
+    version = stoi(pxb_base.substr(minor_p + 1, version_p - minor_p));
+  else
+    version = stoi(pxb_base.substr(minor_p + 1));
+  return major * 10000 + minor * 100 + version;
+}
+
 static bool check_server_version(unsigned long version_number,
                                  const char *version_string,
                                  const char *version_comment,
@@ -392,6 +412,7 @@ static bool check_server_version(unsigned long version_number,
   pxb24 = pxb24 || ((version_number > 100000 && version_number < 100300) &&
                     server_flavor == FLAVOR_MARIADB);
 
+
   if (!version_supported) {
     msg("Error: Unsupported server version: '%s'.\n"
         "This version of Percona XtraBackup can only perform backups and "
@@ -405,6 +426,18 @@ static bool check_server_version(unsigned long version_number,
     if (pxb24) {
       msg("Please use Percona XtraBackup 2.4 for this database.\n");
     }
+  }
+
+
+  auto pxb_base_ver = pxb_base_version();
+
+  DBUG_EXECUTE_IF("simulate_lower_version", pxb_base_ver = 80014;);
+
+  if (!opt_no_server_version_check && pxb_base_ver < version_number) {
+    msg("Error: Unspported server version %lu, Use latest version of PXB "
+        "or to continue with risk, use --no-server-version-check",
+        version_number);
+    version_supported = false;
   }
 
   return (version_supported);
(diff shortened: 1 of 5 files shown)


rahulmalik87:
jenkin results https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/lastCompletedBuild/

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -407,6 +407,23 @@ static bool check_server_version(unsigned long version_number,
     }
   }
 
+  std::string mysql_version = MYSQL_SERVER_VERSION;
```
altmannmarcelo: I would change the variable name to be more clear of what it is about. `mysql_version` seems like this is the server version we are trying to backup, while in fact this is pxb base mysql version.

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -407,6 +407,23 @@ static bool check_server_version(unsigned long version_number,
...
+
+  // replace 8.0.21 first . with 0 and second . with null
+  std::size_t found = mysql_version.find_first_of(".");
+  if (found != std::string::npos) mysql_version[found] = '0';
```
altmannmarcelo: This assumes PXB will always have x.x.xx version pattern, which will require to re-work this function in future. For example, I did a test with PS 8.10.22 and PXB 8.10.21:
```
(gdb) p mysql_version_int
+p mysql_version_int
$3 = 801021
(gdb) p version_number
+p version_number
$4 = 81022
```
I would make this function to use the same logic as mysql_get_server_version, which is where `version_number` is extracted from.

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -407,6 +407,23 @@ static bool check_server_version(unsigned long version_number,
...
+  found = mysql_version.find_first_of(".");
+  if (found != std::string::npos)
+    mysql_version.erase(mysql_version.begin() + found);
+
+  unsigned long mysql_version_int = stoi(mysql_version);
+  if (!opt_no_server_version_check && mysql_version_int < version_number) {
+    msg("Error: Unspported server version %s, Use latest version of PXB "
+        "or to continue with risk, use --no-server-version-check",
+        MYSQL_SERVER_VERSION);
```
altmannmarcelo: > Error: Unspported server version %s

This is reporting PXB version (`MYSQL_SERVER_VERSION`), while it should be reporting MySQL server version - `version_number`.

altmannmarcelo (changes requested):

altmannmarcelo:
LGTM

altmannmarcelo (approved):

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -363,6 +364,25 @@ void parse_show_engine_innodb_status(MYSQL *connection) {
...
+    minor = stoi(pxb_base.substr(major_p + 1, minor_p - major_p));
+
+  std::size_t version_p = pxb_base.find(".", minor_p + 1);
+  if (version_p != std::string::npos)
+    version = stoi(pxb_base.substr(minor_p + 1, version_p - minor_p));
+  else
+    version = stoi(pxb_base.substr(minor_p + 1));
+  return major * 10000 + minor * 100 + version;
+}
+
```
satya-bodapati, line 385: Does it work for debug binaries? And if the server is debug version?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1020,6 +1022,13 @@ struct my_option xb_client_options[] = {
      (uchar *)&opt_force_non_empty_dirs, (uchar *)&opt_force_non_empty_dirs, 0,
      GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
 
+    {"no-server-version-check", OPT_NO_SERVER_VERSION_CHECK,
+     "This option allow backup to proceed"
```
```suggestion
     "This option allows backup to proceed"
```

storage/innobase/xtrabackup/test/t/version_check.sh
```diff
@@ -10,6 +10,10 @@
 
 start_server
 
+#test server version check
+xtrabackup --backup --no-server-version-check --target-dir=$topdir/backup1
+rm -r $topdir/backup1
+
```
satya-bodapati, line 16: Can you add a test case for a negative scenario?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: yes, it works for debug version,  release version and patch version. 
i.e it works for all version like 8.0.12 or  8.0.12.debug or 8.0.12.1

storage/innobase/xtrabackup/test/t/version_check.sh
satya-bodapati: If you find it hard to introduce have_debug.sh/inc, please let me know.

satya-bodapati (changes requested):

satya-bodapati:
Also click on re-request review, when it is ready for review again

rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/lastBuild/

mdl_lock.sh  is due to --lock-ddl-per-table
xb_parallel_encrypt.sh  is successful on local. 
will give another round to see if it is succesful.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Why do I still see "allow" instead of "allows?" I see that you marked it as resolved..

satya-bodapati (changes requested):

altmannmarcelo (approved):
LGTM

satya-bodapati (approved):
Please squash before merging

PR #957, opened 2020-09-22
Fixed PXB-2239 (Partitioned table is not restored correctly when part…

altmannmarcelo:
…itions are changed during backup)

https://jira.percona.com/browse/PXB-2239

MySQL 5.x has not atomic DDL support. In order to mitigate issues with
DDL's that touch multiple tables it record on-going operations into
ddl_log.log file. In case a crash happens in the middle of a DDL, MySQL
will read this file as part of execute_ddl_log_recovery and rollback
half done changes such as temporary partition files.
PXB was not copying this file, which caused such rollback operations to
not be performed as part of MySQL startup after a backup has been
restored.
As part of this fix, we ensure that if we are doing an incremental backup
we always leave the ddl_log.log from last incremental backup if it exists,
or otherwise, remove the file completely.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1603,6 +1603,9 @@ backup_finish()
 		if (file_exists("ib_lru_dump")) {
 			copy_file(ds_data, "ib_lru_dump", "ib_lru_dump", 0);
 		}
+		if (file_exists("ddl_log.log")) {
+			copy_file(ds_data, "ddl_log.log", "ddl_log.log", 0);
+		}
 	}
 
 	msg_ts("Backup created in directory '%s'\n", xtrabackup_target_dir);
@@ -1639,6 +1642,7 @@ ibx_copy_incremental_over_full()
 				   "xtrabackup_info",
 				   "xtrabackup_keys",
 				   "ib_lru_dump",
+				   "ddl_log.log",
 				   NULL};
 	datadir_iter_t *it = NULL;
 	datadir_node_t node;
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/102/

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1603,6 +1603,9 @@ backup_finish()
 		if (file_exists("ib_lru_dump")) {
 			copy_file(ds_data, "ib_lru_dump", "ib_lru_dump", 0);
 		}
+		if (file_exists("ddl_log.log")) {
+			copy_file(ds_data, "ddl_log.log", "ddl_log.log", 0);
+		}
```
satya-bodapati, line 1608: One doubt. So this file is parsed/used by the server and not by PXB full/incremental backups right?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: correct, server uses this file at `execute_ddl_log_recovery()` to decide if it has to rollback any multi table/patition DDL.

satya-bodapati: Ok and testcase is hard to simulate?

altmannmarcelo: @satya-bodapati - there you go https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/104/

satya-bodapati: @altmannmarcelo Good work on the testcase!

satya-bodapati (approved):

PR #960, opened 2020-09-29
PXB-2279 Xbcloud: Upload failed: backup is incomplete

rahulmalik87:
PXB-2279 Xbcloud: Upload failed: backup is incomplete

    Problem:
    When backup is a compressed backup, xbcloud "put" operation fails with an error message
    The "put" operation is successful but wrongly throws that it failed

    Analysis:
    To validate that xbcloud put operation is successful, xbcloud looks for the
    presence of ""xtrabackup_info.00000000000000000000" in the list of files
    in cloud storage. In case of compressed backup, the filename is
    "xtrabackup_info.gz.00000000000"

    Because of the extra ".gz" in the filename, it doesn't match with the expected
    name and xbcloud wrongly prints the error message.

    Fix:
    Change the check to real last file "xtrabackup_tablespaces*" instead of "xtrabackupinfo*"
    Look for the prefix "xtrabackup_tablespaces" without the extensions

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -707,9 +707,20 @@ bool xbcloud_put(Object_store *store, const std::string &container,
     return false;
   }
 
-  // check if xtrabackup_info, almost the last file exists in backup
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  /* check if the last_file (xtrabackup_tablespaces.gz.00000000000 or
+  xtrabackup_tablespaces.00000000000000000000) is uploaded to cloud storage to
+  determine successful xbcloud "put" operation. */
+  std::string last_file_prefix = backup_name + "/xtrabackup_tablespaces";
+  bool file_found = false;
+
+  if (!object_list.empty()) {
+    auto cur_file = object_list.back();
+    auto last_file_size = last_file_prefix.size();
+    if (cur_file.size() >= last_file_size &&
+        cur_file.substr(0, last_file_size).compare(last_file_prefix) == 0)
+      file_found = true;
+  }
+  if (!file_found) {
     msg_ts("%s: Upload failed: backup is incomplete.\n", my_progname);
     return false;
   }
--- a/storage/innobase/xtrabackup/test/t/xbcloud.sh
+++ b/storage/innobase/xtrabackup/test/t/xbcloud.sh
@@ -139,6 +139,33 @@ if ! grep -q "Upload failed" $topdir/pxb-2202.log ; then
     die 'xbcloud did not exit with error on upload'
 fi
 
+#PXB-2279 Xbcloud: Upload failed: backup is incomplete
+vlog "take full backup to test compression"
+run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
+	${full_backup_name} --parallel=4
+rm -r $full_backup_dir
+xtrabackup --backup --stream=xbstream --compress --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
+	    --parallel=4 \
+	    ${full_backup_name} 2> $topdir/pxb-2279.log
+if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
+    die 'xbcloud exit with error on upload'
+fi
+vlog "take full backup to test encryption"
+run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
+	${full_backup_name} --parallel=4
+rm -r $full_backup_dir
+xtrabackup --backup --stream=xbstream  --encrypt=AES256 \
+          --encrypt-key="percona_xtrabackup_is_awesome___" --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
+	    --parallel=4 \
+	    ${full_backup_name} 2> $topdir/pxb-2279.log
+if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
+    die 'xbcloud exit with error on upload'
+fi
+
 # cleanup
 run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
 	${full_backup_name} --parallel=4


rahulmalik87:
Jenkins results https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param/lastCompletedBuild/

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,8 +708,17 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+    if (cur_file.size() > last_file_expecting.size() &&
+        cur_file.substr(0, last_file_expecting.size())
+                .compare(last_file_expecting) == 0) {
+      file_found = true;
+      break;
+    }
+  }
+  if (!file_found) {
     msg_ts("%s: Upload failed: backup is incomplete.\n", my_progname);
     return false;
```
satya-bodapati: Not clear what you are trying to do? before this change you were looking for the file "xtrabackup_info.00000000000000000000" from the cloud backup dir. What is the new strategy for compressed files? Do they end with .xbcrypt extension? Can you please explain with example?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: they end with xtrabackup_info.gz.00000000000 so now we are just checking if there exists file initial substring is xtrabackup_info. So now xtrabackup_info.gz.00000000000 or xtrabackup_info.00000000000000000000 both will pass the check

satya-bodapati: thanks for explanation

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,8 +708,17 @@ bool xbcloud_put(Object_store *store, const std::string &container,
   }
 
   // check if xtrabackup_info, almost the last file exists in backup
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  std::string last_file_expecting = backup_name + "/xtrabackup_info";
```
satya-bodapati:
```suggestion
  std::string last_file_prefix = backup_name + "/xtrabackup_info";
```

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,8 +708,17 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+  bool file_found = false;
+  for (auto cur_file : object_list) {
```
satya-bodapati: if the objec_list maintains insert order, can we check only the last file?
Is there a reverse iterator? ie. from last file in the object_list to the first one?

satya-bodapati: Please add a comment here that we have two possibilities for the last filename  xtrabackup_info.gz.00000000000 or xtrabackup_info.00000000000000000000 and we try to match the prefix "backup/xtrabackup_info" to determine if xbcloud is successful or not

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: it is not exactly the last file. the last could be xtrabackup_tablespaces. also, the exact order of xtrabackup_info is not certain. Either we check for xtrabackup_tablespaces. 
I am not sure if we have some option with which we can generate some additional file in end.

satya-bodapati: ok, two questions
1.  Why don't we check for the "real" last file?
2. What about the reverse iterator?

rahulmalik87: fixed, using the real last file. 
tested with --generate-transition-key and --encryption. works fine.

satya-bodapati: great and for question 2?

rahulmalik87: already using reverse iterator

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -707,9 +707,22 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+  /* check if xtrabackup_tablespaces, the last file exists in backup we have two
+    possibilities for the last filename xtrabackup_tablespaces.gz.00000000000 or
+    xtrabackup_tablespaces.00000000000000000000 and we try to match the prefix
+    "backup/xtrabackup_tablespaces" to determine if xbcloud is successful or not
+  */
+  std::string last_file_prefix = backup_name + "/xtrabackup_tablespaces";
+  bool file_found = false;
+
+  if (object_list.size() > 2) {
+    auto cur_file = *(object_list.end() - 1);
```
satya-bodapati: use !vector.empty() and vector.back() instead

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -707,9 +707,22 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
   }
 
-  // check if xtrabackup_info, almost the last file exists in backup
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  /* check if xtrabackup_tablespaces, the last file exists in backup we have two
+    possibilities for the last filename xtrabackup_tablespaces.gz.00000000000 or
+    xtrabackup_tablespaces.00000000000000000000 and we try to match the prefix
+    "backup/xtrabackup_tablespaces" to determine if xbcloud is successful or not
+  */
```
satya-bodapati: Suggestion:

check if the last_file (xtrabackup_tablespaces.gz.00000000000 or  xtrabackup_tablespaces.00000000000000000000) is uploaded to cloud storage to
determine successful xbcloud "put" operation.

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -139,6 +139,22 @@ if ! grep -q "Upload failed" $topdir/pxb-2202.log ; then
...
 fi
 
+#PXB-2279 Xbcloud: Upload failed: backup is incomplete
+vlog "take full backup to test compression"
+run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
+	${full_backup_name} --parallel=4
+rm -r $full_backup_dir
+xtrabackup --backup --stream=xbstream --compress --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
```
satya-bodapati, line 149: can you please add testcase for encrypted backups too?

satya-bodapati (changes requested):
Looks good. Minor comments

Suggestion on Commit message:

Problem:
When backup is a compressed backup, xbcloud "put" operation fails with an error message
The "put" operation is succesful but wrongly throws that it failed

Analysis:
To validate that xbcloud put operation is successful, xbcloud looks for the
presence of ""xtrabackup_info.00000000000000000000" in the list of files
in cloud storage. In case of compressed backup, the filename is
"xtrabackup_info.gz.00000000000"

Because of the extra ".gz" in the filename, it doesn't match with the expected
name and xbcloud wrongly prints the error message.

Fix
----
1. Change the check to real last file "xtrabackup_tablespaces*" instead of "xtrabackupinfo*"
2. Look for the prefix "xtrabackup_tablespaces" without the extensions

rahulmalik87:
$./run.sh -t t/xbcloud.sh
Using /home/rahul.malik/MySQL/src/x8/bld/storage/innobase/xtrabackup/test/var as test root
Detecting server version...
Running against MySQL 8.0.21-debug (InnoDB 8.0.21)
Using 8 parallel workers

xbcloud                                  w1	[passed]    106
Spent 106 of 124 seconds executing testcases

satya-bodapati (approved):

PR #962, opened 2020-10-01
null-merge Fixed PXB-2239 (Partitioned table is not restored correctly when part…

altmannmarcelo:
…itions are changed during backup)

https://jira.percona.com/browse/PXB-2239

MySQL 5.x has not atomic DDL support. In order to mitigate issues with
DDL's that touch multiple tables it record on-going operations into
ddl_log.log file. In case a crash happens in the middle of a DDL, MySQL
will read this file as part of execute_ddl_log_recovery and rollback
half done changes such as temporary partition files.
PXB was not copying this file, which caused such rollback operations to
not be performed as part of MySQL startup after a backup has been
restored.
As part of this fix, we ensure that if we are doing an incremental backup
we always leave the ddl_log.log from last incremental backup if it exists,
or otherwise, remove the file completely.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1603,6 +1603,9 @@ backup_finish()
 		if (file_exists("ib_lru_dump")) {
 			copy_file(ds_data, "ib_lru_dump", "ib_lru_dump", 0);
 		}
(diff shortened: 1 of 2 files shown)


satya-bodapati (approved):
Best PR ever :stuck_out_tongue_winking_eye:

PR #965, opened 2020-10-01
Fixed PXB-2216 (Backup changing table encryption version)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2216

When a table is created with encryption flag the encryption information
is written in two phases:
1) Encryption flag - this is written synchronously to the .ibd file
2) Encryption information (magic key/version, key, iv, ...) - this is
written on the buffer pool page, generates a redo log and later gets
flushed to .ibd.

This means we can have tables that have encryption flag set but doesn't
have the encryption information yet if we start the backup right after
the table has been created and its page 0 is still dirty.

At the prepare phase we will first open all tablespaces and read
encryption information via:
recv_recovery_from_checkpoint_start->xb_load_single_table_tablespaces->
xb_load_single_table_tablespace->Datafile::validate_first_page->
fsp_header_get_encryption_key->fsp_header_decode_encryption_info.

If the encryption data has not been written to the file yet we will
assume its ENCRYPTION_VERSION_3 as it won't match V1 nor V2 and will
fall the else clause of fsp_header_decode_encryption_info.
This will also change Encryption::max_version from default V2 to V3.
Correct encryption information will be written to page0 in buffer pool
when PXB process the redo log containing that information and table will
have correct key to decrypt the data. However, later when we start the
server at second phase, InnoDB calls innobase_start_or_create_for_mysql,
which will rotate tablespace keys via fil_encryption_rotate().
This will update the tablespace key from each table to the latest entry
on keyring. At this point Keyring will return the latest key as
ENCRYPTION_VERSION_3.

Fix ensures we properly catch each valid encryption version when opening
tables, skip reading encryption information if magic key is not set and
fail in case a not empty or not know version is read from tablespace.

This commit also fixes PXB-2287 (split innodb_encryption test).


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/109/

rahulmalik87 (approved):
LGTM

satya-bodapati (commented):
Looks good.
@altmannmarcelo 
Can you please check if the below tests fail on 2.4 trunk also?
t/bug1135431.sh
t/bug1248331.sh
t/bug1623210.sh

(It looks like PXB 2.4 trunk has lot of failures :cry: )

satya-bodapati (approved):
Please check if these test failures can be fixed as part of 2.4 refresh

PR #970, opened 2020-10-08
2.4 Fixed PXB-1462 (long gtid_executed breaks --history functionality)

altmannmarcelo:
https://jira.percona.com/browse/PXB-1462

xtrabackup_history table has a limited length for binlog_pos.
gtid_executed field can grown in case of gaps of multi source
replication making it not fit the current 128 char long field.

Fix
Change binlog_pos to use TEXT fild type.


altmannmarcelo:
new PR #1069

PR #974, opened 2020-10-09
2.4 PXB-2112 xbcloud: Add support for S3 storage classes

rahulmalik87:
Introduce new storage class support for xbcloud.

Instead of the using "STANDARD" option for S3 storage, we will now support multiple
storage class types.

Introduced --s3-storage-class option which accepts STANDARD|STANDARD_IA|GLACIER|
Introduced --google-storage-class option which accepts STANDARD|NEARLINE|COLDLINE|ARCHIVE

Thanks to Benoît Knecht <bknecht@protonmail.ch> for providing the patch

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -84,6 +84,7 @@ static char *opt_s3_endpoint = nullptr;
 static char *opt_s3_access_key = nullptr;
 static char *opt_s3_secret_key = nullptr;
 static char *opt_s3_session_token = nullptr;
+static char *opt_s3_storage_class = nullptr;
 static char *opt_s3_bucket = nullptr;
 static ulong opt_s3_bucket_lookup;
 static ulong opt_s3_api_version = 0;
@@ -93,6 +94,7 @@ static char *opt_google_endpoint = nullptr;
 static char *opt_google_access_key = nullptr;
 static char *opt_google_secret_key = nullptr;
 static char *opt_google_session_token = nullptr;
+static char *opt_google_storage_class = nullptr;
 static char *opt_google_bucket = nullptr;
 
 static std::string backup_name;
@@ -143,6 +145,7 @@ enum {
   OPT_S3_ACCESS_KEY,
   OPT_S3_SECRET_KEY,
   OPT_S3_SESSION_TOKEN,
+  OPT_S3_STORAGE_CLASS,
   OPT_S3_BUCKET,
   OPT_S3_BUCKET_LOOKUP,
   OPT_S3_API_VERSION,
@@ -152,6 +155,7 @@ enum {
   OPT_GOOGLE_ACCESS_KEY,
   OPT_GOOGLE_SECRET_KEY,
   OPT_GOOGLE_SESSION_TOKEN,
+  OPT_GOOGLE_STORAGE_CLASS,
   OPT_GOOGLE_BUCKET,
 
   OPT_PARALLEL,
@@ -267,6 +271,14 @@ static struct my_option my_long_options[] = {
      &opt_s3_session_token, &opt_s3_session_token, 0, GET_STR_ALLOC,
      REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 
+    {"s3-storage-class", OPT_S3_STORAGE_CLASS,
+     "S3 storage class. STANDARD|STANDARD_IA|GLACIER|...     "
+     "... is meant for passing "
+     "custom storage class names provided by other S3 implementations such as "
+     "MinIO CephRadosGW, etc.",
+     &opt_s3_storage_class, &opt_s3_storage_class, 0, GET_STR_ALLOC,
+     REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
+
     {"s3-bucket", OPT_S3_BUCKET, "S3 bucket.", &opt_s3_bucket, &opt_s3_bucket,
      0, GET_STR_ALLOC, REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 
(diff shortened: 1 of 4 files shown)


rahulmalik87:
need to check the behaviour with google cloud.  Will update once I have the access.

satya-bodapati (changes requested):
Are there tests for new code?

altmannmarcelo:
s3 seem ok.

altmannmarcelo (changes requested):
waiting for GC

rahulmalik87:
> need to check the behaviour with google cloud. Will update once I have the access.

google cloud is working fine.

rahulmalik87:
> Are there tests for new code?

yes. I am have tested manually. google cloud and s3 both are working fine.

satya-bodapati:
> > Are there tests for new code?
> 
> yes. I am have tested manually. google cloud and s3 both are working fine.

Can you paste the results? Is it possible to add testcase (and we will get Jenkins fixed to run these tests (along with xbcloud.test))

satya-bodapati (changes requested):
see the last comment

rahulmalik87:
> > > Are there tests for new code?
> > 
> > 
> > yes. I am have tested manually. google cloud and s3 both are working fine.
> 
> Can you paste the results? Is it possible to add testcase (and we will get Jenkins fixed to run these tests (along with xbcloud.test))

./run.sh -t t/xbcloud.sh
Detecting server version...
Running against MySQL 5.7.31-log (InnoDB 5.7.31)
Using 8 parallel workers

xbcloud                                  w1	[passed]    134
Spent 134 of 140 seconds executing testcases

I don't have the google unit test. but xbcloud is uploading fine to GC. pasting the result of the manual test( result showing storage class NEARLINE in GC)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -1135,7 +1135,7 @@ int main(int argc, char **argv) {
         storage_class,
         opt_google_endpoint != nullptr ? opt_google_endpoint
                                        : "https://storage.googleapis.com/",
-        LOOKUP_DNS, S3_V2));
+        LOOKUP_DNS, S3_V4));
```
satya-bodapati, line 1154: what is the implication of this? will older versions not work?

satya-bodapati (changes requested):
Can you also add the description of the new options introduced and also how to use these options?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: it is signing process. I am not sure why we use V2 with GC and not V4.  GC supports V4 and it recommends users to use that.

satya-bodapati: my concern is about newer xbcloud versions not being able to restore older backups stored in GC. Will there be any issues?
Can you upload a backup without your fix and restore using newer xbcloud (with fix that uses V4)?

rahulmalik87: yes. upload with V2 without my patch. and download with V4 with my patch is working fine.

satya-bodapati: Ok

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -267,6 +271,11 @@ static struct my_option my_long_options[] = {
      &opt_s3_session_token, &opt_s3_session_token, 0, GET_STR_ALLOC,
      REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 
+    {"s3-storage-class", OPT_S3_STORAGE_CLASS,
+     "S3 storage class. STANDARD|STANDARD_IA|GLACIER|...",
```
satya-bodapati: Are these three dots necessary? ".." ? Do you mean that there are more options supported?

satya-bodapati (changes requested):
I've rewritten the description, can you please make that as part of commit message?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
BenoitKnecht: Maybe not on Amazon S3 (I'm not sure), but there are other S3 implementations (such as [Ceph RadosGW](https://docs.ceph.com/en/latest/radosgw/) or [MinIO](https://min.io/)) that can define arbitrary storage classes. Since the option itself will accept any storage class name and pass it on to the API (as it should), perhaps it makes sense to use `...` to suggest there may be other possibilities.

satya-bodapati: Thanks for the clarification!

@rahulmalik87 can you add some documentation about the "..". Please say that it meant for passing custom storage class names provided other S3 implementations such as MinIO CephRadosGW, etc.

rahulmalik87: --s3-storage-class=name
                      S3 storage class. STANDARD|STANDARD_IA|GLACIER|...
                      ... is meant for passing custom storage class names
                      provided by other S3 implementations such as MinIO
                      CephRadosGW, etc.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -1,5 +1,5 @@
 /******************************************************
-Copyright (c) 2014-2020 Percona LLC and/or its affiliates.
+ Name of config fileCopyright (c) 2014-2020 Percona LLC and/or its affiliates.
```
satya-bodapati: Intentional change??

satya-bodapati: Doesn't seem to be. Please revert

satya-bodapati (approved):
Please revert that copyright change and then it is OK to be merged.

BenoitKnecht:
So what about #955, the original PR targeting 8.0?

Will this PR be ported to 8.0, or should #955 be edited to add some of the changes made here?

altmannmarcelo:
All changes affecting both versions are done on the lower version first then ported to the upper version. This commit will be merged to the 8.0 branch soon.

PR #977, opened 2020-10-15
2.4 PXB-2279 Xbcloud: Upload failed: backup is incomplete (#960)

rahulmalik87:
PXB-2279 Xbcloud: Upload failed: backup is incomplete (#960)

    Problem:
    When backup is a compressed backup, xbcloud "put" operation fails with an error message
    The "put" operation is successful but wrongly throws that it failed

    Analysis:
    To validate that xbcloud put operation is successful, xbcloud looks for the
    presence of ""xtrabackup_info.00000000000000000000" in the list of files
    in cloud storage. In case of compressed backup, the filename is
    "xtrabackup_info.gz.00000000000"

    Because of the extra ".gz" in the filename, it doesn't match with the expected
    name and xbcloud wrongly prints the error message.

    Fix:
    Change the check to file  "xtrabackup_info*" instead of "xtrabackup_info*"
    Look for the prefix "xtrabackup_info" without the extensions

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -715,7 +715,7 @@ bool xbcloud_put(Object_store *store, const std::string &container,
     for (auto cur_file : object_list) {
       if (cur_file.size() >= last_file_size &&
 	  cur_file.substr(0, last_file_size).compare(last_file_prefix) == 0) {
-	file_found = true;
+	  file_found = true;
         break;
       }
     }
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -705,9 +705,22 @@ bool xbcloud_put(Object_store *store, const std::string &container,
     return false;
   }
 
-  // check if xtrabackup_info, almost the last file exists in backup
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  /* check if the xtrabackup_info.gz.00000000000 or
+  xtrabackup_info.00000000000000000000 is uploaded to cloud storage to
+  determine successful xbcloud "put" operation. */
+  bool file_found = false;
+  if (!object_list.empty()) {
+    std::string last_file_prefix = backup_name + "/xtrabackup_info";
+    auto last_file_size = last_file_prefix.size();
+    for (auto cur_file : object_list) {
+      if (cur_file.size() >= last_file_size &&
+	  cur_file.substr(0, last_file_size).compare(last_file_prefix) == 0) {
+	file_found = true;
+        break;
+      }
+    }
+  }
+  if (!file_found) {
     msg_ts("%s: Upload failed: backup is incomplete.\n", my_progname);
     return false;
   }
--- a/storage/innobase/xtrabackup/test/t/xbcloud.sh
+++ b/storage/innobase/xtrabackup/test/t/xbcloud.sh
@@ -138,6 +138,33 @@ if ! grep -q "Upload failed" $topdir/pxb-2202.log ; then
     die 'xbcloud did not exit with error on upload'
 fi
 
+#PXB-2279 Xbcloud: Upload failed: backup is incomplete
+vlog "take full backup to test compression"
+run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
+	${full_backup_name} --parallel=4
+rm -r $full_backup_dir
+xtrabackup --backup --stream=xbstream --compress --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
+	    --parallel=4 \
+	    ${full_backup_name} 2> $topdir/pxb-2279.log
+if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
+    die 'xbcloud exit with error on upload'
+fi
+vlog "take full backup to test encryption"
+run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
+	${full_backup_name} --parallel=4
+rm -r $full_backup_dir
+xtrabackup --backup --stream=xbstream  --encrypt=AES256 \
+          --encrypt-key="percona_xtrabackup_is_awesome___" --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
+	    --parallel=4 \
+	    ${full_backup_name} 2> $topdir/pxb-2279.log
+if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
+    die 'xbcloud exit with error on upload'
+fi
+
 # cleanup
 run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf delete \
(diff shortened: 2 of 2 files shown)


satya-bodapati (changes requested):
Please paste test results (the manual ones)
Please coordinate with @altmannmarcelo as there is a merge on-going.

rahulmalik87:
./run.sh -t t/xbcloud.sh
Detecting server version...
Running against MySQL 5.7.31-log (InnoDB 5.7.31)
Using 8 parallel workers

xbcloud                                  w1	[passed]    72
Spent 72 of 79 seconds executing testcases

{smm-12-d40aec x7 o7 xtrabackup-test}$

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -705,9 +705,20 @@ bool xbcloud_put(Object_store *store, const std::string &container,
     return false;
   }
 
-  // check if xtrabackup_info, almost the last file exists in backup
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  /* check if the last_file (xtrabackup_tablespaces.gz.00000000000 or
+  xtrabackup_tablespaces.00000000000000000000) is uploaded to cloud storage to
+  determine successful xbcloud "put" operation. */
+  std::string last_file_prefix = backup_name + "/xtrabackup_tablespaces";
```
altmannmarcelo: I believe there is confusion here. xtrabackup_tablespaces is not a valid file on 2.4. It's only available on 8.0 branch at storage/innobase/xtrabackup/src/space_map.cc .

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -705,9 +705,20 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+  bool file_found = false;
+
+  if (!object_list.empty()) {
+    auto cur_file = object_list.back();
+    auto last_file_size = last_file_prefix.size();
+    if (cur_file.size() >= last_file_size &&
+        cur_file.substr(0, last_file_size).compare(last_file_prefix) == 0)
+      file_found = true;
```
altmannmarcelo: I think we should test the upload on a per-file basis. I don't think what it's been done here can ensure you have a good backup. Here we only ensure that xtrabackup_tablespaces/xtrabackup_info has been uploaded, but what about all the other files? 
Perhaps we should at least compare the number of files on remote dst against the number of files we expect based on a runtime calculation from copy threads.

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: thank you @altmannmarcelo for pointing this out. right. this is not a valid file. 

> ./run.sh -t t/xbcloud.sh
> Using /home/rahul.malik/MySQL/build/x7/xtrabackup-test/var as test root
> Detecting server version...
> Running against MySQL 5.7.31-log (InnoDB 5.7.31)
> Using 'xtrabackup' as xtrabackup binary
> Using 8 parallel workers
> 
> ## ==============================================================================
> TEST WORKER RESULT TIME(s) or COMMENT
> # xbcloud w1 [passed] 72
> Spent 72 of 79 seconds executing testcases
> 
> SUMMARY: 1 run, 1 successful, 0 skipped, 0 failed
> 
> See results/ for detailed output
> {smm-12-d40aec x7 o7 xtrabackup-test}$

This result is wrong. Mistakely I executed the test on wrong build.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: xbcloud is different api. so the fix I am thinking is to parse the total number of files written by xtrabackup to xbcloud API and then verify that. for that, we also have to consider additional files written by backup based on the option provided which need some additional testing as well. 

this bug just fixes the regression where we are checking xtrabackup_info instead of xtrabackup_info.gz
IMHO we can create a separate bug and implement these changes with the best approach.

altmannmarcelo: Agree on making it right to consider compression and raise an enhancement issue to make it more reliable.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
satya-bodapati: What is the correct file then?

This is the reason local results cannot be trusted :cry:

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
rahulmalik87: incase of 2.4 last file can be xtrabackup_info or xtrabackup_log. so instead of relying upon the last file I thinking to  iterate through list xtrabackup_info file is present

rahulmalik87:
the test is succesful with this patch. 
./run.sh -t t/xbcloud.sh
Detecting server version...
Running against MySQL 5.7.31-log (InnoDB 5.7.31)
Using 8 parallel workers

xbcloud                                  w1	[passed]    114
Spent 114 of 120 seconds executing test cases

rahulmalik87:
raised https://jira.percona.com/browse/PXB-2332

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -705,9 +705,22 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
-  if (std::count(object_list.begin(), object_list.end(),
-                 backup_name + "/xtrabackup_info.00000000000000000000") == 0) {
+  /* check if the xtrabackup_info.gz.00000000000 or
+  xtrabackup_info.00000000000000000000 is uploaded to cloud storage to
+  determine successful xbcloud "put" operation. */
+  bool file_found = false;
+  if (!object_list.empty()) {
+    std::string last_file_prefix = backup_name + "/xtrabackup_info";
+    auto last_file_size = last_file_prefix.size();
+    for( auto cur_file : object_list) {
```
satya-bodapati:
```suggestion
    for (auto cur_file : object_list) {
```

satya-bodapati: 1. Please fix the commit message
2. Please fix description message

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -705,9 +705,22 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+  xtrabackup_info.00000000000000000000 is uploaded to cloud storage to
+  determine successful xbcloud "put" operation. */
+  bool file_found = false;
+  if (!object_list.empty()) {
+    std::string last_file_prefix = backup_name + "/xtrabackup_info";
+    auto last_file_size = last_file_prefix.size();
+    for (auto cur_file : object_list) {
+      if (cur_file.size() >= last_file_size &&
+	  cur_file.substr(0, last_file_size).compare(last_file_prefix) == 0) {
+	file_found = true;
```
satya-bodapati:
```suggestion
	  file_found = true;
```

satya-bodapati (approved):
LGTM! Please merge after fixing the space issue (sorry, my eyes caught it only now)

altmannmarcelo (approved):
After fixing last @satya-bodapati 's comment. LGTM

PR #978, opened 2020-10-16
2.4 Fixed PXB-2275 (PXB crashes during backup if an encrypted table is cr…

altmannmarcelo:
…eated)

https://jira.percona.com/browse/PXB-2275
TDE encryption happens in the phases:
1) Encryption flag is written to page0 synchronously
2) Encryption information is written on page0 in memory(buffer pool) and
generates a redo log of type MLOG_WRITE_STRING.

PXB accepts taking a backup if we have a case where tablespace has
encryption flag set but is still missing the encryption information.
It tries to validate if the encryption key having the same tablespace ID
has already been parsed from redo logs by accessing
recv_sys->encryption_list. If it cannot find the key it move forward
anyway on the assumption that the key will eventually be parsed via
redo log follow thread.

Problems:
1) We are trying to iterate through recv_sys->encryption_list assuming
it has already been populated, however, it can still be a NULL pointer.
2) The is no guarantee that said redo log will be parsed and and in an
edge case we may finish the backup before the redo event has been parsed
causing the backup to be incomplete.

Fixes:
1) Validate recv_sys->encryption_list before trying to iterate through.
2) Create a vector to store tablespace IDs that have a missing key when
we first attempted to open it. At the end of the backup, after we have
already stopped redo follow thread (we won't receive any new redo log
event) we test the list of saved tablespace IDs to validate if the
encryption key has been read via redo log during the backup. In case we
find any missing key, we abort the backup as it means we won't be able
to decrypt this table.

### DIFF
--- a/storage/innobase/fsp/fsp0file.cc
+++ b/storage/innobase/fsp/fsp0file.cc
@@ -664,17 +664,19 @@ Datafile::validate_first_page(lsn_t*	flush_lsn,
 
                         if (srv_backup_mode) {
 				mutex_enter(&recv_sys->mutex);
-				encryption_list_t::iterator	it;
-				for (it = recv_sys->encryption_list->begin();
-				     it != recv_sys->encryption_list->end();
-				     it++) {
-					if (it->space_id == m_space_id) {
-						memcpy(m_encryption_key,
-						       it->key,
-						       ENCRYPTION_KEY_LEN);
-						memcpy(m_encryption_iv, it->iv,
-						       ENCRYPTION_KEY_LEN);
-						found = true;
+				if (recv_sys->encryption_list != NULL) {
+					encryption_list_t::iterator	it;
+					for (it = recv_sys->encryption_list->begin();
+					     it != recv_sys->encryption_list->end();
+					     it++) {
+						if (it->space_id == m_space_id) {
+							memcpy(m_encryption_key,
+							       it->key,
+							       ENCRYPTION_KEY_LEN);
+							memcpy(m_encryption_iv, it->iv,
+							       ENCRYPTION_KEY_LEN);
+							found = true;
+						}
 					}
 				}
 				mutex_exit(&recv_sys->mutex);
@@ -687,7 +689,11 @@ Datafile::validate_first_page(lsn_t*	flush_lsn,
 				ut_free(m_encryption_iv);
 				m_encryption_key = NULL;
 				m_encryption_iv = NULL;
-				return(DB_CORRUPTION);
+				ib::info() << "Failed to decrypt table " << m_filepath << " with space"
+					<< " id " << m_space_id << ". Will check if encrytion key has been"
+					<< " parsed at the end of backup.";
+				invalid_encrypted_tablespace_ids.push_back(m_space_id);
+				return(DB_PAGE_IS_BLANK);
                         }
 
 		}
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -38,6 +38,7 @@ extern my_bool opt_lock_ddl_per_table;
 
 extern bool use_dumped_tablespace_keys;
 
+extern std::vector<ulint> invalid_encrypted_tablespace_ids;
 /******************************************************************************
 Callback used in buf_page_io_complete() to detect compacted pages.
 @return TRUE if the page is marked as compacted, FALSE otherwise. */
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -461,6 +461,8 @@ char *opt_server_public_key = NULL;
 static void
 check_all_privileges();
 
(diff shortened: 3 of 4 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/122/

both failures are not related to this PR:
 - t/pxb-1936.sh - False positive. Re-ran passed - https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-param-medium/123/BUILD_TYPE=release,Host=asan,xtrabackuptarget=xtradb57/console
 - t/bug1250375.sh - 
```
2020-10-23 17:14:43: run.sh: Server with id=1 has been started on port 28167, socket /tmp/xbtemp.XB7xd2/mysql.sock.9V60lp
ERROR 1213 (40001) at line 1: Deadlock found when trying to get lock; try restarting transaction
```
Error was even before xtrabackup - at run.sh setup phase.

storage/innobase/fsp/fsp0file.cc
```diff
@@ -664,17 +664,19 @@ Datafile::validate_first_page(lsn_t*	flush_lsn,
...
-						       it->key,
-						       ENCRYPTION_KEY_LEN);
-						memcpy(m_encryption_iv, it->iv,
-						       ENCRYPTION_KEY_LEN);
-						found = true;
+				if (recv_sys->encryption_list != NULL) {
+					encryption_list_t::iterator	it;
+					for (it = recv_sys->encryption_list->begin();
+					     it != recv_sys->encryption_list->end();
+					     it++) {
```
satya-bodapati, line 671: c++11 style can be used here?

storage/innobase/fsp/fsp0file.cc
```diff
@@ -687,7 +689,7 @@ Datafile::validate_first_page(lsn_t*	flush_lsn,
 				ut_free(m_encryption_iv);
 				m_encryption_key = NULL;
 				m_encryption_iv = NULL;
-				return(DB_CORRUPTION);
+				return(DB_IO_DECRYPT_FAIL);
```
satya-bodapati: I feel it is better to add to invalid encrypted table list here.

satya-bodapati (changes requested):
Looks good. minor suggestions

storage/innobase/fsp/fsp0file.cc
altmannmarcelo: I wasn´t sure if it was ok to add a new return type. Shall I use same as 8.0 https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/fsp/fsp0file.cc#L703 here ?

storage/innobase/fsp/fsp0file.cc
altmannmarcelo: You mean
```
for (auto &recv_key : *recv_sys->encryption_list) 
```

I don't think 5.7 accepts it. Will check

satya-bodapati: oh right, I forgot..  please use this only in 8.0

storage/innobase/fsp/fsp0file.cc
satya-bodapati: no need to change return type. Just add to vector here.

altmannmarcelo: Yes, makes (wayyy more)  sense. Will fix

satya-bodapati: but the caller would abort I guess. So maybe you are right :laughing:

altmannmarcelo: We can return DB_PAGE_IS_BLANK which will make it continue to operate if we are in the backup phase, and validate_missing_encryption_tablespaces will ensure that we have parsed proper encryption information by the end of backup.

storage/innobase/xtrabackup/test/t/innodb_keyring_file_PXB_2275.sh
```diff
@@ -0,0 +1,111 @@
+. inc/common.sh
+. inc/keyring_file.sh
+
+if ! $XB_BIN --help 2>&1 | grep -q debug-sync; then
+    skip_test "Requires --debug-sync support"
```
satya-bodapati: Please use smaller case letters for filename.
Also innodb is redundant imo.. how about keyring_pxb_2275.sh

storage/innobase/fsp/fsp0file.cc
satya-bodapati: and we really have an all-zero page right (except page_id)? in that case, DB_PAGE_IS_BLANK makes sense

altmannmarcelo: Correct, the flags that are written syncronously are set (page_id, encryption flag, ...), and the ones that are written async (encryption data, ...) are empty. DB_PAGE_IS_BLANK will indeed make it clear.

satya-bodapati: Cool! :+1:

satya-bodapati (approved):

PR #988, opened 2020-10-27
8.0 Fixed PXB-2275 (PXB crashes during backup if an encrypted table is cr…

altmannmarcelo:
…eated)

https://jira.percona.com/browse/PXB-2275
TDE encryption happens in the phases:

Encryption flag is written to page0 synchronously
Encryption information is written on page0 in memory(buffer pool) and
generates a redo log of type MLOG_WRITE_STRING.
PXB accepts taking a backup if we have a case where tablespace has
encryption flag set but is still missing the encryption information.
It tries to validate if the encryption key having the same tablespace ID
has already been parsed from redo logs by accessing
recv_sys->encryption_list. If it cannot find the key it move forward
anyway on the assumption that the key will eventually be parsed via
redo log follow thread.

Problems:

We are trying to iterate through recv_sys->encryption_list assuming
it has already been populated, however, it can still be a NULL pointer.
The is no guarantee that said redo log will be parsed and and in an
edge case we may finish the backup before the redo event has been parsed
causing the backup to be incomplete.
Fixes:

Validate recv_sys->encryption_list before trying to iterate through.
Create a vector to store tablespace IDs that have a missing key when
we first attempted to open it. At the end of the backup, after we have
already stopped redo follow thread (we won't receive any new redo log
event) we test the list of saved tablespace IDs to validate if the
encryption key has been read via redo log during the backup. In case we
find any missing key, we abort the backup as it means we won't be able
to decrypt this table.

### DIFF
--- a/storage/innobase/fsp/fsp0file.cc
+++ b/storage/innobase/fsp/fsp0file.cc
@@ -683,11 +683,13 @@ dberr_t Datafile::validate_first_page(space_id_t space_id, lsn_t *flush_lsn,
 
       if (srv_backup_mode) {
         mutex_enter(&recv_sys->mutex);
-        for (const auto &recv_key : *recv_sys->keys) {
-          if (recv_key.space_id == m_space_id) {
-            memcpy(m_encryption_key, recv_key.ptr, Encryption::KEY_LEN);
-            memcpy(m_encryption_iv, recv_key.iv, Encryption::KEY_LEN);
-            found = true;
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/318/

satya-bodapati (approved):

PR #995, opened 2020-11-02
PXB-2335 Documentation: Prevent back up if Source system (DB) version…

patrickbirch:
… is higher then current PXB version (8.0)

modified:   source/index.rst
added: source/using_xtrabackup/comparison.rst


rahulmalik87 (approved):
LGTM

PR #1012, opened 2020-12-02
PXB-2357 hang in backup with redo log archive

rahulmalik87:
Problem:
    Hang is backup when redo archive log is enabled and there is
    not enough redo log generated at the end of the backup.

    Analysis:
    When the first log block number in the archive file is lower than the current
    log block number, PXB pauses log copying and wait for archived redo log to
    catch up and switched when the block number and checksum is same as redo
    log file checksum.

    The problem is block checksum of main redo file
    doesn't match with archive redo block and it waits for infinite.
    It happens because pxb copies block number and checksum of partial written
    block and checksum which are later overwritten by the server when more redo are
    generated.

    Fix:
    Since checksum of redo file and archive file can  be different,
    switched only based on block number.

### DIFF
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -610,10 +610,14 @@ uint32_t Archived_Redo_Log_Monitor::get_first_log_block_checksum() const {
   return first_log_block_checksum;
 }
 
-void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn, uint32_t no,
-                                               uint32_t checksum) {
+void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn,
+                                               const byte *redo_buf) {
   bool finished = false;
   lsn_t bytes_read = OS_FILE_LOG_BLOCK_SIZE;
+
+  auto redo_block_no = log_block_get_hdr_no(redo_buf);
+  auto redo_block_checksum = log_block_get_checksum(redo_buf);
+  auto redo_block_len = log_block_get_data_len(redo_buf);
 
   while (true) {
     auto len = reader.read_logfile(&finished);
@@ -621,7 +625,14 @@ void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn, uint32_t no,
          ptr += OS_FILE_LOG_BLOCK_SIZE, bytes_read += OS_FILE_LOG_BLOCK_SIZE) {
       auto arch_block_no = log_block_get_hdr_no(ptr);
       auto arch_block_checksum = log_block_get_checksum(ptr);
-      if (no == arch_block_no && checksum == arch_block_checksum) {
+      auto arch_block_len = log_block_get_data_len(ptr);
+      /* When checksum of redo and archive blocks are different, allow PXB to
+      switch to archive if the data length of blocks is different. This can
+      happen when the last block is partially filled in redolog and when it
+      reaches the archive file, the same block could be filled more */
+      if (redo_block_no == arch_block_no &&
+          (redo_block_checksum == arch_block_checksum ||
+           redo_block_len != arch_block_len)) {
         msg("xtrabackup: Archived redo log has caught up\n");
         reader.set_start_lsn(lsn - bytes_read);
         return;
@@ -1034,9 +1045,8 @@ void Redo_Log_Data_Manager::track_archived_log(lsn_t start_lsn, const byte *buf,
 
   if (archived_log_state == ARCHIVED_LOG_NONE) {
     auto no = log_block_get_hdr_no(buf);
-    auto checksum = log_block_get_checksum(buf);
     if (no > archived_log_monitor.get_first_log_block_no()) {
-      archived_log_monitor.skip_for_block(start_lsn, no, checksum);
+      archived_log_monitor.skip_for_block(start_lsn, buf);
       archived_log_state = ARCHIVED_LOG_MATCHED;
     }
   }
--- a/storage/innobase/xtrabackup/src/redo_log.h
+++ b/storage/innobase/xtrabackup/src/redo_log.h
@@ -231,7 +231,7 @@ class Archived_Redo_Log_Monitor {
   uint32_t get_first_log_block_checksum() const;
 
   /** Read archived log until the given log block. */
-  void skip_for_block(lsn_t lsn, uint32_t no, uint32_t checksum);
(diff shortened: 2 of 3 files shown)


rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/365/#showFailuresLink

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -620,8 +620,7 @@ void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn, uint32_t no,
     for (auto ptr = reader.get_buffer(); ptr < reader.get_buffer() + len;
          ptr += OS_FILE_LOG_BLOCK_SIZE, bytes_read += OS_FILE_LOG_BLOCK_SIZE) {
       auto arch_block_no = log_block_get_hdr_no(ptr);
-      auto arch_block_checksum = log_block_get_checksum(ptr);
-      if (no == arch_block_no && checksum == arch_block_checksum) {
+      if (no == arch_block_no) {
         msg("xtrabackup: Archived redo log has caught up\n");
```
satya-bodapati: you should calculate checksum from the archive block data and also from the header. It should match. If it doesn't, thenkeep retrying until checksum matches

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: We already have a check to error out if mismatch in checksum.
https://github.com/percona/percona-xtrabackup/blob/ec3a018bbb85407e026561cbd3705b80e0a42a6e/storage/innobase/xtrabackup/src/redo_log.cc#L533

satya-bodapati: I think you should skip checksum verification only len mismatch

And in the case of length mismatch, you could skip comparison. But there should be retry logic. Release assert(crash) is not a good idea.

rahulmalik87: length mismatch is done.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -610,18 +610,23 @@ uint32_t Archived_Redo_Log_Monitor::get_first_log_block_checksum() const {
...
   while (true) {
     auto len = reader.read_logfile(&finished);
     for (auto ptr = reader.get_buffer(); ptr < reader.get_buffer() + len;
          ptr += OS_FILE_LOG_BLOCK_SIZE, bytes_read += OS_FILE_LOG_BLOCK_SIZE) {
       auto arch_block_no = log_block_get_hdr_no(ptr);
       auto arch_block_checksum = log_block_get_checksum(ptr);
-      if (no == arch_block_no && checksum == arch_block_checksum) {
+      auto arch_block_len = log_block_get_data_len(ptr);
+      if (no == arch_block_no &&
+          (checksum == arch_block_checksum || data_len != arch_block_len)) {
```
satya-bodapati: Better if you can add a comment here. Explain why we are tolerating checksum mismatch.

satya-bodapati (changes requested):
Please give another go to improve commit message :)

Also fix spelling mistakes (lenght -> length)

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: done.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -610,18 +610,26 @@ uint32_t Archived_Redo_Log_Monitor::get_first_log_block_checksum() const {
...
     auto len = reader.read_logfile(&finished);
     for (auto ptr = reader.get_buffer(); ptr < reader.get_buffer() + len;
          ptr += OS_FILE_LOG_BLOCK_SIZE, bytes_read += OS_FILE_LOG_BLOCK_SIZE) {
       auto arch_block_no = log_block_get_hdr_no(ptr);
       auto arch_block_checksum = log_block_get_checksum(ptr);
-      if (no == arch_block_no && checksum == arch_block_checksum) {
+      auto arch_block_len = log_block_get_data_len(ptr);
+      /* When checksum of redo and archive blocks are different, allow PXB to
+      switch to archive if the data lenght of blocks are different. It
+      indicates they are read during different states.*/
```
satya-bodapati:
```suggestion
      /* When checksum of redo and archive blocks are different, allow PXB to
      switch to archive if the data length of blocks is different. This can happen
      when the last block is partially filled in redolog and when it reaches the
      archive file, the same block could be filled more */
```

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -610,18 +610,26 @@ uint32_t Archived_Redo_Log_Monitor::get_first_log_block_checksum() const {
   return first_log_block_checksum;
 }
 
-void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn, uint32_t no,
-                                               uint32_t checksum) {
+void Archived_Redo_Log_Monitor::skip_for_block(lsn_t lsn, const byte *buf) {
```
satya-bodapati: instead of buf, can we use more meaningful name? is it redo block, archive block buffer?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -610,18 +610,26 @@ uint32_t Archived_Redo_Log_Monitor::get_first_log_block_checksum() const {
...
   bool finished = false;
   lsn_t bytes_read = OS_FILE_LOG_BLOCK_SIZE;
 
+  auto no = log_block_get_hdr_no(buf);
+  auto checksum = log_block_get_checksum(buf);
+  auto data_len = log_block_get_data_len(buf);
```
satya-bodapati: may be prefix redo_ to indicate these are for redo log blocks?

satya-bodapati (changes requested):
Almost there..

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: thank you

satya-bodapati (approved):

satya-bodapati:
@rahulmalik87 please improve commit message too "remove the different states and add the content from source file comment"

PR #1023, opened 2020-12-29
PXB-2274 xtrabackup restore failed when there are DML statements runn…

rahulmalik87:
…ing during backup stage.

PXB takes last_lsn from ps.log_status (which is a consistent snapshot across all engines including binlog). However, it was writing the last_checkpoint by querying the last copied checkpoint from the redo thread. This can cause prepare to fail.

Write last_checkpoint and LSN from ps.log_status

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1444,6 +1444,8 @@ bool backup_start(Backup_context &context) {
 
   context.log_status = log_status_get(mysql_connection);
 
+  debug_sync_point("xtrabackup_after_query_log_status");
+
   if (!write_current_binlog_file(mysql_connection)) {
     return (false);
   }
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -97,7 +97,7 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
   return (true);
 }
 
-bool Redo_Log_Reader::find_last_checkpoint_lsn(lsn_t *lsn) {
+bool Redo_Log_Reader::validate_redo_log_file() {
   ulint max_cp_field;
   auto err = recv_find_max_checkpoint(*log_sys, &max_cp_field);
 
@@ -105,11 +105,6 @@ bool Redo_Log_Reader::find_last_checkpoint_lsn(lsn_t *lsn) {
     msg("xtrabackup: Error: recv_find_max_checkpoint() failed.\n");
     return (false);
   }
-
-  log_files_header_read(*log_sys, max_cp_field);
-
-  *lsn = mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_LSN);
-
   return (true);
 }
 
@@ -1176,12 +1171,10 @@ bool Redo_Log_Data_Manager::is_error() const { return (error); }
 
 Redo_Log_Data_Manager::~Redo_Log_Data_Manager() { os_event_destroy(event); }
 
-bool Redo_Log_Data_Manager::stop_at(lsn_t lsn) {
-  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
-  if (last_checkpoint) {
-    msg("xtrabackup: The latest check point (for incremental): '" LSN_PF "'\n",
-        last_checkpoint_lsn);
-  }
+bool Redo_Log_Data_Manager::stop_at(lsn_t lsn, lsn_t checkpoint_lsn) {
+  last_checkpoint_lsn = checkpoint_lsn;
+  msg("xtrabackup: The latest check point (for incremental): '" LSN_PF "'\n",
+      last_checkpoint_lsn);
   msg("xtrabackup: Stopping log copying thread at LSN " LSN_PF ".\n", lsn);
 
   stop_lsn = lsn;
@@ -1189,6 +1182,11 @@ bool Redo_Log_Data_Manager::stop_at(lsn_t lsn) {
   thread.join();
 
   archived_log_monitor.stop();
+
+  /* check if redo log are disabled */
+  if (!reader.validate_redo_log_file()) {
+    return (false);
+  }
 
   scanned_lsn = reader.get_scanned_lsn();
 
@@ -1203,7 +1201,7 @@ bool Redo_Log_Data_Manager::stop_at(lsn_t lsn) {
     return (false);
   }
 
-  return last_checkpoint;
+  return (true);
 }
 
 void Redo_Log_Data_Manager::close() {
--- a/storage/innobase/xtrabackup/src/redo_log.h
+++ b/storage/innobase/xtrabackup/src/redo_log.h
@@ -43,8 +43,8 @@ class Redo_Log_Reader {
   @return true if success. */
   bool find_start_checkpoint_lsn();
 
-  /** Check if redo logs were disabled during the backup.
-   @return true if success. */
+  /** Check if redo logs are disabled during the backup.
+  @return true if redo logs are disabled */
   bool validate_redo_log_file();
 
   /** Get log header. */
--- a/storage/innobase/xtrabackup/src/redo_log.h
+++ b/storage/innobase/xtrabackup/src/redo_log.h
@@ -43,10 +43,9 @@ class Redo_Log_Reader {
   @return true if success. */
   bool find_start_checkpoint_lsn();
 
-  /** Find last checkpoint lsn.
-  @param[out] lsn               last checkpoint lsn
-  @return true if success. */
-  bool find_last_checkpoint_lsn(lsn_t *lsn);
+  /** Check if redo logs were disabled during the backup.
+   @return true if success. */
+  bool validate_redo_log_file();
 
   /** Get log header. */
   byte *get_header() const;
@@ -278,7 +277,7 @@ class Redo_Log_Data_Manager {
   void abort();
 
   /** Stop log copying at specific lsn. */
-  bool stop_at(lsn_t lsn);
+  bool stop_at(lsn_t lsn, lsn_t checkpoint_lsn);
 
   /** Close log file. */
   void close();
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -4052,7 +4052,8 @@ void xtrabackup_backup_func(void) {
     os_thread_sleep(opt_debug_sleep_before_unlock * 1000);
   }
 
-  if (!redo_mgr.stop_at(backup_ctxt.log_status.lsn)) {
+  if (!redo_mgr.stop_at(backup_ctxt.log_status.lsn,
+                        backup_ctxt.log_status.lsn_checkpoint)) {
     exit(EXIT_FAILURE);
   }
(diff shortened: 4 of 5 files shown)


rahulmalik87:
I added 10 sec wait point after querying from log_status so It generate additional redo entried in log file. and all test are successful

https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/3/

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1177,26 +1177,16 @@ bool Redo_Log_Data_Manager::is_error() const { return (error); }
...
-  if (last_checkpoint_lsn > scanned_lsn) {
-    msg("xtrabackup: error: last checkpoint LSN (" LSN_PF
-        ") is larger than last copied LSN (" LSN_PF ").\n",
-        last_checkpoint_lsn, scanned_lsn);
-    return (false);
+  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
+  if (last_checkpoint) {
+    msg("xtrabackup: The checkpoint LSN in redo log file  '" LSN_PF "'\n",
+        last_checkpoint_lsn);
   }
```
satya-bodapati, line 1189: why this check is removed?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4074,8 +4074,9 @@ void xtrabackup_backup_func(void) {
     strcpy(metadata_type_str, "incremental");
     metadata_from_lsn = incremental_lsn;
   }
-  metadata_to_lsn = redo_mgr.get_last_checkpoint_lsn();
-  metadata_last_lsn = redo_mgr.get_stop_lsn();
+
+  metadata_to_lsn = backup_ctxt.log_status.lsn_checkpoint;
```
satya-bodapati: afaik,  this is the core change Instead of using a checkpiont_lsn by query (at alter point of time). we now use the checkpoint_lsn from the sync point (log_status). Is the understanding right?

Also, what will happen to the transactions that happened after checkpoint_lsn but before backup ended. With your approach, do we still copy the extra redo?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: readded with lsn instead of last_checkpoint.

rahulmalik87: readded the check.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: yeah, The understanding is correct. 

metadata_last_lsn would be applied and then rollback if checkpoint LSN is less.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1177,25 +1177,22 @@ bool Redo_Log_Data_Manager::is_error() const { return (error); }
 Redo_Log_Data_Manager::~Redo_Log_Data_Manager() { os_event_destroy(event); }
 
 bool Redo_Log_Data_Manager::stop_at(lsn_t lsn) {
-  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
```
altmannmarcelo, line 1180: Why we moved this check? Any specific reason?

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1443,6 +1443,9 @@ bool backup_start(Backup_context &context) {
                  false);
 
   context.log_status = log_status_get(mysql_connection);
+  os_thread_sleep(10000000);
```
altmannmarcelo: Did you left this by mistake?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1177,25 +1177,22 @@ bool Redo_Log_Data_Manager::is_error() const { return (error); }
...
+  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
+  if (last_checkpoint) {
+    msg("xtrabackup: The checkpoint LSN in redo log file  '" LSN_PF "'\n",
+        last_checkpoint_lsn);
+  }
 
-  scanned_lsn = reader.get_scanned_lsn();
-
-  if (last_checkpoint_lsn > scanned_lsn) {
+  if (lsn > reader.get_scanned_lsn()) {
```
altmannmarcelo: Where is scanned_lsn variable set now? We still use it on below check.

altmannmarcelo (changes requested):
@rahulmalik87  I think we are over complicating things here.

```
  bool last_checkpoint = reader.find_last_checkpoint_lsn(&last_checkpoint_lsn);
```

`reader.find_last_checkpoint_lsn` function is only used here and at the moment it reads the last checkpoint from header of redo log. A much simpler and clean change will be to make this fuction report `backup_ctxt.log_status.lsn_checkpoint` instead of querying live redo. Something like:

```
bool Redo_Log_Reader::find_last_checkpoint_lsn(lsn_t *lsn) {
  if (backup_ctxt.log_status.lsn_checkpoint == 0) {
    msg("xtrabackup: Error: backup_ctxt.log_status.lsn_checkpoint not set.\n");
    return (false);
  }

  *lsn = backup_ctxt.log_status.lsn_checkpoint

  return (true);
}
```

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: reader.find_last_checkpoint_lsn does addition task to let pxb know if redo logs are disabled  if so we error out the backup.  That's why I didn't remove this check.

storage/innobase/xtrabackup/src/backup_copy.cc
rahulmalik87: no , I was doing some testing. Should have created another branch to test. removing it.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: in case we are using redo log: 
https://github.com/percona/percona-xtrabackup/blob/0a93a67b02fe5a523e54ae1995f5e5bc02822b21/storage/innobase/xtrabackup/src/redo_log.cc#L204

if we are using archive file
https://github.com/percona/percona-xtrabackup/blob/0a93a67b02fe5a523e54ae1995f5e5bc02822b21/storage/innobase/xtrabackup/src/redo_log.cc#L515

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: can you please add testcase for this scenario and verify that rollback happens and the data we see is before binlog_pos + checkpoint_lsn?

storage/innobase/xtrabackup/test/t/xb_pause_after_log_status.sh
```diff
@@ -0,0 +1,92 @@
...
+
+  (eval $backup_cmd) &
+  job_pid=$!
+
+  wait_for_xb_to_suspend $pid_file
+
+  xb_pid=`cat $pid_file`
+
+  checksum1=`checksum_table sakila t`
+  $MYSQL $MYSQL_ARGS -Ns -e "INSERT INTO t (a) VALUES (10), (20), (30);" sakila
```
satya-bodapati, line 48: so if you select data from prepared backup dir, 10,20,30 would be rolledback?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_pause_after_log_status.sh
rahulmalik87: redo from this insert won't be applied since they have LSN after we queried from log_status.

storage/innobase/xtrabackup/test/t/xb_pause_after_log_status.sh
```diff
@@ -0,0 +1,103 @@
...
+  vlog "row count in table t is $row_count"
+
+  if [ "$row_count" != 3 ]; then
+   vlog "rows in table t is $row_count when it should be 3"
+   exit -1
+  fi
+
+  checksum2=`checksum_table sakila t`
+
+  $MYSQL $MYSQL_ARGS -Ns -e "select * from t;" sakila
```
satya-bodapati, line 79: stupid me, I thought this is mtr framework. Sorry.
Can you please assert the rows returned are < 4.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4052,7 +4052,8 @@ void xtrabackup_backup_func(void) {
     os_thread_sleep(opt_debug_sleep_before_unlock * 1000);
   }
 
-  if (!redo_mgr.stop_at(backup_ctxt.log_status.lsn)) {
+  if (!redo_mgr.stop_at(backup_ctxt.log_status.lsn,
+                        backup_ctxt.log_status.lsn_checkpoint)) {
```
satya-bodapati, line 4056: is the gtid issue PS-7490 related to this PXB issue?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: no it is  related to  PXB-2375

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -97,19 +97,16 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
   return (true);
 }
 
-bool Redo_Log_Reader::find_last_checkpoint_lsn(lsn_t *lsn) {
+/* Try to open the redo log file. It checks if redo logs were disabled during
+    the backup */
```
satya-bodapati:
```suggestion
/** Tries to open the redo log file. It checks if redo logs were disabled during
the backup */
```

satya-bodapati: Also add @return true if .. , else .. in the comment

satya-bodapati: or completely remove here and add only in declaration place

satya-bodapati (changes requested):
Looks good mostly. Minor comments, please address them soon

storage/innobase/xtrabackup/src/redo_log.h
```diff
@@ -43,10 +43,9 @@ class Redo_Log_Reader {
   @return true if success. */
   bool find_start_checkpoint_lsn();
 
-  /** Find last checkpoint lsn.
-  @param[out] lsn               last checkpoint lsn
-  @return true if success. */
-  bool find_last_checkpoint_lsn(lsn_t *lsn);
+  /** Check if redo logs were disabled during the backup.
+   @return true if success. */
+  bool validate_redo_log_file();
```
satya-bodapati:
```suggestion
  /** Check if redo logs are disabled during the backup.
  @return true if redo logs are disabled */
  bool validate_redo_log_file();
```

satya-bodapati (approved):
Please address the last review comment and then merge

PR #1025, opened 2020-12-29
PXB-2327:  [8.0] Add 32 bit support to bootstrap.sh

Sudokamikaze:
Required for xenial-x32


Sudokamikaze:
I've checked, it's not needed as xtradb80 doesn't support 32-bit

PR #1031, opened 2021-01-11
Fixed PXB-2402 (Fix bootstrap script for PS 8.0.22)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2402

Problem:
Bootstrap script is trying to download glib version 12 for 8.0.22,
however it was compiled and it's named glib 17.

Fix:
Consider versions higher than .22 to download file with glib
version 17.


Sudokamikaze (approved):
LGTM

PR #1032, opened 2021-01-14
Pxb 8  2394 Correct spelling in xbcloud help

rahulmalik87:


satya-bodapati:
@rahulmalik87 shouldn't it be only 2 commits?

satya-bodapati:
after merge, it will be 3

satya-bodapati (commented):
see my above comments..

PR #1033, opened 2021-01-25
Pxb 8023 merge

rahulmalik87:


rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/24/

I am looking for failure xb_perm_basic. It is crash in server because we changing the write permisison on files. trying to reproduce it . will update later here. 

xb_pause_after_log_status. I am going to revert PXB-2274 and create a branch for Manish to testing.

storage/innobase/fil/fil0fil.cc
```diff
@@ -3922,18 +3922,6 @@ void Fil_shard::close_all_files() {
...
-  bool should_validate_space_reference_count = srv_fast_shutdown == 0;
-  DBUG_EXECUTE_IF("buf_disable_space_reference_count_check",
-                  should_validate_space_reference_count = false;);
-
-  if (should_validate_space_reference_count) {
-    auto buffer_pool_references = buf_LRU_count_space_references();
-    for (auto shard : m_shards) {
-      shard->validate_space_reference_count(buffer_pool_references);
-    }
-  }
```
satya-bodapati: 1. why do we need to remove this code? 
2. what is the value of srv_fast_shutdown for PXB (either prepare or backup)?
3. Can this value be changed?

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
rahulmalik87: > 1) why do we need to remove this code?
there was crash during the backup stage because mutex initialisation.  IIUC there should not be anything in buffer pool during backup stage. so I removed this check

2) for backup it is 0 and for prepare it is 1.

satya-bodapati: For 1) Since the ref count is for dropping tablespaces, it is not relevant to PXB.  But you could have disabled with ifndef XTRBACKUP instead of removing it

For 2) Do u know the reason for the difference in value? is it set in PXB code?

rahulmalik87: fixed 1.
it is set in PXB code. for backup we use default value of 0. but incase of prepare we set explicit  [at](https://github.com/percona/percona-xtrabackup/blob/fea8a0ef0e29a010972e944e24536f638c43c9bb/storage/innobase/xtrabackup/src/xtrabackup.cc#L2346). I am not sure why

storage/innobase/fil/fil0fil.cc
```diff
@@ -11344,15 +11344,14 @@ byte *fil_tablespace_redo_encryption(byte *ptr, const byte *end,
...
-      ut_a(found);
     }
+  } else {
+    ulint master_key_id = mach_read_from_4(ptr + Encryption::MAGIC_SIZE);
+    if (Encryption::get_master_key_id() < master_key_id) {
+      Encryption::set_master_key(master_key_id);
+    }
+    bool found = xb_fetch_tablespace_key(space_id, key, iv);
+    ut_a(found);
   }
```
satya-bodapati: Previously, the code says, after decode encryption is successful, we will use xb_fetch_tablespace_key(). Which in turn looks for prefedeined xtrabackup_keys. heh? we do why do this? :thinking: 

I don't understand this change as well. Looks more dangerous to use this without the decoding.

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
rahulmalik87: this is previous [code](https://github.com/percona/percona-xtrabackup/blob/3144bd4c22cced64cc5444b81f7f82187c938a13/storage/innobase/fil/fil0fil.cc#L11222)

I split change in two commits. that's why it look confusing.  xb_fetch_tablespace_key is done with transition-key in that case we don't do decode. we rely on xb_fetch_tablespace_key.

satya-bodapati: Ok, so the conflict resolution moved it to wrong place and this commit moves it to the original place again? (Essentially we brought it to pre-merge state?)?

storage/innobase/fil/fil0fil.cc
```diff
@@ -9599,15 +9561,6 @@ dberr_t fil_set_autoextend_size(space_id_t space_id, uint64_t autoextend_size) {
 @return DB_SUCCESS or error code */
 dberr_t fil_set_encryption(space_id_t space_id, Encryption::Type algorithm,
                            byte *key, byte *iv) {
-<<<<<<< HEAD
-=======
-  ut_ad(space_id != TRX_SYS_SPACE);
-
-  if (fsp_is_system_or_temp_tablespace(space_id)) {
-    return DB_IO_NO_ENCRYPT_TABLESPACE;
-  }
```
satya-bodapati: Do we have PXB test that tests ibdata and ibtmp1, *.ibt encrypted tablespaces?

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
rahulmalik87: yes, we brought the logic to pre conflict state.

storage/innobase/fil/fil0fil.cc
rahulmalik87: test pxb-1793.sh, shared_tablespace_encryption.sh test backup with innodb-default-encryption=on

satya-bodapati:
@rahulmalik87 can u please explain/reason these test failures?

t/pxb-1694.sh
t/xbcloud.sh
t/binlog_info.sh
t/bug1002688.sh
t/bug1206363.sh
t/bug606981.sh
t/bug759701.sh
t/xb_perm_basic.sh
t/xb_perm_stream.sh
t/xb_pause_after_log_status.sh

storage/innobase/fil/fil0fil.cc
satya-bodapati: ok, got it

satya-bodapati:
@rahulmalik87 also can you elaborate on the SDI parsing issue. ie why PXB8022 failed to parse SDI from 8023?

is it just the DD_VERSION bump? or is there an actual DD change introduced?

rahulmalik87:
> @rahulmalik87 can u please explain/reason these test failures?
> 
> t/pxb-1694.sh
> t/xbcloud.sh
> t/binlog_info.sh
> t/bug1002688.sh
> t/bug1206363.sh
> t/bug606981.sh
> t/bug759701.sh
> t/xb_perm_basic.sh
> t/xb_perm_stream.sh
> t/xb_pause_after_log_status.sh

xb_perm_basic failure are because of changing read permission on running server. Server crash during io on doublewrite buffer.
 WIth innodb_wait_for_flush_all before changin the permission on files of data dir, we can fix the test case. If you are fine with I will push the changes?
 
 xb_pause_after_log_status.sh are because of PXB-2274 regression. I am working on it.
 pxb-1694.sh and binlog_info.sh are known failure fore last few releases. 
 
  t/bug1002688.sh
> t/bug1206363.sh
> t/bug606981.sh
> t/bug759701.sh
these are because of upstream [101358](https://bugs.mysql.com/bug.php?id=101358). which are fixed in 8.0.23. so you can only see them with PS which is based on 8.0.22

rahulmalik87:
> @rahulmalik87 also can you elaborate on the SDI parsing issue. ie why PXB8022 failed to parse SDI from 8023?
> 
> is it just the DD_VERSION bump? or is there an actual DD change introduced?

they have introduce autoextend_size=0 in SDI.

rahulmalik87:
xbcloud test https://pxb.cd.percona.com/view/PXB%208.0/job/pxb80-single-platform-run/8/CMAKE_BUILD_TYPE=RelWithDebInfo/console

rahulmalik87:
xbcloud and other test on PS 8.0.22 https://pxb.cd.percona.com/view/PXB%208.0/job/pxb80-single-platform-run/lastCompletedBuild/testReport/

rahulmalik87:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/33/

jenkin results

satya-bodapati (approved):
LGTM! please decide on xb_pause_log_status post merge
Also check if there are extra binaries in the PXB package

PR #1034, opened 2021-01-28
PXB-2280: SELinux and AppArmor policy

kamil-holubicki:
https://jira.percona.com/browse/PXB-2280

Added SELinux and AppArmor configuration files.
Please note that these configurations are not distributed along with
PXB. It is up to the user to set SELinux/AppArmor with the guidance of
the online documentation provided by Percona. The purpose of these files
is to provide simple, working configuration which can be adjusted to the
specific needs.

### DIFF
--- a/packaging/percona/apparmor/apparmor.d/usr.sbin.xtrabackup
+++ b/packaging/percona/apparmor/apparmor.d/usr.sbin.xtrabackup
@@ -1,0 +1,77 @@
+#include <tunables/global>
+
+/usr/bin/xtrabackup {
+  #include <abstractions/base>
+  #include <abstractions/mysql>
+
+  capability dac_override,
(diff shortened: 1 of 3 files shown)


patrickbirch (approved):
LGTM

altmannmarcelo (approved):
File location LGTM.

@hrvojem when testing it, please make sure that things such as `--stream=xbstream` and `--stream=tar` work. I have not see those been whitelisted to be invoked by xtrabackup.

Also, test decompress, I've not seem any mention of `qpress` on the policies.

kamil-holubicki:
Please note that SELinux/AA configuration is not distributed together with PXB. These are just default configuration files to be used along with the online instructions on how to use them to set up SELinux/AA manually.

PR #1038, opened 2021-02-11
PXB-2418 link error. dyld: Library not loaded: libprotobuf-lite.3.11.14

rahulmalik87:
Problem:
PXB complaints about libprotobuf if executed from install path:

Fix
PROTOBUF_LITE_LIBRARY has been removed because PXB doesn't directly use immutable string and we don't want a protobuf.so dependency to use PXB

### DIFF
--- a/sql/CMakeLists.txt
+++ b/sql/CMakeLists.txt
@@ -804,7 +804,7 @@ TARGET_LINK_LIBRARIES(sql_main ${MYSQLD_STATIC_PLUGIN_LIBS}
 # sql/immutable_string.h uses
 # google::protobuf::io::CodedOutputStream::WriteVarint64ToArray
 # which may or may not be inlined.
-TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY})
+#TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY}) have been removed because PXB doesn't directly use immutable string and we don't want a protobuf.so dependency to use PXB
 
 ADD_LIBRARY(sql_gis STATIC ${SQL_GIS_SOURCES})
 ADD_DEPENDENCIES(sql_gis GenServerSource)


sql/CMakeLists.txt
```diff
@@ -804,7 +804,7 @@ TARGET_LINK_LIBRARIES(sql_main ${MYSQLD_STATIC_PLUGIN_LIBS}
 # sql/immutable_string.h uses
 # google::protobuf::io::CodedOutputStream::WriteVarint64ToArray
 # which may or may not be inlined.
-TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY})
+#TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY})
```
satya-bodapati: Please add a comment on why we not linking PROTOBUF. So that future merges we don't uncomment by mistake.

satya-bodapati: may be remove the line

satya-bodapati (changes requested):

sql/CMakeLists.txt
rahulmalik87: removed

satya-bodapati (changes requested):
Please add a comment above removal. Write something like "TARGET_LINK_LIBRARIES(.. ) have been removed because PXB doesn't directly use immutable string and we don't want a protobuf.so dependency to use PXB".

sql/CMakeLists.txt
```diff
@@ -804,7 +804,7 @@ TARGET_LINK_LIBRARIES(sql_main ${MYSQLD_STATIC_PLUGIN_LIBS}
 # sql/immutable_string.h uses
 # google::protobuf::io::CodedOutputStream::WriteVarint64ToArray
 # which may or may not be inlined.
-TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY})
+#TARGET_LINK_LIBRARIES(sql_main ${PROTOBUF_LITE_LIBRARY}) have been removed because PXB doesn't directly use immutable string and we don't want a protobuf.so dependency to use PXB
```
satya-bodapati, line 807: has been removed

satya-bodapati: Now that you know the wording and reason, can you please update the commit message as well?

satya-bodapati (changes requested):

satya-bodapati (approved):

PR #1040, opened 2021-02-16
PXB-2371 - Vulnerability: Path traversal

ldonoso:
https://jira.percona.com/browse/PXB-2371
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/48/
https://pxb.cd.percona.com/view/PXB%208.0/job/pxb80-single-platform-run/21/

Modify xbstream so it creates/extracts streams as gnu tar does.

When creating a stream, the leading directory slashes and relative
directories are removed from the filename. Hence, when the stream is
extracted, the files will be created in the directory specified, e.i.
their paths are relative.

The parameter `-P`, or `--absolute-names`, has been added to allow
absolute paths.

It is an error if there is a file with an absolute path when extracting
a stream and the parameter `-P` has not been specified.

### DIFF
--- a/storage/innobase/xtrabackup/src/file_utils.h
+++ b/storage/innobase/xtrabackup/src/file_utils.h
@@ -1,0 +1,28 @@
+/******************************************************
+Copyright (c) 2021 Percona LLC and/or its affiliates.
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; version 2 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+*******************************************************/
+
+#ifndef FILE_UTILS_H
+#define FILE_UTILS_H
+
+/* Return a safer suffix of FILE_NAME, or "." if it has no safer
+   suffix.  Check for fully specified file names and other atrocities.
+   Warn the user if we do not return NAME. */
+/* Based on gnu tar */
+const char *safer_name_suffix(char const *file_name, int *prefix_len_out);
+
+#endif
--- a/storage/innobase/xtrabackup/src/CMakeLists.txt
+++ b/storage/innobase/xtrabackup/src/CMakeLists.txt
@@ -93,6 +93,7 @@ MYSQL_ADD_EXECUTABLE(xtrabackup
   ds_tmpfile.c
   ds_xbstream.c
   fil_cur.cc
+  file_utils.c
   quicklz/quicklz.c
   read_filt.cc
   write_filt.cc
@@ -147,6 +148,7 @@ MYSQL_ADD_EXECUTABLE(xbstream
   ds_stdout.c
   ds_decrypt.c
   datasink.c
+  file_utils.c
   xbstream.c
   xbstream_read.c
   xbstream_write.c
--- a/storage/innobase/xtrabackup/src/file_utils.c
+++ b/storage/innobase/xtrabackup/src/file_utils.c
@@ -1,0 +1,46 @@
+#include "file_utils.h"
+#include "common.h"
+
+#if defined _WIN32 || defined __CYGWIN__ \
+    || defined __EMX__ || defined __MSDOS__ || defined __DJGPP__
+# define _IS_DRIVE_LETTER(C) \
+    (((C) >= 'A' && (C) <= 'Z') || ((C) >= 'a' && (C) <= 'z'))
+# define HAS_DEVICE(Filename) \
+    (_IS_DRIVE_LETTER ((Filename)[0]) && (Filename)[1] == FN_DEVCHAR)
+# define FILE_SYSTEM_PREFIX_LEN(Filename) (HAS_DEVICE (Filename) ? 2 : 0)
+#else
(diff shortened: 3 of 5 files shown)


rahulmalik87:
we need fix for it in PXB 2.4 as well.

rahulmalik87:
also provide jenknin results. you can use https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -0,0 +1,46 @@
+#include "common.h"
+#include "my_io.h"
+
+#if defined _WIN32 || defined __CYGWIN__ \
```
rahulmalik87: PXB do not support Windows but thats fine if you have include the support.

storage/innobase/xtrabackup/src/file_utils.h
```diff
@@ -0,0 +1,28 @@
...
+
+*******************************************************/
+
+#ifndef FILE_UTILS_H
+#define FILE_UTILS_H
+
+/* Return a safer suffix of FILE_NAME, or "." if it has no safer
+   suffix.  Check for fully specified file names and other atrocities.
+   Warn the user if we do not return NAME. */
+/* Based on gnu tar */
```
rahulmalik87: Please follow MySQL standard.
for example
@param[in,out]  log               redo log
@param[in,out]  contiguous_lsn    log sequence number
                                  until which all redo log has been
                                  scanned
@param[in,out]  to_lsn            LSN to stop recovery at */

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -322,7 +328,8 @@ static int mode_create(int argc, char **argv) {
       goto err;
     }
 
-    file = xb_stream_write_open(stream, filepath, &mystat, NULL, NULL);
+    filepath_dst = opt_absolute_names ? filepath : safer_name_suffix(filepath, &filepath_prefix_len);
```
rahulmalik87: follow clang format. maximum lenght 80.

rahulmalik87 (changes requested):
Could you also write some real test involving xtrabackup. there are lot of test related to xbstream in tes directory storage/innobase/xtrabackup/test/t

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -0,0 +1,46 @@
...
+  char const *p;
+
+  /* Skip file system prefixes, leading file name components that contain
+     "..", and leading slashes.  */
+
+  int prefix_len = FILE_SYSTEM_PREFIX_LEN(file_name);
+
+  // Remove ../
+  for (p = file_name + prefix_len; *p;) {
+    if (p[0] == '.' && p[1] == '.' && (is_directory_separator(p[2]) || !p[2]))
```
kamil-holubicki, line 25: what if the length of the string pointed by file_name is < 3? Wouldn't it cause memory access violation?

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -0,0 +1,46 @@
...
+# define _IS_DRIVE_LETTER(C) \
+    (((C) >= 'A' && (C) <= 'Z') || ((C) >= 'a' && (C) <= 'z'))
+# define HAS_DEVICE(Filename) \
+    (_IS_DRIVE_LETTER ((Filename)[0]) && (Filename)[1] == FN_DEVCHAR)
+# define FILE_SYSTEM_PREFIX_LEN(Filename) (HAS_DEVICE (Filename) ? 2 : 0)
+#else
+# define FILE_SYSTEM_PREFIX_LEN(Filename) ((void) (Filename), 0)
+#endif
+
+const char *safer_name_suffix(char const *file_name, int* prefix_len_out) {
```
kamil-holubicki: Wouldn't it be easier/cleaner/safer?
```
const string safer_name_suffix(const string &file_name, size_t* prefix_len_out) {

    static const string dots("..");
    static const string slash("/");
    auto start = file_name.rfind(dots);

    start = string::npos == start ? 0 : start + dots.length();
    start = file_name.find_first_not_of(slash, start);
    *prefix_len_out = string::npos == start ? file_name.length() : start;

    return file_name.substr(*prefix_len_out);
}
```

storage/innobase/xtrabackup/src/file_utils.cc
ldonoso: Good question Kamil. As it is a c string and `&&` is a short-circuit operator, I guess the first time if finds a `\0` the evaluation of the expression ends and the rest of the array is not traversed.

storage/innobase/xtrabackup/src/file_utils.cc
ldonoso: I agree. But, please, mind I based this code in [gnu tar](https://git.savannah.gnu.org/cgit/paxutils.git/tree/paxlib/names.c#n100) so we can be confident the code is correct. On the other hand, the code in v2.4, and by extension v8.0, is C.

storage/innobase/xtrabackup/src/file_utils.cc
kamil-holubicki: You are right, indeed. If this is always zero-terminated string and its length is 1, p[1] == '\0'

storage/innobase/xtrabackup/src/file_utils.cc
kamil-holubicki: OK, it follows xbstream style which is C in 99.99%. If @rahulmalik87 is OK with this, I'm OK as well. Just wanted to stress out that C++ implementation took me 5 minutes, while I spent 30 minutes analyzing C-version pointers :)

kamil-holubicki (approved):
Approved (please fix issues mentioned by other reviewers)

rahulmalik87 (approved):
LGTM

PR #1043, opened 2021-02-19
PXB-2426 lock-ddl-per-table does not work after lock-ddl was enabled …

rahulmalik87:
…by default

If lock-ddl-per-table is given in the backup command
when lock-ddl is enabled by default, this error is displayed:
Error: --lock-ddl and --lock-ddl-per-table are mutually exclusive

So Modified the error message to
 --lock-ddl and --lock-ddl-per-table are mutually exclusive.
Please specify --lock-ddl=false to use --lock-ddl-per-table

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -6843,8 +6843,8 @@ bool xb_init() {
 
   /* sanity checks */
   if (opt_lock_ddl && opt_lock_ddl_per_table) {
-    msg("Error: %s and %s are mutually exclusive\n", "--lock-ddl",
-        "--lock-ddl-per-table");
+    msg("Error: --lock-ddl and --lock-ddl-per-table are mutually exclusive. "
+        "Please specify --lock-ddl=false to use --lock-ddl-per-table.\n");
     return (false);
   }
 


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6839,14 +6839,8 @@ bool xb_init() {
...
-  if (opt_no_lock || opt_no_backup_locks) opt_lock_ddl = false;
-
-  /* sanity checks */
-  if (opt_lock_ddl && opt_lock_ddl_per_table) {
-    msg("Error: %s and %s are mutually exclusive\n", "--lock-ddl",
-        "--lock-ddl-per-table");
-    return (false);
-  }
+  if (opt_no_lock || opt_no_backup_locks || opt_lock_ddl_per_table)
+    opt_lock_ddl = false;
```
satya-bodapati: What if user provides --lock-ddl and --lock-ddl_per_table at command line? :)

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Also since none of the tests failed, I assume there is no testcase. Please add one just to verify the fix works.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6843,8 +6843,9 @@ bool xb_init() {
 
   /* sanity checks */
   if (opt_lock_ddl && opt_lock_ddl_per_table) {
-    msg("Error: %s and %s are mutually exclusive\n", "--lock-ddl",
-        "--lock-ddl-per-table");
+    msg("Error: %s and %s are mutually exclusive. Please specify "
+        "--lock-ddl=false to use --lock-ddl-per-table.\n",
+        "--lock-ddl", "--lock-ddl-per-table");
```
satya-bodapati: do we need %s?

satya-bodapati (changes requested):

satya-bodapati (approved):

satya-bodapati (approved):

PR #1052, opened 2021-03-10
Add exponential backoff to upload/put retries

timvaillancourt:
This PR causes `S3_client::upload_callback` to wait exponentially between retries. `xbcloud` has maximum retries hardcoded to `3`

Currently `xbcloud` retries without any wait period, retrying as fast as possible. When there is an intermittent problem this usually causes the retry logic to fail to fix the problem

For now a hardcoded wait period of 1000ms is used and grows exponentially on each retry. I plan to create a 2nd PR to add command line flags to tune these settings

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/s3.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/s3.cc
@@ -541,7 +541,7 @@ void S3_client::upload_callback(
     S3_client *client, std::string bucket, std::string name, Http_request *req,
     Http_response *resp, const Http_client *http_client, Event_handler *h,
     S3_client::async_upload_callback_t callback, CURLcode rc,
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
Hi @timvaillancourt .

Thank you so much for your PR and contribution. This work will be performed over #1111 where we also extended the retries to Swift upload/download and S3 download. We have also implemented exponential backoff with some further configuration as the number of retries and maximum exponential backoff sleep upper mark.

We really appreciate your contribution in making the product better.

timvaillancourt:
@altmannmarcelo thanks so much for making this stop-gap fix much more robust. Since deploying this patch our new problem is the lack of retries in the "get" operation (fails the whole download often), so this is very helpful!

PR #1065, opened 2021-03-30
Increased size of binlog_pos field on xtrabackup_history table

acampoh:
We have found that in our systems, sometimes xtrabackup fails silently when the binlog_pos data is bigger than the column size, so as we rely on this table to get the last LSN for incremental backups, all of them are failing. We have manually done an alter on this table increasing the size of this field so it doesn't complain anymore.

We are using percona xtradb clusters and as sometimes the GTID can get so long depending on how many servers acted as master you can go up further than 128 characters. For instance:

filename 'binlog.000256', position '234', GTID of the last change '1b0bd4a1-d4b5-11e9-a234-246e968699a0:1-21387,
ab258228-0993-ee17-4f8a-71ca3cf400e8:1-133171360'

so we propose to make this field a varchar(255) to give support to xtradb clusters with at least 3 nodes.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -1862,7 +1862,7 @@ write_xtrabackup_info(MYSQL *connection)
 		"start_time TIMESTAMP NULL DEFAULT NULL,"
 		"end_time TIMESTAMP NULL DEFAULT NULL,"
 		"lock_time BIGINT UNSIGNED DEFAULT NULL,"
-		"binlog_pos VARCHAR(128) DEFAULT NULL,"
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
Hi @acampoh . Thank you for your pull request.

This issue is related to https://jira.percona.com/browse/PXB-1462 . I've had a pull request for this, but some tests were unstable and I never got the chance to validate them and merge the PR.

I've adjusted the tests and they are all passing now on all matrix of platforms and mysql versions.
Also, I've adjusted xtrabackup to also upgrade from previous versions of the schema (eg.: if you had a table originally created with 128 char long).

Once again thanks a lot for your contribuition. Closing in favor of #1069

PR #1067, opened 2021-04-01
PXB-2455 xtrabackup prepare fails if checkpoint lsn is more than last…

rahulmalik87:
… lsn

Problem:
Backups taken before fix of pxb-2274 can have checkpoint lsn more than
the last LSN from backup.
This could happen because checkpoint lsn were copied from the redo file
instead of querying from log_status.

Fix:
set last_lsn to checkpoint_lsn if it is more than last_lsn


rahulmalik87:
jenkin results (pending atm) https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/54/

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2312,6 +2312,12 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
                                                      : incremental_last_lsn;
   }
 
+  /* backup taken prior to 8.0.23 can have checkoint lsn less more than last
+   * lsn. if so use checkpoint_lsn instead of last_lsn */
```
satya-bodapati: what is less more than? :wink: 
Please fix

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2312,6 +2312,12 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
...
+  if (metadata_last_lsn < metadata_to_lsn) metadata_last_lsn = metadata_to_lsn;
+  if (incremental_last_lsn < incremental_to_lsn)
+    incremental_last_lsn = incremental_to_lsn;
```
satya-bodapati: Also wondering, if we can we do this for backups less than 8.0.23 only?

satya-bodapati: I would love to see a testcase. may be use zipped datadir?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: created the test case.

rahulmalik87: any suggestion how to do?

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2306,10 +2306,19 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
...
 
   lsn_t to_lsn = LSN_MAX;
+
   if (for_apply_log && (metadata_type == METADATA_FULL_BACKUP ||
                         xtrabackup_incremental_dir != nullptr)) {
-    to_lsn = (xtrabackup_incremental_dir == nullptr) ? metadata_last_lsn
-                                                     : incremental_last_lsn;
+    /* backup taken prior to 8.0.23 can have checkoint lsn more than last
+      lsn. if so use checkpoint_lsn instead of last_lsn */
+    if (xtrabackup_incremental_dir == nullptr) {
```
```suggestion
    /* Backups taken prior to 8.0.23 can have checkoint lsn more than last
    copied lsn. If so, use checkpoint_lsn instead of last_lsn */
    if (xtrabackup_incremental_dir == nullptr) {
```

storage/innobase/xtrabackup/test/t/pxb-2455.sh
```diff
@@ -0,0 +1,60 @@
...
+# 2. check data
+#
+
+require_server_version_higher_than 8.0.22
+
+start_server
+stop_server
+function backup_local() {
+	xtrabackup --backup --debug-sync=xtrabackup_after_query_log_status --target-dir=$topdir/backup
+}
```
satya-bodapati: do we need backup_local()? we are used already backed up dir right?

satya-bodapati (changes requested):
Can you please test this scenario (using the debug xtrabackup_after_query_log_status)
1. full backup on 8.0.22
2. upgrade to 8.0.23
3. take incremental on 8.0.23

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Thanks

rahulmalik87:
jenkin results https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium/56/

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: This review comment is not addressed (may be lost). Please address

satya-bodapati (changes requested):
Please submit job on upstream 8.0.23 as well?
How do you plan to deal two different datadirs (PS vs upstream)?
may be have both zipped and select based on PS vs upstream?
or upstream datadir should be compatible should be sufficient for both (PS&upstream)? Please check

satya-bodapati (approved):

PR #1075, opened 2021-04-27
PXB-2443: Version check fails with apparmor profile

kamil-holubicki:
https://jira.percona.com/browse/PXB-2443

Allowed lsb_release execution by Percona Toolkit (PUx to avoid
configuration of PT dependencies)


mchawla16 (approved):
Looks good, thanks Kamil

PR #1077, opened 2021-04-30
PXB-2468 - upstream 8.0.24 & 8.0.25 merge

altmannmarcelo:
https://jira.percona.com/browse/PXB-2468

PXB-2468 - upstream 8.0.24 & 8.0.25 merge.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/34/ 

No new failures compared with trunk (https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/33/)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4060,7 +4060,7 @@ void xtrabackup_backup_func(void) {
 
   if (opt_debug_sleep_before_unlock) {
     msg_ts("Debug sleep for %u seconds\n", opt_debug_sleep_before_unlock);
-    os_thread_sleep(opt_debug_sleep_before_unlock * 1000);
```
rahulmalik87: so you corrected the code to use seconds. previously it wasn't using .01 seconds.

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -61,6 +62,43 @@ const char *BINLOG_KEY_MAGIC = "MYSQL-BINARY-LOG";
...
+  /* Add module specific initialization here */
+  if (innobase::encryption::init_keyring_services(reg_srv) == false) {
+    msg("xtrabackup: init_keyring_services failed\n");
+    cleanup();
+    return false;
+  }
+  msg("xtrabackup: xb_intitialize_service_handles suceeded\n");
+  return true;
+}
+/** Deinitialize compoent service handles */
```
rahulmalik87: component

sql/dd/impl/sdi.cc
```diff
@@ -379,6 +379,7 @@ Sdi_type serialize(const Tablespace &tablespace) {
 bool CheckDefaultCompatibility(const RJ_Document &doc) {
   assert(doc.HasMember("mysqld_version_id"));
 
+#if !defined(XTRABACKUP)
```
satya-bodapati: why do we need this ifndef? Don't we need this for SDI deserialize version check?

sql/dd/impl/sdi.cc
```diff
@@ -387,57 +388,33 @@ bool CheckDefaultCompatibility(const RJ_Document &doc) {
...
   uint dd_version = dd_version_val.GetUint();
   if (dd_version != Dictionary_impl::get_target_dd_version()) {
     // Incompatible change
     my_error(ER_IMP_INCOMPATIBLE_DD_VERSION, MYF(0), dd_version,
              Dictionary_impl::get_target_dd_version());
     return true;
   }
-#endif
+#endif /* !XTRABACKUP */
```
satya-bodapati: are we suppressing the DD version check during SDI deserialization? (ie during prepare operation)?

satya-bodapati (changes requested):

sql/dd/impl/sdi.cc
altmannmarcelo: as per slack conversation. PXB skip checking if mysql_version / dd/ sdi of source are higher than pxb version. We let the code assert on later stage.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2331,10 +2331,12 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
...
-                              false, DICT_ERR_IGNORE_NONE);
-    dict_persist->table_buffer = UT_NEW_NOKEY(DDTableBuffer());
+    if (dict_sys->dynamic_metadata == nullptr)
+      dict_sys->dynamic_metadata =
+          dd_table_open_on_name(NULL, NULL, "mysql/innodb_dynamic_metadata",
+                                false, DICT_ERR_IGNORE_NONE);
+    if (dict_persist->table_buffer == nullptr)
+      dict_persist->table_buffer = UT_NEW_NOKEY(DDTableBuffer());
     srv_dict_recover_on_restart();
   }
```
satya-bodapati: I see that upstream commit,  it only moves the dd_table_buffer change little earlier. Can you please share the re-opening stack trace? Aslo check how it worked in previous version? was there any close table before reuse?

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
```diff
@@ -6893,6 +6893,9 @@ bool Fil_shard::space_extend(fil_space_t *space, page_no_t size) {
       log_write_up_to() is not needed here, because no file shrinks and
       duplicate extending is allowed. And log_write_up_to() here helps
       nothing for fallocate() inconsistency. */
+#if defined(XTRABACKUP)
+      log_write_up_to(*log_sys, mtr.commit_lsn(), true);
+#endif /* XTRABACKUP */
```
satya-bodapati: When you have time, can you please check why the pages in flush_list are not flushed without this change?

Guesses:
1. Pages in flush list do not have proper LSN set?
2. are we creating pages without writing redo?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -71,7 +109,25 @@ static void init_plugins(int argc, char **argv) {
   memcpy(t_argv, argv, sizeof(char *) * t_argc);
 
   mysql_optional_plugins[0] = 0;
-  mysql_mandatory_plugins[0] = 0;
+
+  /* initialize only daemon_keyring_proxy_plugin */
+  if (initialize_keyring_proxy) {
```
satya-bodapati: Do we need this bool variable? ie. Do we call init_plugins() twice?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: Since now we saving the metadata after the redo apply phase rather updating the in memory objects. Please check if it is interfering with  PXB-2180 fix.

rahulmalik87 (changes requested):

storage/innobase/xtrabackup/src/keyring_plugins.cc
altmannmarcelo: Not required. Removed

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Correct. The only test using it is https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/test/t/bug1180922.sh

storage/innobase/xtrabackup/src/keyring_plugins.cc
altmannmarcelo: Fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: @rahulmalik87 the upstream bug is meant to fix an issue while the server is doing crash recovery. While the server is running and generating MLOG_TABLE_DYNAMIC_META the workflow continues the same. Server generates MLOG_TABLE_DYNAMIC_META and convert them to proper MLOG_INSERT/DELETE... at checkpoint. For PXB we need to still ignore MLOG_TABLE_DYNAMIC_META if we are not in the last incremental backup apply. However, I will ask @mchawla16 to test this again with 8.0.24

altmannmarcelo: @satya-bodapati :

>  Can you please share the re-opening stack trace? 

```
InnoDB: Assertion failure: dict0dict.cc:1886:table->get_ref_count() == 0
InnoDB: thread 140139586184512InnoDB: We intentionally generate a memory trap.
InnoDB: Submit a detailed bug report to https://jira.percona.com/projects/PXB.
InnoDB: http://dev.mysql.com/doc/refman/8.0/en/forcing-innodb-recovery.html
12:58:40 UTC - mysqld got signal 6 ;
Most likely, you have hit a bug, but this error can also be caused by malfunctioning hardware.
Thread pointer: 0x0
terribly wrong...
stack_bottom = 0 thread_stack 0x46000
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(my_print_stacktrace(unsigned char const*, unsigned long)+0x55) [0x55fe144a9846]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(handle_fatal_signal+0x2a8) [0x55fe12bd3945]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0) [0x7f74cae323c0]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb) [0x7f74ca47a18b]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x12b) [0x7f74ca459859]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(ut_dbg_assertion_failed(char const*, char const*, unsigned long)+0x150) [0x55fe13324d8b]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(+0x3b7462d) [0x55fe12e6a62d]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(dict_table_remove_from_cache(dict_table_t*)+0x1d) [0x55fe12e6add2]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(dict_close()+0x1ee) [0x55fe12e72cec]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(srv_shutdown()+0x4f1) [0x55fe1328e5ae]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(+0x3092402) [0x55fe12388402]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(+0x309e916) [0x55fe12394916]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(main+0x7bb) [0x55fe123972a3]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x7f74ca45b0b3]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(_start+0x2e) [0x55fe12380dae]
```

> Aslo check how it worked in previous version? 

We always close the table at srv_shutdown, but before we never enter the place where we would open it on `srv_start` because `srv_log_file_size_requested != srv_log_file_size` and `srv_n_log_files_found != srv_n_log_files` will both evaluate to false, never entering the palce we would open it, thus having only one reference count to innodb_dynamic_metadata.

altmannmarcelo:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-test-param/36/ - test results after keyring component functionality.

components/keyrings/keyring_file/keyring_file.cc
```diff
@@ -183,10 +185,14 @@ bool set_paths(const char *component_path, const char *instance_path) {
                   Existing data remains as it is.
 */
 bool init_or_reinit_keyring() {
+#ifndef XTRABACKUP
   /* Get config */
   std::unique_ptr<Config_pod> new_config_pod;
   if (keyring_file::config::find_and_read_config_file(new_config_pod))
     return true;
+#else
+  std::unique_ptr<Config_pod> new_config_pod = std::move(xtrabackup::components::new_config_pod);
```
rahulmalik87: exceeding 80 characters.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1245,6 +1248,11 @@ struct my_option xb_client_options[] = {
      &opt_generate_transition_key, &opt_generate_transition_key, 0, GET_BOOL,
      NO_ARG, 0, 0, 0, 0, 0, 0},
 
+     {"keyring-file-data", OPT_KEYRING_FILE_DATA,
```
rahulmalik87: what is the use of opt_keyring_file_data? I can't see it is being used anywhere

rahulmalik87 (changes requested):
I am getting linker error on compiling code
53/437] Linking CXX shared module plugin_output_directory/component_keyring_file.so
FAILED: plugin_output_directory/component_keyring_file.so
: && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -std=c++14 -fno-omit-frame-pointer -ftls-model=initial-exec  -Wall -Wextra -Wformat-security -Wvla -Wundef -Wmissing-format-attribute -Woverloaded-virtual -Wcast-qual -Wno-null-conversion -Wno-unused-private-field -Wconditional-uninitialized -Wdeprecated -Wextra-semi -Wheader-hygiene -Wnon-virtual-dtor -Wundefined-reinterpret-cast -Winconsistent-missing-destructor-override -Winconsistent-missing-override -Wshadow-field -DSAFE_MUTEX -DENABLED_DEBUG_SYNC -g -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk -bundle -Wl,-headerpad_max_install_names -L/usr/local/opt/openssl/lib -o plugin_output_directory/component_keyring_file.so components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_encryption_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_generator_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_load_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_keys_metadata_iterator_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_metadata_query_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_reader_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/service_implementation/keyring_writer_service_definition.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/backend/backend.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/config/config.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/keyring_file.cc.o components/keyrings/keyring_file/CMakeFiles/component_keyring_file.dir/component_callbacks.cc.o  archive_output_directory/libkeyring_common.a  library_output_directory/./libssl.dylib  library_output_directory/./libcrypto.dylib  components/library_mysys/liblibrary_mysys.a && cd /Users/rahulmalik/MySQL/src/x8/bld/components/keyrings/keyring_file && install_name_tool -change libcrypto.1.1.dylib @loader_path/libcrypto.1.1.dylib /Users/rahulmalik/MySQL/src/x8/bld/plugin_output_directory/component_keyring_file.so && install_name_tool -change libssl.1.1.dylib @loader_path/libssl.1.1.dylib /Users/rahulmalik/MySQL/src/x8/bld/plugin_output_directory/component_keyring_file.so
Undefined symbols for architecture x86_64:
  "xtrabackup::components::new_config_pod", referenced from:
      keyring_file::init_or_reinit_keyring() in keyring_file.cc.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[62/437] Building CXX object storage/innobase/CMakeFiles/innobase.dir/fil/fil0fil.cc.o
ninja: build stopped: subcommand failed.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Ok, understood. So we have metadata table opening in srv_start() that could open the tables. Ok with != nullptr checks.

One concern is that srv_dict_meta->store() is applying changes to metadata_table(inside mysql.ibd) directly. This can happen for all prepares. Should we ensure that this happens only on last incremental backup prepare?

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -71,7 +108,20 @@ static void init_plugins(int argc, char **argv) {
...
+      exit(EXIT_FAILURE);
+    }
+    for (st_mysql_plugin **plugin = mysql_mandatory_plugins; *plugin;
+         plugin++) {
+      if (strcmp((*plugin)->name, "daemon_keyring_proxy_plugin") == 0) {
+        mysql_mandatory_plugins[0] = *plugin;
+        mysql_mandatory_plugins[1] = 0;
+        break;
+      }
+    }
```
satya-bodapati: So keyring_proxy is required only if component keyring cannot be initialized?

storage/innobase/xtrabackup/src/keyring_components.cc
```diff
@@ -0,0 +1,214 @@
+/******************************************************
```
rahulmalik87: can you format this file using clang-format. MySQL also using same standard for formatting.

storage/innobase/xtrabackup/test/run.sh
```diff
@@ -846,6 +860,7 @@ recognized for compatibility";;
 done
 
 set_vars
+config_local_components
```
rahulmalik87: what is the condition for calling config_local_components?

altmannmarcelo: we call it regardless of any condition. If testing on a server prior to 8.0.24 this change won't take effect. If calling it on a server from 8.0.24 forward, the server will load the components (if configured by tests) from datadir instead of bin & plugin. This is called global config vs local config. It's necessary otherwise enabling the keyring component in one test would enable it on all tests running in parallel, since they share the same bin & plugin folders.

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -719,6 +720,11 @@ bool get_mysql_vars(MYSQL *connection) {
     mysql_free_result(res);
   }
 
+  if(server_version >= 80024)
```
satya-bodapati: if (

storage/innobase/xtrabackup/src/keyring_components.cc
```diff
@@ -0,0 +1,214 @@
...
+  if (reg_srv != nullptr) {
+    mysql_plugin_registry_release(reg_srv);
+  }
+}
+static bool initialize_manifest_file_components(std::string components) {
+  g_deployed_components =
+      new (std::nothrow) Deployed_components(my_progname, components);
+  if (g_deployed_components == nullptr ||
+      g_deployed_components->valid() == false) {
+    /*Error would have been raised by Deployed_components constructor */
```
satya-bodapati: /* Error

storage/innobase/xtrabackup/src/keyring_components.h
```diff
@@ -0,0 +1,63 @@
...
+    @param [in]  connection       MYSQL connection to running server
+
+    @return false in case of error, true otherwise
+    @keyring_component_initialized false keyring component not initialized
+    @keyring_component_initialized true  keyring component initialized
+  */
+bool keyring_init_online(MYSQL *connection);
+
+/**
+  Initialize Keyring component by reading xtrackup_componenet file
```
satya-bodapati: typo component

storage/innobase/xtrabackup/src/keyring_components.h
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+  Write backup keyring component config file
+
+  @return false in case of error, true otherwise
+*/
+bool write_componet_config_file();
+
+  /**
+    Initialize Keyring component by querying config from a running server
```
satya-bodapati: Would be nice if you can explain when PXB uses online vs offline

storage/innobase/xtrabackup/test/inc/keyring_common.sh
```diff
@@ -8,7 +8,11 @@ function test_do()
 {
 	table_options=$1
 	transition_key=$2
-
+  keyring_type=$3
+  vlog "-- running test with: -- "
+  vlog "-- table_options: ${table_options} --"
+  vlog "-- transition_key: ${transition_key} --"
+  vlog "-- keyring_type: ${keyring_type} --"
```
satya-bodapati: alignment looks out of place, should it be tabs?

storage/innobase/xtrabackup/test/inc/keyring_common.sh
```diff
@@ -95,9 +103,15 @@ EOF
 		   $prepare_options
 
 	# check that stats works
-	xtrabackup --stats --datadir=$topdir/backup --plugin-load=${plugin_load}
+  if [ -z ${plugin_load+x} ]; then
```
satya-bodapati: what is the +x for?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -1,6 +1,6 @@
 /******************************************************
 hot backup tool for InnoDB
-(c) 2009-2015 Percona LLC and/or its affiliates
+(c) 2009-2021 Percona LLC and/or its affiliates
```
satya-bodapati: Thanks for reminding me :) we should write a git commit hook that verifies this.
ie. list of modified files should have current year

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -109,11 +75,8 @@ static void init_plugins(int argc, char **argv) {
 
   mysql_optional_plugins[0] = 0;
 
-    /* Initialize component service handles */
-    if (xb_intitialize_service_handles() == false) {
-      msg("xtrabackup: Error: failed to component service handles\n");
-      exit(EXIT_FAILURE);
-    }
+  /* initialize only daemon_keyring_proxy_plugin */
```
satya-bodapati: comment alignment

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: didn't we have this option name already? :thinking: how did we pass keyring_file_data before?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7401,6 +7426,15 @@ void setup_error_messages() {
...
 
+void xb_set_plugin_dir()
+{
+  if (opt_xtra_plugin_dir != NULL) {
+    strncpy(opt_plugin_dir, opt_xtra_plugin_dir, FN_REFLEN);
+  } else {
+    strcpy(opt_plugin_dir, PLUGINDIR);
+  }
+}
+
```
satya-bodapati: is this required for component? we didn't use this until now?

satya-bodapati (changes requested):
Still getting hang of it :)

storage/innobase/xtrabackup/test/run.sh
satya-bodapati: regarding this global vs local config, where do we create the component config files? Should the global config file be created at a specific location?

CMakeLists.txt
```diff
@@ -1603,11 +1603,8 @@ ELSE()
   ENDIF()
 ENDIF()
 
-IF(WITHOUT_SERVER)
-  ADD_SUBDIRECTORY(components/libminchassis)
-ELSE()
+IF(NOT WITHOUT_SERVER)
   CONFIGURE_COMPONENTS()
-  ADD_SUBDIRECTORY("components/libminchassis")
 ENDIF()
```
satya-bodapati: Dont we need libminchassis for components?

storage/innobase/fil/fil0fil.cc
altmannmarcelo: As per slack conversation, the issue comes when we call `fil_op_write_space_extend` this is generating the extra redo that causes the hang at shutdown. We will skip the whole block for pxb.

storage/innobase/xtrabackup/test/run.sh
altmannmarcelo: @satya-bodapati global files are read as follow by the server:
bin/mysqld.my - this is the global file for mysqld, this file will be shared from all instances, so if we put the config to read the componenet here, all tests invoking mysqld from the same location (which is what pxb / mysql mtr does) will have it enabled. Instead we configure it to read the instance datadir. Same goes to component config. Once the componenet is loaded via the instance manifest file, we will read the componenet specific file, which is located at the same folder where the .so file is located, has to have the same name but instead of ending with .so it ends with .cnf. We do the same here, we instruct it globally to read from the datadir. So this way we have control on a per-test basis in regards which test will have it loaded or not. This is the same way upstream does for MTR on 8.0.24.

storage/innobase/xtrabackup/src/keyring_plugins.cc
altmannmarcelo: Correct, this part changed a bit on the latest commit with the keyring component. On server side, only one keyring type can be enabled, either component or plugin. If the server has the component enabled, keyring plugin will be loaded but not used. On PXB we follow the same rule - https://github.com/altmannmarcelo/percona-xtrabackup/blob/8.0-PXB-2468/storage/innobase/xtrabackup/src/xtrabackup.cc#L3933 we only go to initialize plugins part if `xtrabackup::components::keyring_component_initialized` is false

storage/innobase/xtrabackup/test/inc/keyring_common.sh
altmannmarcelo: This is to validate the variable is set, will evaluate to nothing is plugin_load is not set.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: @satya-bodapati we did have it before as part of the keyring plugin, if the keyring_plugin.so is read parsing that parameter will result on opt_keyring_file_data been available, however, to keep the same parameters for keyring component (when the plugin is not read) we need to make pxb understand that command line parameter.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: We used this at plugin_init at - https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/src/keyring_plugins.cc#L76 . Since we will also need this at compoenent to make it read the component_keyring_file.so from the correct directory, I've made this code a function and I call it at https://github.com/altmannmarcelo/percona-xtrabackup/blob/8.0-PXB-2468/storage/innobase/xtrabackup/src/xtrabackup.cc#L7487 so the pluging dir is already set when we init components and plugins.

storage/innobase/xtrabackup/src/keyring_components.h
altmannmarcelo: online is used when the mysql server is online - in summary, read the config from ps.keyring_component_status - used at backup
offline is used when we required reading the config after the backup, saved ad xtrabackup_keyring_componenet.cnf (mimics the manifest file the server reads at startup) - used at prepare, stats, move/copy back.

storage/innobase/xtrabackup/test/inc/keyring_common.sh
altmannmarcelo: @satya-bodapati file itself is wrong, we should be using two spaces instead of tabs at 8.0 . Shall we fix the whole file ?

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -62,6 +63,42 @@ constexpr size_t BINLOG_KEY_MAGIC_SIZE = sizeof("MYSQL-BINARY-LOG");
...
+    return false;
+  }
+
+  /* Add module specific initialization here */
+  if (innobase::encryption::init_keyring_services(reg_srv) == false) {
+    msg("xtrabackup: init_keyring_services failed\n");
+    cleanup();
+    return false;
+  }
+  msg("xtrabackup: xb_intitialize_service_handles suceeded\n");
```
satya-bodapati: typo: intitialize

sql/sql_component.cc
```diff
@@ -121,6 +121,22 @@ Deployed_components::Deployed_components(const std::string program_name)
     components_.clear();
   }
 }
+#ifdef XTRABACKUP
+Deployed_components::Deployed_components(const std::string program_name, std::string component)
```
satya-bodapati: I think second parameter should be named as component_names? and it can be a "," separated list. May be just provide an example.
Do we ever load multiple components during PXB online/offline init?

sql/sql_component.cc
```diff
@@ -234,6 +250,34 @@ bool Deployed_components::load() {
...
+#ifdef XTRABACKUP
+bool Deployed_components::load_from_memory()
+{
+  std::vector<const char *> urns;
+  auto free_memory = [&urns]() {
+    for (auto element : urns)
+      if (element != nullptr) my_free(const_cast<char *>(element));
+  };
+
+  if (make_urns(urns) == false) {
```
satya-bodapati: what if the component string is passed ""?

sql/sql_component.h
```diff
@@ -91,6 +91,9 @@ class Sql_cmd_uninstall_component : public Sql_cmd {
 class Deployed_components final {
  public:
   explicit Deployed_components(const std::string program_name);
+#ifdef XTRABACKUP
+  explicit Deployed_components(const std::string program_name, std::string component);
```
satya-bodapati: rename second parameter as "component_names"

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1514,7 +1527,11 @@ bool backup_finish(Backup_context &context) {
   if (!mysql_slave_position.empty() && opt_slave_info) {
     msg("MySQL slave binlog position: %s\n", mysql_slave_position.c_str());
   }
-
+  if (xtrabackup::components::keyring_component_initialized &&
+      !xtrabackup::components::write_componet_config_file()) {
```
satya-bodapati: fix function name typo: componet

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1514,7 +1527,11 @@ bool backup_finish(Backup_context &context) {
...
+    msg("xtrabackup: write_componet_config_file failed\n");
```
satya-bodapati: typo: componet

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2125,11 +2143,17 @@ bool copy_back(int argc, char **argv) {
...
     fclose(f);
-
-    if (!xb_keyring_init_for_copy_back(argc, argv)) {
+    if(!xtrabackup::components::keyring_init_offline())
+    {
+      msg("xtrabackup: Error: failed to init keyring component\n");
+      return (false);
+    }
+    if (!xtrabackup::components::keyring_component_initialized &&
+        !xb_keyring_init_for_copy_back(argc, argv)) {
```
satya-bodapati: can we refactor this a bit? move everything inside the init_for_copy_back() ?

satya-bodapati: move this xtrabackup::components::intitialize_service_handles() inside copy_back() too

storage/innobase/xtrabackup/src/keyring_components.h
```diff
@@ -0,0 +1,63 @@
...
+/** Data types */
+extern std::unique_ptr<keyring_file::config::Config_pod> new_config_pod;
+extern bool keyring_component_initialized;
+
+/**
+  Write backup keyring component config file
+
+  @return false in case of error, true otherwise
+*/
+bool write_componet_config_file();
```
satya-bodapati: typo: componet

storage/innobase/xtrabackup/src/keyring_components.h
satya-bodapati: Please add this documentation to the two functions.

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -932,5 +884,5 @@ void xb_keyring_shutdown() {
   }
 
   free_list(opt_plugin_load_list_ptr);
-  xb_deinitialize_service_handles();
+  xtrabackup::components::deinitialize_service_handles();
```
satya-bodapati: unconditional call is OK? sometimes we load only plugin and not component?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3897,15 +3905,37 @@ void xtrabackup_backup_func(void) {
...
+  }
 
-  if (!xb_keyring_init_for_backup(mysql_connection)) {
-    msg("xtrabackup: Error: failed to init keyring plugin\n");
+  /* create target dir if not exist */
+  if (!my_stat(xtrabackup_target_dir, &stat_info, MYF(0)) &&
+      (my_mkdir(xtrabackup_target_dir, 0777, MYF(0)) < 0)) {
+    msg("xtrabackup: Error: cannot mkdir %d: %s\n", my_errno(),
+        xtrabackup_target_dir);
     exit(EXIT_FAILURE);
```
satya-bodapati: any reason for moving them earlier?

storage/innobase/xtrabackup/test/inc/keyring_common.sh
satya-bodapati: well better to adjust the current file. Aslo the rule for 8.0 is mostly for C++ sources (clang-format). sh files we are free to use any style.

storage/innobase/xtrabackup/test/run.sh
satya-bodapati: Thanks for the detailed explanation!

satya-bodapati (changes requested):
minor comments

CMakeLists.txt
altmannmarcelo: We do. Added it to CONFIGURE_COMPONENTS()

sql/sql_component.cc
altmannmarcelo: @satya-bodapati we don't have the ability to load multiple components. Only a single keyring will be loaded at any given time.

sql/sql_component.cc
altmannmarcelo: it will make the make_urns return true, however, we call this from both keyring_init(online|offline) we won't reach the point to call this function if component is empty.

storage/innobase/xtrabackup/src/keyring_plugins.cc
altmannmarcelo: yes, service handles will always be initialized

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: No needed. I reverted this change. It was required to write the xtrabackup_component_keyring_file.cnf at the start of the backup, but I have deferred the write operation to backup_finish function.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Fixed

components/keyrings/keyring_file/keyring_file.cc
altmannmarcelo: @rahulmalik87 fixed the 80 characters and MacOS compilation issue. Added '-undefined dynamic_lookup' in case of MacOS to tell the linker that dynamic symbols will be looked at runtime instead of compilation time.

altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/39/ - test results from PXB 8.0.25 against MS 8.0.25 and PS 8.0.23

sql/sql_component.cc
satya-bodapati: Do we ever start PXB with empty component string? If not, can you please add debug assert?

satya-bodapati (changes requested):
Almost there..  close to approval.

Is this Jenkins run the latest? I am interested specifically in keyring tests.

altmannmarcelo:
@satya-bodapati 
> Is this Jenkins run the latest? I am interested specifically in keyring tests.

Yes, this is the latest run. No new force-pushes after the Jenkins run (so far :) )

satya-bodapati:
> @satya-bodapati
> 
> > Is this Jenkins run the latest? I am interested specifically in keyring tests.
> 
> Yes, this is the latest run. No new force-pushes after the Jenkins run (so far :) )

Can you please check if the 'leftover mysqlds' is a test issue? or source issue?
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-pipeline/1247/consoleFull

storage/innobase/xtrabackup/test/t/innodb_keyring_file.sh
```diff
@@ -2,67 +2,80 @@
...
-
-XB_EXTRA_MY_CNF_OPTS="${XB_EXTRA_MY_CNF_OPTS:-""}
-xtrabackup-plugin-dir=${plugin_dir}
-"
-prepare_options="--xtrabackup-plugin-dir=${plugin_dir} ${keyring_args}"
-start_server
-run_cmd $MYSQL $MYSQL_ARGS test -e "SELECT @@server_uuid"
-run_cmd $MYSQL $MYSQL_ARGS test -e "CREATE TABLE tb1 (ID INT PRIMARY KEY) ENCRYPTION='Y'"
-for i in $(seq 1 10);
+for KEYRING_TYPE in "${KEYRING_TEST_TYPES[@]}";
```
satya-bodapati: This loop will not setup for component keyring ? only plugin?

satya-bodapati (changes requested):

altmannmarcelo:
@satya-bodapati 
> Can you please check if the 'leftover mysqlds' is a test issue? or source issue?

This is also present on trunk - https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-pipeline/1179/console
The reason is we currently don't enforce in reviews that all test cases should end with a clean-up stopping running server. As part of clean up of run.sh it stop all running servers

satya-bodapati:
> @satya-bodapati
> 
> > Can you please check if the 'leftover mysqlds' is a test issue? or source issue?
> 
> This is also present on trunk - https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-pipeline/1179/console
> The reason is we currently don't enforce in reviews that all test cases should end with a clean-up stopping running server. As part of clean up of run.sh it stop all running servers

Ok, then we can do it as post task. I think not enforcing proper shutdown can affect some tests when run in parallel.

storage/innobase/xtrabackup/test/t/innodb_keyring_file.sh
altmannmarcelo: @satya-bodapati it will set `KEYRING_TYPE` , then on keyring_common.sh -> do_test we will react accordning to the keyring type.

satya-bodapati: sorry, I just saw the top "KEYRING_TEST_TYPES=("plugin") only :smile: 
I see multiple types for version > 8023 ? but 8023? shouldn't this be 8024?

storage/innobase/xtrabackup/test/t/bug1630841.sh
```diff
@@ -98,6 +102,10 @@ start_server
 mysql -e "CREATE USER pbx@'localhost' IDENTIFIED BY 'password1'"
 mysql -e "GRANT BACKUP_ADMIN, PROCESS, RELOAD, LOCK TABLES, REPLICATION CLIENT ON *.* TO pbx@'localhost'"
 mysql -e "GRANT SELECT ON performance_schema.log_status TO pbx@'localhost'"
+if is_server_version_higher_than 8.0.23
```
satya-bodapati: here too, 8024?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/bug1630841.sh
altmannmarcelo: @satya-bodapati  the higher than doesn't include equal :D got tricked on that one too - https://github.com/percona/percona-xtrabackup/blob/percona-xtrabackup-8.0.23-16/storage/innobase/xtrabackup/test/inc/common.sh#L783

satya-bodapati: brain freeze moment :flushed:

storage/innobase/xtrabackup/test/t/innodb_keyring_file.sh
satya-bodapati: brain freeze moment :flushed:

satya-bodapati (approved):
LGTM!

storage/innobase/xtrabackup/test/t/innodb_keyring_plugin_to_component.sh
```diff
@@ -0,0 +1,127 @@
...
+function run_insert() {
+  trap "finish=true" SIGUSR1
+  finish="false"
+  while [ $finish != "true" ] ; do
+    run_cmd $MYSQL $MYSQL_ARGS test -e "INSERT INTO tb1 VALUES (NULL)"
+    sleep 0.1
+  done
+}
+
+if is_server_version_higher_than 8.0.23
```
rahulmalik87: can we add skip test if server version is not higher than 8.0.23?

storage/innobase/xtrabackup/test/run.sh
rahulmalik87: It is saving mysqld.my file in 8.0.23 server directory. not a big deal but better to to put condition if server_version greater 8.0.23.

altmannmarcelo: Done

rahulmalik87 (approved):
LGTM.

PR #1088, opened 2021-05-17
Fixed PXB-2496 - backup prepare always uses the daemon_keyring_proxy_…

altmannmarcelo:
…plugin without encryption

https://jira.percona.com/browse/PXB-2496

Problem:
As part of the merge of 8.0.24, PXB was always starting
daemon_keyring_proxy_plugin.

Fix:
Only start daemon_keyring_proxy_plugin in case any keyring plugin will
be loaded.

### DIFF
--- a/storage/innobase/xtrabackup/src/keyring_plugins.cc
+++ b/storage/innobase/xtrabackup/src/keyring_plugins.cc
@@ -65,6 +65,34 @@ constexpr size_t BINLOG_KEY_MAGIC_SIZE = sizeof("MYSQL-BINARY-LOG");
 
 static MEM_ROOT argv_alloc{PSI_NOT_INSTRUMENTED, 512};
 
+/* adjust mysql_mandatory_plugins to load keyring proxy only */
+static bool set_keyring_proxy_plugin() {
+  I_List_iterator<i_string> iter(*opt_plugin_load_list_ptr);
+  i_string *item;
+  while (nullptr != (item = iter++)) {
+    const size_t prefix_len = strlen("keyring_");
+    const char *plugin_name = item->ptr;
+    if (strlen(plugin_name) > prefix_len &&
+        strncmp(plugin_name, "keyring_", prefix_len) == 0) {
+      /* initialize only daemon_keyring_proxy_plugin */
+      set_srv_keyring_implementation_as_default();
+      for (st_mysql_plugin **plugin = mysql_mandatory_plugins; *plugin;
+           plugin++) {
+        if (strcmp((*plugin)->name, "daemon_keyring_proxy_plugin") == 0) {
+          mysql_mandatory_plugins[0] = *plugin;
+          mysql_mandatory_plugins[1] = 0;
+          return true;
+        }
+      }
+    }
+  }
+
+  if (strcmp(mysql_mandatory_plugins[0]->name, "daemon_keyring_proxy_plugin") !=
+      0)
+    mysql_mandatory_plugins[0] = 0;
+  return false;
+}
+
 /** Load plugins and keep argc and argv untouched */
 static void init_plugins(int argc, char **argv) {
   int t_argc = argc;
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/41/

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -75,6 +75,7 @@ static void init_plugins(int argc, char **argv) {
 
   mysql_optional_plugins[0] = 0;
 
+  if (!opt_plugin_load_list_ptr->is_empty()) {
```
rahulmalik87: instead of checking not empty. can we check if it contains keyring_file.so to be loaded?

altmannmarcelo: @rahulmalik87 In this case we should check for all possible plugins. At the moment only keyring_file.so and keyring_vault. However, we enforce that at load time (we only load keyring%). Do you think it's necessary to repeat the work here?

satya-bodapati: @altmannmarcelo  If we ever allow a new plugin to be loaded in pxb, we will definitely not know to come and fix this code here. Better to use std::string find with 'keyring'? shouldn't be much overhead

altmannmarcelo: @satya-bodapati @rahulmalik87 done.

rahulmalik87 (approved):
LGTM, Could you paste latest jenkin test result.

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -74,17 +74,28 @@ static void init_plugins(int argc, char **argv) {
...
-    set_srv_keyring_implementation_as_default();
-    for (st_mysql_plugin **plugin = mysql_mandatory_plugins; *plugin;
-         plugin++) {
-      if (strcmp((*plugin)->name, "daemon_keyring_proxy_plugin") == 0) {
-        mysql_mandatory_plugins[0] = *plugin;
-        mysql_mandatory_plugins[1] = 0;
-        break;
+  I_List_iterator<i_string> iter(*opt_plugin_load_list_ptr);
+  i_string *item;
+  while (nullptr != (item = iter++)) {
```
satya-bodapati, line 72: better to avoid assignment in the condition

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -74,17 +74,28 @@ static void init_plugins(int argc, char **argv) {
...
+    const size_t prefix_len = sizeof("keyring_") - 1;
```
satya-bodapati: = strlen("keyring_");

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -74,17 +74,28 @@ static void init_plugins(int argc, char **argv) {
...
+    const char *plugin_name = item->ptr;
+    if (strlen(plugin_name) > prefix_len &&
+        strncmp(plugin_name, "keyring_", prefix_len) == 0) {
+      /* initialize only daemon_keyring_proxy_plugin */
+      set_srv_keyring_implementation_as_default();
+      for (st_mysql_plugin **plugin = mysql_mandatory_plugins; *plugin;
+           plugin++) {
+        if (strcmp((*plugin)->name, "daemon_keyring_proxy_plugin") == 0) {
+          mysql_mandatory_plugins[0] = *plugin;
+          mysql_mandatory_plugins[1] = 0;
```
satya-bodapati, line 83: will this be initialized multiple times? if both keyring_file/keyring_component is passed?
Maybe passing both is invalid? :thinking:

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/keyring_plugins.cc
```diff
@@ -65,6 +65,34 @@ constexpr size_t BINLOG_KEY_MAGIC_SIZE = sizeof("MYSQL-BINARY-LOG");
 
 static MEM_ROOT argv_alloc{PSI_NOT_INSTRUMENTED, 512};
 
+/* adjust mysql_mandatory_plugins to load keyring proxy only */
+static bool set_keyring_proxy_plugin() {
+  I_List_iterator<i_string> iter(*opt_plugin_load_list_ptr);
+  i_string *item;
+  while (nullptr != (item = iter++)) {
+    const size_t prefix_len = strlen("keyring_") - 1;
```
satya-bodapati: I think you should remove -1 now, previously it was sizeof() -1.

satya-bodapati (changes requested):

altmannmarcelo:
@rahulmalik87 https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/44/

satya-bodapati (approved):

PR #1090, opened 2021-05-21
Fixed PXB-2504 - Stats functionality does not work if keyring compone…

altmannmarcelo:
…nt is needed on a mysql datadir

https://jira.percona.com/browse/PXB-2504

Problem:
Components expect manifest file to be placed at the same folder of
running binary to decide which components to load, with that list of
components, it will go to the plugin dir, read the corresponding .so
and a .cnf file with the same name of component.
The component implementation on xtrabackup was done in two parts:
  * Online - used at backup - We query PS.keyring_component_status,
load the component and config from there and save this config on
backup --target-dir as xtrabackup_keyring_component_file.cnf
  * Offline - used at everything else (prepare, copy/move back, stats) -
we read xtrabackup_keyring_component_file.cnf to load the component,
if --keyring-file-data is passed, we read the keyring master keys from
the file passed to it instead of what is saved at
xtrabackup_keyring_component_file.cnf.

The problem with status is that if executing against a datadir instead
of a backup dir. xtrabackup_keyring_component_file.cnf will not be
present. Here the issue is where to load the manifest file to decide if
encryption has to be enabled and from where to read the .cnf file from.

Fix:
Added a new variable named --component-keyring-file-config.
This variable is used on offline operations and will take precedence
over xtrabackup_keyring_component_file.cnf file if it exists.

### DIFF
--- a/storage/innobase/xtrabackup/src/keyring_components.cc
+++ b/storage/innobase/xtrabackup/src/keyring_components.cc
@@ -160,7 +160,9 @@ bool keyring_init_online(MYSQL *connection) {
 
 bool keyring_init_offline() {
   char fname[FN_REFLEN];
-  if (xtrabackup_stats) {
+  if (opt_component_keyring_file_config != nullptr) {
+    strncpy(fname, opt_component_keyring_file_config, FN_REFLEN);
+  } else if (xtrabackup_stats) {
     if (fn_format(fname, XTRABACKUP_KEYRING_FILE_CONFIG, mysql_real_data_home,
                   "", MY_UNPACK_FILENAME | MY_SAFE_PATH) == NULL) {
       return (false);
@@ -185,12 +187,32 @@ bool keyring_init_offline() {
   std::string config;
   File_reader component_config(fname, false, config);
   std::string component_name = "file://component_keyring_file";
-  if (!component_config.valid()) return true;
-  if (config.length() == 0) return false;
+  if (!component_config.valid()) {
+    if (opt_component_keyring_file_config != nullptr) {
+      msg("xtrabackup: Error: Component configuration file is not readable or "
+          "not found.\n");
+      return false;
+    }
+    /* XTRABACKUP_KEYRING_FILE_CONFIG not found. Attempt to init plugin. */
+    return true;
+  }
+  if (config.length() == 0) {
+    msg("xtrabackup: Error: Component configuration file is empty.\n");
+    return false;
+  }
 
   rapidjson::Document config_json;
   config_json.Parse(config);
-  if (!config_json.HasMember("path")) return false;
+  if (config_json.HasParseError()) {
+    msg("xtrabackup: Error: Component configuration file is not a valid "
+        "JSON.\n");
+    return false;
+  }
+  if (!config_json.HasMember("path")) {
+    msg("xtrabackup: Error: Component configuration does not have path "
+        "member.\n");
+    return false;
+  }
 
   std::string path;
   if (opt_keyring_file_data != nullptr) {
--- a/storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
+++ b/storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
@@ -1,0 +1,87 @@
+#
+# Component keyring file specific variables
+#
+require_server_version_higher_than 8.0.23
+
+KEYRING_TYPE="component"
+. inc/keyring_common.sh
+. inc/keyring_file.sh
+configure_server_with_component
+
+run_cmd $MYSQL $MYSQL_ARGS test <<EOF
+CREATE TABLE t1 (c1 VARCHAR(100)) ENCRYPTION='y';
+INSERT INTO t1 (c1) VALUES ('ONE'), ('TWO'), ('THREE');
+INSERT INTO t1 (c1) VALUES ('10'), ('20'), ('30');
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+ALTER INSTANCE ROTATE INNODB MASTER KEY;
+CREATE TABLE t2 (c1 VARCHAR(100)) ENCRYPTION='y';
+INSERT INTO t2 SELECT * FROM t1;
+EOF
+
+xtrabackup --backup --target-dir=$topdir/backup1 \
+--component-keyring-file-config=${MYSQLD_DATADIR}/component_keyring_file.cnf 2>&1 | tee $topdir/pxb.log
+grep_count "xtrabackup: Warning: \-\-component-keyring-file-config will be ignored for \-\-backup operation" $topdir/pxb.log 1
+
+cp -r $topdir/backup1 $topdir/backup2
+record_db_state test
+shutdown_server
+
+xtrabackup --stats --datadir=${mysql_datadir} \
+  --component-keyring-file-config=${MYSQLD_DATADIR}/component_keyring_file.cnf \
+  --xtrabackup-plugin-dir=${plugin_dir} 2>&1 | tee $topdir/pxb.log
+
+grep_count "Encryption can't find master key, please check the keyring is loaded." $topdir/pxb.log 0
+
+# Test 1: Pass non existing configuration file:
+run_cmd_expect_failure $XB_BIN $XB_ARGS \
+--prepare --target-dir=$topdir/backup1 --xtrabackup-plugin-dir=${plugin_dir} \
+--component-keyring-file-config=$topdir/xtrabackup_component_keyring_file_invalid.cnf 2>&1 | tee $topdir/pxb.log
+
+grep_count "Component configuration file is not readable or not found" $topdir/pxb.log 1
+
+# Test 2: Pass non existing configuration file:
+touch $topdir/xtrabackup_component_keyring_file_invalid.cnf
+run_cmd_expect_failure $XB_BIN $XB_ARGS \
+--prepare --target-dir=$topdir/backup1 --xtrabackup-plugin-dir=${plugin_dir} \
+--component-keyring-file-config=$topdir/xtrabackup_component_keyring_file_invalid.cnf 2>&1 | tee $topdir/pxb.log
+
+grep_count "Component configuration file is empty" $topdir/pxb.log 1
+
+# Test 3: Try to prepare with rubbish on component config on target-dir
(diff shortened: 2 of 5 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/46/

storage/innobase/xtrabackup/src/keyring_components.cc
```diff
@@ -160,7 +160,9 @@ bool keyring_init_online(MYSQL *connection) {
 
 bool keyring_init_offline() {
   char fname[FN_REFLEN];
-  if (xtrabackup_stats) {
```
rahulmalik87, line 163: do we throw any error if component configration file is not used?

altmannmarcelo: @rahulmalik87 If you have not passed a configuration file and there is none in the --target-dir --datadir (in case of status) we will attempt to load the plugin. Then here it follows the current behavior. It won't be able to find the master key when trying to decrypt the table.
If you pass a component config and it has any error in it, then pxb understands that you want to use the component but the config is wrong. In this case, we abort right away.

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
```diff
@@ -0,0 +1,61 @@
...
+xtrabackup --backup --target-dir=$topdir/backup1
+cp -r $topdir/backup1 $topdir/backup2
+record_db_state test
+shutdown_server
+
+xtrabackup --stats --datadir=${mysql_datadir} \
+  --component-keyring-file-config=${MYSQLD_DATADIR}/component_keyring_file.cnf \
+  --xtrabackup-plugin-dir=${plugin_dir}
+
+if grep -q "Encryption can't find master key, please check the keyring is loaded." $OUTFILE
```
satya-bodapati: OUTFILE is set in common header?

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "bla" > $topdir/backup1/xtrabackup_component_keyring_file.cnf
+run_cmd_expect_failure $XB_BIN $XB_ARGS \
+--prepare --target-dir=$topdir/backup1 --xtrabackup-plugin-dir=${plugin_dir}
+
+# Test 2: Try to prepare with good component config on target-dir but
+# pass rubish as parameter
+run_cmd_expect_failure $XB_BIN $XB_ARGS \
+--prepare --target-dir=$topdir/backup2 \
+--component-keyring-file-config=$topdir/backup1/xtrabackup_component_keyring_file.cnf \
+--xtrabackup-plugin-dir=${plugin_dir}
```
satya-bodapati: is there in difference in ${plugin_dir} vs $plugin_dir?

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
```diff
@@ -0,0 +1,61 @@
...
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+ALTER INSTANCE ROTATE INNODB MASTER KEY;
+CREATE TABLE t2 (c1 VARCHAR(100)) ENCRYPTION='y';
+INSERT INTO t2 SELECT * FROM t1;
+EOF
+
+xtrabackup --backup --target-dir=$topdir/backup1
```
satya-bodapati: Please add a test scenarios with --component-ekyring-file-config passed to backup
1. passing a valid file
2. passing empty without value
3. passing rubbish file

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
```diff
@@ -0,0 +1,61 @@
...
+xtrabackup --stats --datadir=${mysql_datadir} \
+  --component-keyring-file-config=${MYSQLD_DATADIR}/component_keyring_file.cnf \
+  --xtrabackup-plugin-dir=${plugin_dir}
+
+if grep -q "Encryption can't find master key, please check the keyring is loaded." $OUTFILE
+then
+  die "Cannot read component keyring file config to run stats on live server"
+fi
+
+# Test 1: Try to prepare with rubish on component config on target-dir
```
satya-bodapati: s/rubish/rubbish/

all occurences

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
satya-bodapati: Can you paste the output of PXB command? Would like to see the error msg printed

satya-bodapati (changes requested):
minor comments..

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
altmannmarcelo: @satya-bodapati for backup this is not an option. Backup uses the online function and whatever is loaded into the server is the source of truth.

satya-bodapati: Right, u could say negative testing :wink:

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
altmannmarcelo:
```
2021-05-21 09:18:47: run.sh: ===> /work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup --defaults-file=/work/pxb/ins/8.0/xtrabackup-test/test/var/w1/var1/my.cnf --no-version-check --prepare --target-dir=/work/pxb/ins/8.0/xtrabackup-test/test/var/w1/var1/backup1 --xtrabackup-plugin-dir=/work/pxb/ins/8.0/xtrabackup-test/test/../../lib/plugin/
xtrabackup: recognized server arguments: --innodb_checksum_algorithm=crc32 --innodb_log_checksums=1 --innodb_data_file_path=ibdata1:12M:autoextend --innodb_log_files_in_group=2 --innodb_log_file_size=50331648 --innodb_page_size=16384 --innodb_undo_directory=./ --innodb_undo_tablespaces=2 --server-id=1 --innodb_log_checksums=ON --innodb_redo_log_encrypt=1 --innodb_undo_log_encrypt=1
xtrabackup: recognized client arguments: --no-version-check=1 --prepare=1 --target-dir=/work/pxb/ins/8.0/xtrabackup-test/test/var/w1/var1/backup1 --xtrabackup-plugin-dir=/work/pxb/ins/8.0/xtrabackup-test/test/../../lib/plugin/
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup version 8.0.25-17 based on MySQL server 8.0.25 Linux (x86_64) (revision id: bc921ae98cf)
xtrabackup: cd to /work/pxb/ins/8.0/xtrabackup-test/test/var/w1/var1/backup1/
xtrabackup: This target seems to be not prepared yet.
Number of pools: 1
xtrabackup: xtrabackup_logfile detected: size=8388608, start_lsn=(18143441)
xtrabackup: using the following InnoDB configuration for recovery:
xtrabackup:   innodb_data_home_dir = .
xtrabackup:   innodb_data_file_path = ibdata1:12M:autoextend
xtrabackup:   innodb_log_group_home_dir = .
xtrabackup:   innodb_log_files_in_group = 1
xtrabackup:   innodb_log_file_size = 8388608
xtrabackup: Error: failed to init keyring component
```

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
altmannmarcelo: @satya-bodapati 

```
marcelo@marce-bld:/work/pxb/ins/8.0/xtrabackup-test/test$ ./run.sh -t t/innodb_keyring_file_component.sh
Using /work/pxb/ins/8.0/xtrabackup-test/test/var as test root
Detecting server version...
Running against MySQL 8.0.25 (InnoDB 8.0.25)
Using 16 parallel workers

innodb_keyring_file_component            w1	[passed]    18
Spent 18 of 25 seconds executing testcases

marcelo@marce-bld:/work/pxb/ins/8.0/xtrabackup-test/test$ grep "\-\-component-keyring-file-config will be ignored for \-\-backup operation" results/innodb_keyring_file_component
xtrabackup: Warning: --component-keyring-file-config will be ignored for --backup operation
marcelo@marce-bld:/work/pxb/ins/8.0/xtrabackup-test/test$
```

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
satya-bodapati: Possible to print that keyring component file passed is invalid?

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
satya-bodapati: Thanks, I will check the updated test

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
altmannmarcelo: @satya-bodapati no, they are the same. Just in case you have like:
var=foo
and want to print $var + the string bar you cannot do $varbar you will have to do ${var}bar

storage/innobase/xtrabackup/test/t/innodb_keyring_file_component.sh
altmannmarcelo: Yes, adjusted this a bit. I always miss a function to pass a pattern, file and occurrences I expect and that will exit in case the number of occurrences of patter on file doesn't match. Adjusted xtrabackup commands inside this test case to tee the output to a specific file and used that new file (which gets overwritten every time) as the source of grep file.

altmannmarcelo:
@satya-bodapati @rahulmalik87 test ran for `-t t/innodb_keyring_file_component.sh -t t/innodb_keyring_file.sh` after the new messages. The only one affected is `innodb_keyring_file_component.sh` but also ran `innodb_keyring_file.sh` to be sure. 

https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/47/

rahulmalik87 (approved):
LGTM

satya-bodapati (approved):
Thanks for improving the testcase

PR #1094, opened 2021-05-27
PXB-2106 : copy-back creates wrong binlog.index

kennt-percona:
PXB-2457 : Incorrect binlog names if binlog name contains periods

Issue
When performing copy-back, PXB will add incorrect file names to the binlog index.
This occurs when:
(1) an absolute path is used for the binlog and
(2) when the path is the datadir path.

Also, if the binlog name contains a period ('.'), we were not building
the binlog file names the same way as MySQL.

Solution
This occurs because the system treats the path as an absolute path
(because it is specified as an absolute path) and it treats it as
a relative path (because the path is in the datadir).
This exposed a bug in how we were parsing the paths.

Binlog names with a period ('.') are not generated the same way as MySQL.
Everything after the first '.' is removed.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1564,7 +1564,7 @@ bool copy_if_ext_matches(const char **ext_list, const datadir_entry_t &entry,
 }
 
 struct binlog_file_location {
-  /* binlog file path */
+  /* binlog file path (full path including filename) */
   std::string path;
 
   /* binlog file name */
@@ -1602,19 +1602,21 @@ struct binlog_file_location {
     binlog_file_location r;
 
     if (!name.empty()) {
-      std::string suffix = fn_ext(name.c_str());
-
       if (opt_log_bin != nullptr) {
+        std::string suffix = fn_ext(name.c_str());
+        r.name = std::string(opt_log_bin).substr(dirname_length(opt_log_bin));
+
+        /* Truncate at the first '.' like MySQL */
+        r.name = r.name.substr(0, r.name.find(FN_EXTCHAR)) + suffix;
+
         if (Fil_path::is_absolute_path(opt_log_bin)) {
-          r.name =
+          r.path =
               std::string(opt_log_bin).substr(0, dirname_length(opt_log_bin)) +
-              suffix;
-          r.path = opt_log_bin + suffix;
+              r.name;
         } else {
           char buf[FN_REFLEN];
           fn_format(buf, opt_log_bin, datadir.c_str(), "", MY_UNPACK_FILENAME);
-          r.name = opt_log_bin + suffix;
-          r.path = buf + suffix;
+          r.path = std::string(buf).substr(0, dirname_length(buf)) + r.name;
         }
       } else {
         char buf[FN_REFLEN];
--- a/storage/innobase/xtrabackup/test/t/binlog.sh
+++ b/storage/innobase/xtrabackup/test/t/binlog.sh
@@ -1,3 +1,4 @@
+. inc/common.sh
 
 function backup() {
   mkdir $topdir/binlog-dir1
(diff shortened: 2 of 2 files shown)


storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1589,15 +1589,24 @@ struct binlog_file_location {
 
       if (opt_log_bin != nullptr) {
         if (Fil_path::is_absolute_path(opt_log_bin)) {
-          r.name =
+          r.name = std::string(opt_log_bin).substr(dirname_length(opt_log_bin));
+
+          /* Truncate at the first '.' like MySQL */
+          r.name = r.name.substr(0, r.name.find(FN_EXTCHAR)) + suffix;
```
altmannmarcelo: This same code is used in the else branch. Can we move `r.name` to before  `if (Fil_path::is_absolute_path(opt_log_bin)) {` and inside the if/else compute only `r.path`

altmannmarcelo: Move the declaration of `suffix` to this part of the code too.

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1589,15 +1589,24 @@ struct binlog_file_location {
...
           fn_format(buf, opt_log_bin, datadir.c_str(), "", MY_UNPACK_FILENAME);
-          r.name = opt_log_bin + suffix;
-          r.path = buf + suffix;
+
+          r.name = std::string(opt_log_bin).substr(dirname_length(opt_log_bin));
+
+          /* Truncate at the first '.' like MySQL */
+          r.name = r.name.substr(0, r.name.find(FN_EXTCHAR)) + suffix;
+
+          r.path = std::string(buf).substr(0, dirname_length(buf)) + r.name;
```
altmannmarcelo, line 1619: Can you please let me know why this part is needed?
If understand correctly, the issue only happens if you have a full path that points to datadir. This else branch falls into the not absolute path.

altmannmarcelo (changes requested):
A few inline comments plus:

* Is this 8.0 only? If not, please start with 2.4 path
* Can you please adjust the commit message to include the link to jira ticket, like https://github.com/percona/percona-xtrabackup/commit/80c1ba62aa65d9b0583ca2ad0eb21616b45264a2
* Can you please run jenkins against it. For PXB we use https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/ to trigger a compile + test tasks. After ready, please send us the link for your test param (https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/)

storage/innobase/xtrabackup/src/backup_copy.cc
kennt-percona: This is the other bug that was being fixed.  PXB-2457.

storage/innobase/xtrabackup/src/backup_copy.cc
kennt-percona: code moved

kennt-percona:
@altmannmarcelo 
This is 8.0 only.

altmannmarcelo (approved):
@kennt-percona Thanks for fixing my comments. LGTM

Techshrimp:
This is due to the fact that the binlog name is different before and after the backup and causes problems during parsing, which can be replicated normally, and is related to the target_location function.

from percona-xtrabackup-8.0.6 to percona-xtrabackup-8.0.23-16

binlog_file_location target_location(const std::string &datadir) const {
    binlog_file_location r;

    if (!name.empty()) {
      std::string suffix = fn_ext(name.c_str());

      if (opt_log_bin != nullptr) {
        if (Fil_path::is_absolute_path(opt_log_bin)) {
          r.name =
              std::string(opt_log_bin).substr(0, dirname_length(opt_log_bin)) +
              suffix;
          r.path = opt_log_bin + suffix;
        } else {
          char buf[FN_REFLEN];
          fn_format(buf, opt_log_bin, datadir.c_str(), "", MY_UNPACK_FILENAME);
          r.name = opt_log_bin + suffix;
          r.path = buf + suffix;
        }

PR #1099, opened 2021-05-28
Fixed PXB-2486 - PXB doesn't handle broken pipe correctly when --encr…

altmannmarcelo:
…ypt and --parallel is used

https://jira.percona.com/browse/PXB-2486

Problem:

Inside compress_write and encrypt_write we take thd->ctrl_mutex to
validate only one thread is writting at the time. In case of failure at
xb_crypt_write_chunk / ds_write / write_uint64_le or write_uint32_le we
will terminate the threads without releasing the mutex the thread owns.
In case other thread is waiting on this mutex, it will hang forever.

Fix:

Ensure threads release ctrl_mutex and data_mutex as part of error
handling.
Also added capabilities to detect if we are running on a debug binary.
Note that broken pipe test will not skip in case of non debug. Because
the test for --encrypt can be easily reproducible.

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_compress.c
+++ b/storage/innobase/xtrabackup/src/ds_compress.c
@@ -190,6 +190,7 @@ compress_write(ds_file_t *file, const void *buf, size_t len)
 	uint			i;
 	const char		*ptr;
 	ds_file_t		*dest_file;
+	my_bool simulate_compress_error = FALSE;
 
 	comp_file = (ds_compress_file_t *) file->ptr;
 	comp_ctxt = comp_file->comp_ctxt;
@@ -240,13 +241,14 @@ compress_write(ds_file_t *file, const void *buf, size_t len)
 			}
 
 			xb_a(threads[i].to_len > 0);
+			DBUG_EXECUTE_IF("compress_error", simulate_compress_error = TRUE;);
 
 			if (ds_write(dest_file, "NEWBNEWB", 8) ||
 			    write_uint64_le(dest_file,
-					    comp_file->bytes_processed)) {
+					    comp_file->bytes_processed) || simulate_compress_error ) {
 				msg("compress: write to the destination stream "
 				    "failed.\n");
-				return 1;
+				goto err;
 			}
 
 			comp_file->bytes_processed += threads[i].from_len;
@@ -256,7 +258,7 @@ compress_write(ds_file_t *file, const void *buf, size_t len)
 					   threads[i].to_len)) {
 				msg("compress: write to the destination stream "
 				    "failed.\n");
-				return 1;
+				goto err;
 			}
 
 			pthread_mutex_unlock(&threads[i].data_mutex);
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-test-param/35/

storage/innobase/xtrabackup/src/ds_compress.c
```diff
@@ -190,6 +190,7 @@ compress_write(ds_file_t *file, const void *buf, size_t len)
 	uint			i;
 	const char		*ptr;
 	ds_file_t		*dest_file;
+	my_bool simulate_compress_error = FALSE;
```
satya-bodapati, line 193: doesn't this cause relase build failure? this variable used only on debug builds?

storage/innobase/xtrabackup/test/t/broken_pipe.sh
```diff
@@ -24,3 +24,27 @@ xtrabackup --backup --stream=xbstream \
...
+	--target-dir=$topdir/backup1 --parallel=10 --encrypt=AES256 \
+	--encrypt-key=2222233333232323 | dd bs=1024 count=100 of=/dev/null
+
+if [ ${PIPESTATUS[0]} -eq 124 ] ; then
+	die "xtrabackup timeout after 20 seconds when it was expected to fail"
+fi
+
+if [[ is_debug_pxb_version ]]; then
+	# let the command fail and capture the output
+	set +e
```
satya-bodapati, line 41: whats up with this set +e and -e commands here?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ds_compress.c
altmannmarcelo: @satya-bodapati this is always false on relwithdebinfo. Unless you are in debug with the debug flag set, then this will be changed to true.

storage/innobase/xtrabackup/test/t/broken_pipe.sh
altmannmarcelo: @satya-bodapati  By default, tests run with `set -e` which will fail if the command returns an error. 
At `run_cmd` and `run_cmd_expect_error` we also do change it to `set +e ... set -e`  however, here I want to capture the return code of the command to validate if it failed or timeout out (which gives 124 as return code)

storage/innobase/xtrabackup/src/ds_compress.c
satya-bodapati: on upstream, when you have variables that are used only in DEBUG_EXECUTE_IF() cause release build failures.The variable is declared under  #ifndef DBUG_OFF. Maybe this is OK in PXB

storage/innobase/xtrabackup/test/t/broken_pipe.sh
satya-bodapati: ok, special scenario

storage/innobase/xtrabackup/src/ds_compress.c
altmannmarcelo: the way it's built here it will be used on relwithdebinfo, but it will always be false. Inside DEBUG_EXECUTE_IF we change it to true.

satya-bodapati: ignore, it is used at on more place apart from DBUG_EXECUTE_IF()

satya-bodapati (approved):

PR #1100, opened 2021-05-29
Fixed PXB-2517 - Check if files-from is successfully closed before rsync

garenchan:
PXB-2517 : rsync non-InnoDB tables and files failed: No such file or directory

https://jira.percona.com/browse/PXB-2517

Issue
When hard disks ard under heavy workload, `fclose() ` of `xtrabackup_rsyncfiles_pass` may fail.
And the content of any unwritten output buffer will be lost.
So `rsync` will fail due to incomplete `--files-from` or synchronize incomplete file list.

Solution
Check if files-from is successfully closed before rsync.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1035,6 +1035,7 @@ cleanup:
 
 bool backup_files(const char *from, bool prep_mode) {
   char rsync_tmpfile_name[FN_REFLEN];
+  char errbuf[MYSYS_STRERROR_SIZE];
   FILE *rsync_tmpfile = NULL;
   bool ret = true;
 
@@ -1079,7 +1080,11 @@ bool backup_files(const char *from, bool prep_mode) {
       rsync_list.insert("ib_lru_dump");
     }
 
-    fclose(rsync_tmpfile);
+    if (fclose(rsync_tmpfile) != 0) {
+      msg("Error: can't close file %s: %s\n", rsync_tmpfile_name,
+          my_strerror(errbuf, sizeof(errbuf), my_errno()));
+      return (false);
+    }
     rsync_tmpfile = NULL;
 
     cmd << "rsync -t . --files-from=" << rsync_tmpfile_name << " "


storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1079,7 +1080,11 @@ bool backup_files(const char *from, bool prep_mode) {
       rsync_list.insert("ib_lru_dump");
     }
 
-    fclose(rsync_tmpfile);
+    if(fclose(rsync_tmpfile) != 0) {
+        msg("Error: can't close file %s: %s\n", rsync_tmpfile_name,
+            my_strerror(errbuf, sizeof(errbuf), my_errno()));
+        return (false);
+    }
```
altmannmarcelo:
```suggestion
    if (fclose(rsync_tmpfile) != 0) {
      msg("Error: can't close file %s: %s\n", rsync_tmpfile_name,
          my_strerror(errbuf, sizeof(errbuf), my_errno()));
      return (false);
    }
```

altmannmarcelo (changes requested):
Hi @garenchan. 
Thank you a lot for your contribution.

If I understand correctly, the current behavior means xtrabackup will fail to run rsync.
The proposed solution is to halt the backup if we cannot close the file.
Both will lead to the same end result - a failed backup.

I don't know if it will be possible after the first fclose attempt, but ideally, we should retry to close the file a few more times before giving up.

Would you be willing to write a test case that can explore this issue? We might need a debug binary and use of DBUG_EXECUTE_IF - example 4ffef1477

From PXB-2517 one thing caught my attention:
> rsync: link_stat "/var/lib/mysql/demo/user.f" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1052)

`user.f` is not a valid file extension for mysql. I believe this was intended to be `user.frm` . Do you know if xtrabackup created the file with the wrong name in the file list or that is just a buffer issue from rsync?

storage/innobase/xtrabackup/src/backup_copy.cc
garenchan: I think it is possible that the content of  `xtrabackup_rsyncfiles_pass` has been truncated due to a failure to `fclose` the file.
If it happens to be truncated at the end of a line, then `rsync` will synchronize an incomplete list of files. So I think we should terminate the program as soon as `fclose` fails.

altmannmarcelo:
Hi @garenchan,

I'm still not convinced that just erroring out here is the best scenario. We should always attempt to complete the backup if it's possible.
Do you have some instructions in which we can try to reproduce it? If not I will attempt to reproduce it via forcing fclose system call failure.

On your side, what error do you get about the failed fclose?
Since the main issue here is a truncated file passed to rsync, what about doing a fsync on rsync_tmpfile before attempting to fclose it ?

altmannmarcelo (changes requested):
Hi @garenchan,

I'm still not convinced that just erroring out here is the best scenario. We should always attempt to complete the backup if it's possible.
Do you have some instructions in which we can try to reproduce it? If not I will attempt to reproduce it via forcing fclose system call failure.

On your side, what error do you get about the failed fclose?
Since the main issue here is a truncated file passed to rsync, what about doing a fsync on rsync_tmpfile before attempting to fclose it ?

garenchan:
> Hi @garenchan,
> 
> I'm still not convinced that just erroring out here is the best scenario. We should always attempt to complete the backup if it's possible.
> Do you have some instructions in which we can try to reproduce it? If not I will attempt to reproduce it via forcing fclose system call failure.
> 
> On your side, what error do you get about the failed fclose?
> Since the main issue here is a truncated file passed to rsync, what about doing a fsync on rsync_tmpfile before attempting to fclose it ?

Hi, @altmannmarcelo,

The following code is part of the `fclose` implementation in `glibc-2.17`.  When `fclose` is called, it implicitly performs a `flush` operation. It frees the buffer even if `flush` fails. 

So if we don't check the return value of `fclose`, we might end up with an incomplete `files-from` file. This may cause `rsync` to copy an incomplete list of non-InnoDB tables and files. In the end we will succeed in getting a broken backup. 

In the worst case, we can still use the broken backup to restore data, and it may take a long time for the problem to be discovered. Compared to such a bad situation, I believe it is easier to accept that the backup process fails early. And we usually do backups periodically, it's okay to fail occasionally.

```c
int
_IO_new_file_close_it (fp)
     _IO_FILE *fp;
{
  int write_status;
  if (!_IO_file_is_open (fp))
    return EOF;

  if ((fp->_flags & _IO_NO_WRITES) == 0
      && (fp->_flags & _IO_CURRENTLY_PUTTING) != 0)
    write_status = _IO_do_flush (fp);
  else
    write_status = 0;

  _IO_unsave_markers (fp);

  int close_status = ((fp->_flags2 & _IO_FLAGS2_NOCLOSE) == 0
		      ? _IO_SYSCLOSE (fp) : 0);

  /* Free buffer. */
#if defined _LIBC || defined _GLIBCPP_USE_WCHAR_T
  if (fp->_mode > 0)
    {
      if (_IO_have_wbackup (fp))
	_IO_free_wbackup_area (fp);
      _IO_wsetb (fp, NULL, NULL, 0);
      _IO_wsetg (fp, NULL, NULL, NULL);
      _IO_wsetp (fp, NULL, NULL);
    }
#endif
  _IO_setb (fp, NULL, NULL, 0);
  _IO_setg (fp, NULL, NULL, NULL);
  _IO_setp (fp, NULL, NULL);

  _IO_un_link ((struct _IO_FILE_plus *) fp);
  fp->_flags = _IO_MAGIC|CLOSED_FILEBUF_FLAGS;
  fp->_fileno = -1;
  fp->_offset = _IO_pos_BAD;

  return close_status ? close_status : write_status;
}
libc_hidden_ver (_IO_new_file_close_it, _IO_file_close_it)
```

garenchan:
> `user.f` is not a valid file extension for mysql. I believe this was intended to be `user.frm` . Do you know if xtrabackup created the file with the wrong name in the file list or that is just a buffer issue from rsync?

The following code is part of `rsync-3.0.9`. It reads `files-from` file line by line, and the buffer size `1024` is large enough. So I think this problem is not caused by `rsync`.

https://github.com/WayneD/rsync/blob/40afd365cc8ca968fd16e161d24df5b8a8a520cc/rsync.h#L609-L611

```c
#ifndef MAXPATHLEN
#define MAXPATHLEN 1024
#endif
```

https://github.com/WayneD/rsync/blob/40afd365cc8ca968fd16e161d24df5b8a8a520cc/flist.c#L2030-L2036

```c
	while (1) {
		char fbuf[MAXPATHLEN], *fn, name_type;


		if (use_ff_fd) {
			if (read_line(filesfrom_fd, fbuf, sizeof fbuf, rl_flags) == 0)
				break;
			sanitize_path(fbuf, fbuf, "", 0, SP_KEEP_DOT_DIRS);
```

altmannmarcelo (approved):
Hi @garenchan .
I can reproduce the issue by commenting out the call to `fclose` so I can validate the issue exists. However, as mentioned before, we always try to make the backup complete to provide a better user experience. I have incorporated your patch and also added a call to [fflush](https://cplusplus.com/reference/cstdio/fflush/) in order to flush any remaining buffers before attempting to close the file. 

Thanks for your contribution.

PR #1117, opened 2021-06-29
8.0 PXB-2477 (xbcloud should retry on error and utilize incremental backoff)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2477

Problem
Currently we only support retries on s3 api for upload/put action.
The current implementation retries 3 times without any delay in between
attemps. This is goes against network transmission best practices.

Solution

Implemented retry callback for swift (upload/download) and
S3 (download) - upload was already implemented.

Implement truncated exponential backoff to avoid network congestion.
This alows users to choose how many times they will want a chunk to be
retransmitted before erroring out. Also, users will have the ability to
choose a maximum delay in between retries, in which the exponential
time delay will stop increasing while still retrying.

Added an option to extend the fixed list of curl errors that will be
considered as retriable. This can be used with --verbose flag to display
which error curl has returned and which parameter value has to be added
in order to allow that extra error(s).

Fixed some swift error messages that were missing \n.

Fixed Http_client::make_async_request to report proper action, as it
had "upload object" action hardcoded.

Thanks to @timvaillancourt for the initial contribution on this subject.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/http.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/http.cc
@@ -1,5 +1,5 @@
 /******************************************************
-Copyright (c) 2019 Percona LLC and/or its affiliates.
+Copyright (c) 2019, 2021 Percona LLC and/or its affiliates.
 
 HTTP client implementation using cURL.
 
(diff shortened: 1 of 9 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/56/

satya-bodapati (approved):

PR #1120, opened 2021-07-06
PXB-2101 prepare fails due to duplicate sdi

rahulmalik87:
https://jira.percona.com/browse/PXB-2101

Problem:
Prepare can fail when tables exist with duplicate SDI.

Analysis:
MySQL tables can have duplicate SDI. This could be
because of import tablespace, Exchange Partition.
PXB process both SDI and creates different dict_table_t
objects and one of the objects can belong to another table
involved in exchange partition

Fix:
If a table has more than one SDI key. Query the right
key from the mysql/tables and only process that key.

MySQL 8.0.24 has fixed the duplicated SDI issue
but tables upgraded from lower version can still have duplicate
SDI


storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1309,6 +1309,34 @@ function innodb_checkpoint_lsn()
         grep "Last checkpoint at" | awk '{ print $4 }'
 }
 
+#general method to restore
+########################################################################
+function backup_and_restore() {
```
satya-bodapati, line 1359: why backup_restore is added again? I remember that you or marce added this already

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
```diff
@@ -0,0 +1,44 @@
...
+DROP TABLE t1;
+CREATE TABLE t1(a  INT NOT NULL, PRIMARY KEY(a)) engine = InnoDB;
+ALTER TABLE t1 DISCARD TABLESPACE;
+CREATE DATABASE test2;
+use test2;
+CREATE TABLE t1(a  INT NOT NULL, PRIMARY KEY(a)) engine = InnoDB;
+FLUSH table t1 for EXPORT;
+UNLOCK TABLES;
+EOF
+
```
satya-bodapati: we need a zipped datadir. Old version will have duplicate SDI (after doing exchange, import etc). Please create a backup with such tables.

satya-bodapati: with current version, your patch isn't even executed, isn't it?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,94 @@ static bool innodb_init_param(void) {
...
+    dd_process_schema_rec(heap, rec, sys_tables, &mdl_on_tab, &mtr, &rec_name,
+                          &rec_schema_id);
+    mem_heap_empty(heap);
+
+    if (rec_name.compare(db_name) == 0) {
+      schema_id = rec_schema_id;
+      break;
+    }
+    rec = (rec_t *)dd_getnext_system_rec(&pcur, &mtr);
+  }
```
satya-bodapati: after the first scan, we should create the in-memory map and avoid future scans.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2178,6 +2261,16 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
     goto error;
   }
 
+  /* duplicate SDI can exists in version lower than MySQL 8.0.24 or if database
```
satya-bodapati: duplicate SDI can exist in file_per_table IBD files lower MySQL version 8.0.24 if table is used in EXCHANGE PARTITION or IMPORT. Even after upgrade to latest version 8.0.25 (which fixed the duplicate SDI issue), such
tables continue to contain duplicate SDI. PXB will scan the DD table "mysql.tables" to determine correct SDI

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2212,10 +2314,6 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
                                   compressed_sdi_len);
     decompressor.decompress();
 
-    if (ib_key.sdi_key->type != 1 /* dd::Sdi_type::TABLE */) {
-      continue;
-    }
-
```
satya-bodapati, line 2218: why did you remove this? we are interested only in TABLE type records right?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: for some reason, It was not there. may be some mistake with final merge. Also there was issue in jenkin. It was not reporting the correct failures which Illia has fixed. So it didn't show up in jenkins.

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: that's why I executed it against 8.0.23 in jenkins.  let me create a zipper datadir as well

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: I had this in mind. but given upstream has already fixed duplicate SDI and there won't be lot of duplicate just leftover from upgrade. If you still think it can be useful. I will create the map instead of scanning again and again.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: I moved this little above few lines to avoid some more computation if sdi_key->type is 1

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: atleast for schemas, which are less in number compared to tables, we can do this.


For tables, am worried about memory usage. There can be like 1million entries. So not recommending at this point

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: well, there is no GCA involved here.. are you working on old commit? Please update to latest 8.0 and retry.

"for some reason" has to be found :wink:

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
satya-bodapati: Yes, better to have zipped datadir because, in the future, we won't specifically test with 8.0.23 version. we won't remember.

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: the patch is based on latest commit of 8.0. The backup_and_restore  function is not in 8.0. As I mentioned it was not caught in jenkin because of some issue which Illia has fixed.
anyway, the Jenkin is fixed, now we will come to know if any test is failing

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: added the zipped datadir

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2226,12 +2233,12 @@ void get_id_from_mysql_table_scan(std::string name, uint64 *id, THD *thd) {
...
     }
 
+    mtr_start(&mtr);
     rec = (rec_t *)dd_getnext_system_rec(&pcur, &mtr);
   }
 
-  ut_ad(*id != 0);
+  ut_a(*id != 0);
 
-  mtr_commit(&mtr);
```
satya-bodapati: isn't this mtr_commit required?

satya-bodapati: original dd processing function does mtr commit at end

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2226,12 +2233,12 @@ void get_id_from_mysql_table_scan(std::string name, uint64 *id, THD *thd) {
       break;
```
satya-bodapati: upstream dd processing code, never breaked out of the loop. I guess you need a mtr_start() before break to keep it consistent?

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
```diff
@@ -40,5 +45,18 @@ mysql -e "ALTER TABLE t1 IMPORT TABLESPACE;" test
...
+#cleanup backup and place 8.0.23 backupdir
+rm -rf $topdir/backup && mkdir -p $topdir/backup
+run_cmd tar -xf inc/duplicate_sdi_backup.tar.gz -C $topdir/backup
+#prepare
+run_cmd xtrabackup --prepare --target-dir=$topdir/backup
+#restore
+[[ -d $mysql_datadir ]] && rm -rf $mysql_datadir && mkdir -p $mysql_datadir
+xtrabackup --copy-back --target-dir=$topdir/backup
+
+start_server
```
satya-bodapati, line 78: can you please check the contents of table here (after starting server)? or table checksums?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: in our case, It should always break. so that's why I have removed mtr_commit before dd_table_close. It would be same thing adding mtr_start here and followed by mtr_commit before dd_table_close. 
Since we have assert before dd_table_close(in case it doesn't break)

satya-bodapati: if it doesn't break, we need the mtr_commit() ?

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: Sorry, I didn't understand completely. You will remove this duplicate code?

storage/innobase/dict/dict0dd.cc
```diff
@@ -5550,6 +5550,75 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
...
+
+  field = rec_get_nth_field(
+      rec, offsets, dd_tables->field_number("name") + DD_FIELD_OFFSET, &len);
+  auto name_csr = (const char *)field;
+
+  name->assign(name_csr, name_csr + len);
+  mtr_commit(mtr);
+}
+
+/** Process one mysql.tables record and get the id
```
satya-bodapati: get the DD id and the table name

storage/innobase/dict/dict0dd.cc
```diff
@@ -5550,6 +5550,75 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
...
+@param[in]	heap		Temp memory heap
+@param[in]	rec		mysql.tables record
+@param[in]	dd_tables	dict_table_t obj of dd system table
+@param[in]	mdl		MDL on the table
```
satya-bodapati: mdl parameter can be removed?

storage/innobase/dict/dict0dd.cc
satya-bodapati: and schema_id too

storage/innobase/include/dict0dd.h
```diff
@@ -756,6 +756,32 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
                                  mtr_t *mtr, const char *system_table_name,
                                  dict_table_t **table);
 
+/** Process one mysql.metadata record and get the id
+@param[in]	heap		Temp memory heap
+@param[in]	rec		mysql.tables record
+@param[in]	dd_tables	dict_table_t obj of dd system table
+@param[in]	mdl		MDL on the table
```
satya-bodapati: remove mdl?

storage/innobase/include/dict0dd.h
```diff
@@ -756,6 +756,32 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
                                  mtr_t *mtr, const char *system_table_name,
                                  dict_table_t **table);
 
+/** Process one mysql.metadata record and get the id
```
satya-bodapati: get the DD id of schema?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,103 @@ static bool innodb_init_param(void) {
...
+@param[out]     sdi_id		id of table
+@param[in]	thd 		THD */
+void get_id_from_mysql_table_scan(std::string name, uint64 *id, THD *thd) {
+  std::string db_name = name.substr(0, name.find(OS_PATH_SEPARATOR));
+
+  std::string table_name = name.substr(name.find(OS_PATH_SEPARATOR) + 1);
+  if (table_name.find("#P#") != std::string::npos ||
+      table_name.find("#p#") != std::string::npos) {
+    table_name.resize(table_name.find('#'));
+  }
```
satya-bodapati: we don't need to look into mysql.table_partitions?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,103 @@ static bool innodb_init_param(void) {
...
+    MDL_ticket *mdl_on_tab = nullptr;
+    uint64 rec_schema_id;
+    std::string rec_name;
+    uint64 rec_id;
+
+    dd_process_dd_tables_rec(heap, rec, sys_tables, &mdl_on_tab, &mtr,
+                             &rec_schema_id, &rec_name, &rec_id);
+    mem_heap_empty(heap);
+
+    if (schema_id == rec_schema_id && table_name.compare(rec_name) == 0) {
```
satya-bodapati: @rahulmalik87 Can you also please test with a table name that has non-ascii characters?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: yes.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: no not required. it works fine. we store key_id in mysql.tables only. to be sure,  I tried exchanging p3( since metaata is stored in p1) with t1 table and it works fine.

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: backup_and_restore code is not in 8.0

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: it doesn't work, we need to handle it.  thanks you for pointing

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: ok, thanks for checking

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: ok, got it. Then we will introduce as part of this PR

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: please add such table in zipped backup dir.

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: done.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: added

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2160,16 +2160,14 @@ static bool innodb_init_param(void) {
...
-      table_name.find("#p#") != std::string::npos) {
-    table_name.resize(table_name.find('#'));
-  }
+void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
+                                  THD *thd) {
+  std::string db_name;
+  std::string table_name;
   uint64 schema_id = 0;
 
+  dict_name::get_table(name, db_name, table_name);
```
satya-bodapati, line 2176: Can you please add a comment here?

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1309,32 +1309,50 @@ function innodb_checkpoint_lsn()
         grep "Last checkpoint at" | awk '{ print $4 }'
 }
 
+#method to check count in a database/table
```
satya-bodapati: check row count?

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1309,32 +1309,50 @@ function innodb_checkpoint_lsn()
...
+  local db=$1
   record_db_state $db
   rm -rf $topdir/backup
   mkdir -p $topdir/backup
 
   ## backup and prepare based on if keyring is used
   if [ -z ${plugin_dir+x} ]
   then
-    run_cmd xtrabackup --backup --target-dir=$topdir/backup
+    xtrabackup --backup --target-dir=$topdir/backup
```
satya-bodapati, line 1368: why run_cmd is removed?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: U forgot to address this review comment?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,99 @@ static bool innodb_init_param(void) {
...
+    }
+
+    mtr_start(&mtr);
+    rec = (rec_t *)dd_getnext_system_rec(&pcur, &mtr);
+  }
+
+  ut_a(*id != 0);
+
+  dd_table_close(sys_tables, thd, &mdl, true);
+  mem_heap_empty(heap);
```
satya-bodapati: mtr_commit() missing here. Incase we traverse all DD records and reach here (without the break)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2178,6 +2266,18 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
     goto error;
   }
 
+  /* 8.0.24 if the table is used in EXCHANGE PARTITION or IMPORT. Even after
```
satya-bodapati: Before 8.0.24, if the table is ..

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: I replied in previous comment.

> in our case, It should always break. so that's why I have removed mtr_commit before dd_table_close. It would be same thing adding mtr_start here and followed by mtr_commit before dd_table_close.
Also we have assert before dd_table_close(in case it doesn't break)

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: xtrabackup function internally call run_cmd that's why we don't require explicitly.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: We will assert in that case. It is not expected to happpen.

satya-bodapati: A debug assert is fine. Release asserts should be used judiciously. Infact, I see that you used lot of ut_a().

Here please try to add mtr_commit and return error

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: 'xtrabackup' here is the binary right? not the function? :thinking:

rahulmalik87: It is function defined in commoh.sh

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: Added exit with message in case table not found in mysql/schemata or mysql/tables

storage/innobase/xtrabackup/src/xtrabackup.cc

satya-bodapati:
@rahulmalik87 can you please get the Jenkins link here?

storage/innobase/dict/dict0dd.cc
```diff
@@ -5550,6 +5550,73 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
 */
 static const int DD_FIELD_OFFSET = 2;
 
+/** Process one mysql.metadata record and get the id
+@param[in]	heap		Temp memory heap
+@param[in]	rec		mysql.tables record
+@param[in]	dd_tables	dict_table_t obj of dd system table
+@param[in]	mtr		Mini-transaction
+@param[out]	name 		Name of table database/name
```
satya-bodapati: 1. Name of the schema
2. Use lower case for start of the comments(every where)
3. Ok to use doxygen function comments only at .h file and optional for .cc
4. "Process one mysql.metadata record and extract schema name and id"

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,111 @@ static bool innodb_init_param(void) {
...
+    return mysql_schema;
+  };
+
+  static auto mysql_schema = build_mysql_schema();
+
+  schema_id = mysql_schema[db_name];
+
+  ut_ad(schema_id != 0);
+  if (schema_id == 0) {
+    msg("xtrabackup: can't find %s entry in mysql/schemata for space %s\n",
```
satya-bodapati: for tablespace

satya-bodapati: I know that you added exit() but an mtr_commit() would be good here. its like not closing a file and exiting. it may be ok but better to follow proper cleanups/exits

also, am not a fan of exit() calls. Better to do proper exit of threads, propagation of errors, cleanups, and then return. This you can ignore though [ follow original style ]

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2186,6 +2298,15 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
     uint32_t compressed_sdi_len = compressed_buf_len;
     uint32_t uncompressed_sdi_len = uncompressed_buf_len;
 
+    if (ib_key.sdi_key->type != 1 /* dd::Sdi_type::TABLE */) {
+      continue;
+    }
+
+    /* in case of duplicate sdi, sdi_id is update to correct sdi */
```
```suggestion
    /* In case of duplicate SDIs, sdi_id is the latest id according
    to DD, so we skip other dd::Table SDIs in the IBD file */
```

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2298,8 +2414,13 @@ static void dict_load_from_spaces_sdi() {
     return (DB_SUCCESS);
   });
 
+  dict_load_tables_from_space_id(dict_sys_t::s_space_id, thd, trx);
```
satya-bodapati: Add a comment here on why we need to load DD tables (SDI from mysql.ibd) early?

satya-bodapati (changes requested):
Almost there, also please add jenkins link

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
```diff
@@ -0,0 +1,78 @@
...
+run_cmd check_count test t1 5 
+run_cmd check_count test pt1 4
+
+
+
```
satya-bodapati: remove these extra lines

satya-bodapati (changes requested):

storage/innobase/dict/dict0dd.cc
satya-bodapati: @rahulmalik87 you forgot to address this review comment? or it is not clear?

satya-bodapati (changes requested):
Please address all review comments

storage/innobase/dict/dict0dd.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: mtr_commit is not required here. for first call it is done just above  dd_table_close. let me follow orignal style.

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: added

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
```diff
@@ -0,0 +1,75 @@
...
+mysql test <<EOF
+DROP TABLE t1;
+CREATE TABLE t1(a  INT NOT NULL, PRIMARY KEY(a)) engine = InnoDB;
+ALTER TABLE t1 DISCARD TABLESPACE;
+CREATE DATABASE test2;
+use test2;
+CREATE TABLE t1(a  INT NOT NULL, PRIMARY KEY(a)) engine = InnoDB;
+FLUSH table t1 for EXPORT;
+UNLOCK TABLES;
+EOF
```
satya-bodapati, line 52: can you please one more scenario for a table that has been imported and then exchanged?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2883,7 +3009,6 @@ static bool xtrabackup_copy_datafile(fil_node_t *node, uint thread_n) {
 
   bool is_system = !fsp_is_ibd_tablespace(node->space->id);
 
-
```
satya-bodapati, line 2886: no need to touch this function? (revert new line removal)?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2334,7 +2461,6 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
       to_lsn = incremental_last_lsn < incremental_to_lsn ? incremental_to_lsn
                                                          : incremental_last_lsn;
     }
-
```
satya-bodapati, line 2337: no need to touch this function? (revert new line removal)?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,114 @@ static bool innodb_init_param(void) {
...
+      break;
+    }
+
+    mtr_start(&mtr);
+    rec = (rec_t *)dd_getnext_system_rec(&pcur, &mtr);
+  }
+
+  ut_ad(*id != 0);
+  if (id == 0) {
+    msg("xtrabackup: can't find %s entry in mysql/tables for space %s\n",
```
satya-bodapati: s/space/tablespace

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,114 @@ static bool innodb_init_param(void) {
   return (TRUE);
 }
 
+/** Scan the sdi id from mysql.tables
+@param[in]	name 		Name of table database/name
+@param[out]     sdi_id		id of table
+@param[in]	thd 		THD */
```
satya-bodapati: align these properly

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,114 @@ static bool innodb_init_param(void) {
   return (TRUE);
 }
 
+/** Scan the sdi id from mysql.tables
```
```suggestion
/** Scan the sdi id from DD table "mysql.tables"
```

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2156,114 @@ static bool innodb_init_param(void) {
...
+@param[in]	name 		Name of table database/name
+@param[out]     sdi_id		id of table
+@param[in]	thd 		THD */
+void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
```
satya-bodapati: get_id_from_dd_scan()

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2178,6 +2281,18 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
...
   }
 
+  /* Before 8.0.24 if the table is used in EXCHANGE PARTITION or IMPORT. Even
+  after upgrade to the latest version 8.0.25 (which fixed the duplicate SDI
+  issue), such tables continue to contain duplicate SDI. PXB will scan the DD
+  table "mysql.tables" to determine the correct SDI */
+  if (ib_vector.sdi_vector->m_vec.size() > 2 &&
+      strcmp(space->name, "mysql") != 0 &&
+      fsp_is_file_per_table(space_id, space->flags)) {
+    msg("xtrabackup: duplicate sdi for space %s.\n", space->name);
```
satya-bodapati: Also, suggest user to consider doing OPTIMIZE TABLE on this table
"Please do OPTIMIZE TABLE to rebuild tablespace and the duplicate SDI will be removed"

Can you also please test before adding this comment?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: it was modified automatically by clang-format. Should I remove or let it be?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2178,6 +2281,18 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
...
+
+    get_id_from_mysql_table_scan(space->name, &sdi_id, thd);
```
satya-bodapati: There is a way to avoid this DD scan can multiple times. Idea is something like this
1. save the list of all tables with duplicate SDI (table_names, current sdi_ids, etc)
2. after the normal IBDs are processed, if there are tables with duplicate SDI (saved_list not empty)
    we now start DD scan
3. for every DD record, we match from the saved_list and return correct sdi_ids
4. for the processed duplicate SDI tablespace, we remove the saved_list
5. when either saved_list is empty or when we reach the end of DD we stop

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: if it is by clang-format, let's leave it as is.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: On second thought, this will be helpful only for future full backups. incremental backups are not impacted by duplicate SDI, right?

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: added

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: incremental backup are also impacted by duplicate sdi. once we rebuild them we can get rid of duplicated sdi in next incremental backup.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: building the map of <schema id, name> and key during first scan of duplicate sdi

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2178,9 +2177,10 @@ void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
   MDL_ticket *mdl = nullptr;
   mem_heap_t *heap = mem_heap_create(1000);
 
-  mutex_enter(&(dict_sys->mutex));
 
-  auto build_mysql_schema = [&]() {
+  /* map of schema name and schema id build from scanning mysql/schemata */
+  auto map_schema = [&]() {
```
satya-bodapati: dd_schema_map

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2203,11 +2203,45 @@ void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
     dd_table_close(sys_tables, thd, &mdl, true);
     mem_heap_empty(heap);
 
+    mutex_exit(&(dict_sys->mutex));
     return mysql_schema;
   };
 
-  static auto mysql_schema = build_mysql_schema();
+  /* map of <schema id, name> and key build from scanning mysql/tables */
+  auto map_tables = [&]() {
```
satya-bodapati: dd_tables_map

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2217,43 +2251,15 @@ void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
...
-      *id = rec_id;
-      btr_pcur_close(&pcur);
-      mtr_start(&mtr);
-      break;
-    }
-
-    mtr_start(&mtr);
-    rec = (rec_t *)dd_getnext_system_rec(&pcur, &mtr);
-  }
+  static auto mysql_tables = map_tables();
```
satya-bodapati: when do we free this map?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2192,7 +2192,7 @@ void get_id_from_mysql_table_scan(const std::string name, uint64 *id,
 
       dd_process_schema_rec(heap, rec, sys_tables, &mtr, &rec_name,
                             &rec_schema_id);
-      mysql_schema.insert(std::pair<std::string, int>(rec_name, rec_schema_id));
+      mysql_schema.insert(std::make_pair(rec_name, rec_schema_id));
```
satya-bodapati: when do we free this?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: fixed.

rahulmalik87: we free it inside dict_load_from_spaces_sdi

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2438,6 +2438,9 @@ static void dict_load_from_spaces_sdi() {
     dict_load_tables_from_space_id(space_id, thd, trx);
   }
 
+  dd_schema_map.empty();
+  dd_table_map.empty();
```
satya-bodapati: https://www.cplusplus.com/reference/map/map/empty/

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: my bad :( fixed.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -191,6 +191,12 @@ static regex_list_t regex_exclude_list;
 static hash_table_t *tables_include_hash = NULL;
 static hash_table_t *tables_exclude_hash = NULL;
 
+/* map of schema name and schema id */
+static std::map<std::string, int> dd_schema_map;
+
+/* map of <schema id, name> and key */
+static std::map<std::pair<int, std::string>, int> dd_table_map;
```
satya-bodapati: sometimes the ids are declared as uint64 and sometimes as int.

Better to use uint64 everywhere

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -191,6 +191,12 @@ static regex_list_t regex_exclude_list;
 static hash_table_t *tables_include_hash = NULL;
 static hash_table_t *tables_exclude_hash = NULL;
 
+/* map of schema name and schema id */
+static std::map<std::string, int> dd_schema_map;
+
+/* map of <schema id, name> and key */
```
satya-bodapati: schema id, table_name

satya-bodapati: what is "key" here?

storage/innobase/dict/dict0dd.cc
```diff
@@ -5550,6 +5550,73 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
...
+  mtr_commit(mtr);
+}
+
+/** Process one mysql.tables record and get the dd id, table name, schema name
+@param[in]	heap		temp memory heap
+@param[in]	rec		mysql.tables record
+@param[in]	dd_tables	dict_table_t obj of dd system table
+@param[in]	mtr		mini-transaction
+@param[out]	name 		name of table
+@param[out]	schema_id 	schema id of table
```
satya-bodapati: wrong order

storage/innobase/include/dict0dd.h
```diff
@@ -756,6 +756,29 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
...
+                           dict_table_t *dd_tables, mtr_t *mtr,
+                           std::string *name, uint64 *id);
+
+/** Process one mysql.tables record and get the table id, name and schema_id
+@param[in]	heap		temp memory heap
+@param[in]	rec		mysql.tables record
+@param[in]	dd_tables	dict_table_t obj of dd system table
+@param[in]	mtr		mini-transaction
+@param[out]	name 		name of table
+@param[out]	schema_id 	schema id of table
```
satya-bodapati: wrong order

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2162,114 @@ static bool innodb_init_param(void) {
   return (TRUE);
 }
 
+/** Scan the sdi id from DD table "mysql.tables"
```
satya-bodapati: everywhere (except variable names) use SDI instead of sdi because SDI is an abbreviation for Serialized Dictionary Information

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2162,114 @@ static bool innodb_init_param(void) {
...
+  ut_ad(db_name.compare("mysql") != 0);
+
+  dict_table_t *sys_tables = nullptr;
+  btr_pcur_t pcur;
+  const rec_t *rec = nullptr;
+  mtr_t mtr;
+  MDL_ticket *mdl = nullptr;
+  mem_heap_t *heap = mem_heap_create(1000);
+
+  /* map of schema name and id build from scanning mysql/schemata */
```
satya-bodapati: built from

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2162,114 @@ static bool innodb_init_param(void) {
...
+    }
+
+    mtr_commit(&mtr);
+    dd_table_close(sys_tables, thd, &mdl, true);
+    mem_heap_empty(heap);
+
+    mutex_exit(&(dict_sys->mutex));
+  }
+
+  /* map of <schema id, name> and key build from scanning mysql/tables */
```
satya-bodapati: built

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
```diff
@@ -0,0 +1,80 @@
+
+############################################################################
+# PXB-2101 --prepare crash if exchange partition happened
+############################################################################
+require_server_version_higher_than 8.0.22
```
satya-bodapati, line 5: btw, why do we need server version > 8.0.22 ?

satya-bodapati: also is it possible for a test in PXB (without changing bootstrap or other config files)
1. ask for a specific server version 8.0.22
2. and run the same test on the latest server version

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2156,11 +2162,114 @@ static bool innodb_init_param(void) {
...
+    mutex_exit(&(dict_sys->mutex));
+  }
+
+  schema_id = dd_schema_map[db_name];
+
+  ut_ad(schema_id != 0);
+  if (schema_id == 0) {
+    msg("xtrabackup: can't find %s entry in mysql/schemata for tablespace %s\n",
+        db_name.c_str(), name.c_str());
+    exit(EXIT_FAILURE);
```
satya-bodapati: btw, these exits() are a bad idea. I found a feature request that PXB should log the status of a backup. and in case of failure, they want the reason for failure to be logged.

Can you please try to avoid these exits() and do a graceful exit (return until main)

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Ok, can you adjust the message then?

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/dict/dict0dd.cc
rahulmalik87: fixed

storage/innobase/include/dict0dd.h
rahulmalik87: fixed

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/test/t/xb_duplicate_sdi.sh
rahulmalik87: since we are running prepare against saved backup.

satya-bodapati: Ok

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: fixed. now we see messages like this
xtrabackup: can't find test entry in mysql/schemata for tablespace test/pt1@4f60@597d#p#p1
xtrabackup: innodb_init(): Error occured.
2021-07-29T06:31:17.706469Z 0 [Note] [MY-010733] [Server] Shutting down plugin 'keyring_file'
2021-07-29T06:31:17.706936Z 0 [Note] [MY-010733] [Server] Shutting down plugin 'daemon_keyring_proxy_plugin'

satya-bodapati: :clap:

satya-bodapati (changes requested):
1. Squash all commits
2. Please add Jenkins link

rahulmalik87:
jenkin results https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/71/

satya-bodapati (approved):

PR #1122, opened 2021-07-09
PXB-2519 test mdl_locks is not deterministic

altmannmarcelo:
https://jira.percona.com/browse/PXB-2519

Problem:
Mdl_locks.sh test uses utf8 characters to validate the mdl lock query
executed by PXB will be able to handle those type of characters. However
we are not forcing the connection to use utf8 resulting in some failures
depending on the OS charset.

Fix:
Enforce  UTF-8 charset to mysql connection.

### DIFF
--- a/storage/innobase/xtrabackup/test/t/mdl_locks.sh
+++ b/storage/innobase/xtrabackup/test/t/mdl_locks.sh
@@ -6,7 +6,7 @@ start_server
 
 load_sakila
 
-mysql -e 'CREATE TABLE t_hello_你好(id INT, PRIMARY KEY(id)) ENGINE=InnoDB' test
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-test-param/49/

satya-bodapati (approved):
Thank you for chasing this and fixing it!
It was annoying to see it fail on few platforms.

altmannmarcelo:
@satya-bodapati thanks, the aim is to have all tests passing on 8.0.26 :D

satya-bodapati:
> @satya-bodapati thanks, the aim is to have all tests passing on 8.0.26 :D

That's great marce!

When possible, we should try to cleanup test failures (even known ones) as part of merges.
Not all in one merge, at least a few every merge. Eventually, we should have a clean run.

PR #1127, opened 2021-07-12
Implemented PXB-2550 (move bitmap page track tests to a separate suite)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2550

Fix:
Moved related tests to its own suite.
Updated help section of run.sh


storage/innobase/xtrabackup/test/run.sh
```diff
@@ -44,7 +44,7 @@ Usage: $0 [-f] [-g] [-h] [-s suite] [-t test_name] [-d mysql_basedir] [-c build_
 -g          Debug mode
 -t path     Run only a single named test. This option can be passed multiple times.
 -h          Print this help message
--s suite    Select a test suite to run. Possible values: experimental, gr, keyring, main.
+-s suite    Select a test suite to run. Possible values: bitmap, experimental, gr, keyring, main.
```
satya-bodapati, line 47: this is only help change. Isn't there any code changes that looks for 'bitmap' dir?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/run.sh
altmannmarcelo: @satya-bodapati no, that is only if we want it to be ON by default, which is not the case for bitmap.

altmannmarcelo: here is one example of its run:

```
./run.sh -s bitmap
Using /work/pxb/ins/8.0/xtrabackup-test/test/var as test root
Detecting server version...
Running against Percona Server 8.0.23-14-debug (XtraDB 8.0.23-14)
Using 16 parallel workers
Running suite(s): bitmap

xb_incremental_compressed_bitmap_16kb    w5	[passed]    69
xb_incremental_bitmap_misc               w4	[passed]    71
xb_incremental_compressed_bitmap_8kb     w9	[passed]    80
xb_incremental_compressed_bitmap_4kb     w8	[passed]    118
xb_incremental_compressed_bitmap_2kb     w7	[passed]    147
xb_incremental_compressed_bitmap_1kb     w6	[passed]    166
ib_incremental_force_full_scan           w3	[passed]    228
ib_incremental_bitmap_lsn                w2	[passed]    230
ib_incremental_bitmap                    w1	[passed]    231
Spent 1340 of 245 seconds executing testcases

SUMMARY: 9 run, 9 successful, 0 skipped, 0 failed

```

satya-bodapati: before your change, it is run as part of main, shouldn't bitmap suite be in default suites?

altmannmarcelo: @satya-bodapati the goal is to have 8.0 running with all tests passing. I'm fixing what is fixable. We know bitmap is broken on 8.0 and we are not going to fix it. Instead of just removing the tests, I'm disabling them as we will need to ensure the same type of tests are executed when page tracking is implemented.
There is no point in leaving those tests running and failing if we are not going to fix them. It only makes us used to see test failing and not paying attention.

satya-bodapati: ok, hopefully these testcases will be converted/adjusted when PXB moves to upstream implementation. @rahulmalik87 are you adjusting them in your page tracking PR?

satya-bodapati (approved):

storage/innobase/xtrabackup/test/run.sh
altmannmarcelo: @satya-bodapati yes, I had requested him to make sure we duplicate all bitmap tests on the pagetracking work. I can see page tracking has its own suite with all tests (the 9 we are disabling on this PR) - https://github.com/rahulmalik87/percona-xtrabackup/tree/PXB-8.0-2434/storage/innobase/xtrabackup/test/suites/pagetracking

rahulmalik87: @satya-bodapati I have page-tracking test in page-tracking suite. but need to add this suite as default will do that.

PR #1132, opened 2021-07-14
8.0 Implemented PXB-2554 (logic to skip tests under ASAN is not working)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2554

Problem:
currently we have some logic to skip tests under ASAN that require
$ASAN_OPTIONS variable to be populated. However there is no place that
we populate this variable to let tests know we are running under ASAN.

Fix:
Create proper functions on common.sh to let tests to be skipped under
ASAN.

### DIFF
--- a/storage/innobase/xtrabackup/test/inc/common.sh
+++ b/storage/innobase/xtrabackup/test/inc/common.sh
@@ -885,6 +885,25 @@ function require_tokudb()
     fi
 }
 
+########################################################################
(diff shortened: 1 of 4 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-asan/7/

satya-bodapati (approved):

PR #1146, opened 2021-08-10
PXB-2557 - 8.0.26 merge

altmannmarcelo:
https://jira.percona.com/browse/PXB-2557


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/93/

storage/innobase/dict/dict0dd.cc
```diff
@@ -1007,8 +1007,8 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
     uint64 trx_id = 0;
     dd::Object_id index_space_id = dd_index->tablespace_id();
 
-    if (dd_table->tablespace_id() == dict_sys_t::s_dd_space_id) {
-      sid = dict_sys_t::s_space_id;
+    if (dd_table->tablespace_id() == dict_sys_t::s_dd_dict_space_id) {
+      sid = dict_sys_t::s_dd_dict_space_id;
```
satya-bodapati: sid should be dict_sys_t::s_dict_space_id (no DD here )

satya-bodapati: I think Rahul pointed this already, did you fix this as separate commit?

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
```diff
@@ -419,12 +419,18 @@ static dberr_t create_log_files(char *logfilename, size_t dirnamelen, lsn_t lsn,
...
-    ut_ad(err == DB_SUCCESS);
     if (use_dumped_tablespace_keys && !srv_backup_mode) {
-      xb_insert_tablespace_key(log_space->id, log_space->encryption_key,
-                               log_space->encryption_iv);
+      byte key[Encryption::KEY_LEN];
+      byte iv[Encryption::KEY_LEN];
+
+      bool found = xb_fetch_tablespace_key(log_space->id, key, iv);
+      ut_a(found);
+      err = fil_set_encryption(log_space->id, Encryption::AES, key, iv);
```
satya-bodapati: so this is the unencrypted key right? and this sets only in-mem copy?

Also, we were inserting the key to some hash before? is that not necessary now?

storage/innobase/xtrabackup/src/backup_copy.h
```diff
@@ -68,6 +68,14 @@ bool backup_start(Backup_context &context);
 @return true if success. */
 bool backup_finish(Backup_context &context);
 
+/**
+  Copies redo log encrypted info from xtraback_logfile to ib_lofile0
```
satya-bodapati: ib_logfile0

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -861,6 +861,133 @@ static void page_checksum_fix(byte *page, const page_size_t &page_size) {
...
+    ut_free(log_buf);
+  }
+  return true;
+}
+
+/************************************************************************
+Reencrypt redo header with new master key for copy-back.
+@return true in case of success. */
+static bool reencrypt_redo_header(const char *dir, const char *filepath,
+                                  uint thread_n) {
```
satya-bodapati: use new doxygen style.
Also comment parameters

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -861,6 +861,133 @@ static void page_checksum_fix(byte *page, const page_size_t &page_size) {
...
+  }
+  success = os_file_read(read_request, src_path, src_file, log_buf, 0,
+                         LOG_FILE_HDR_SIZE);
+
+  ulint encryption_offset = LOG_HEADER_CREATOR_END + LOG_ENCRYPTION;
+  success = os_file_write(write_request, dst_path, dst_file,
+                          log_buf + encryption_offset, encryption_offset,
+                          Encryption::INFO_SIZE);
+  if (!success) {
+    msg("  xtrabackup: Fatal error: cannot write encryption to redo log "
```
satya-bodapati: space before xtrabackup can be removed

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -861,6 +861,133 @@ static void page_checksum_fix(byte *page, const page_size_t &page_size) {
...
+    }
+    return false;
+  }
+
+  dst_file = os_file_create_simple_no_error_handling(
+      0, dst_path, OS_FILE_OPEN, OS_FILE_READ_WRITE, srv_read_only_mode,
+      &success);
+  if (!success) {
+    os_file_get_last_error(TRUE);
+    msg("  xtrabackup: Fatal error: cannot find %s.\n", dst_path);
```
satya-bodapati: space before xtrabackup. Remove it everywhere

storage/innobase/xtrabackup/src/keyring_components.cc
```diff
@@ -122,6 +120,33 @@ void create_component_config_data() {
...
+    msg("xtrabackup: Error: failed to load backup-my.cnf\n");
+    return (false);
+  }
+  memset(server_uuid, 0, Encryption::SERVER_UUID_LEN + 1);
+  if (uuid != NULL) {
+    strncpy(server_uuid, uuid, Encryption::SERVER_UUID_LEN);
+  }
+  return (true);
+}
+
```
satya-bodapati: can you move this to separate file? we have common.cc?

storage/innobase/xtrabackup/src/keyring_components.cc
```diff
@@ -122,6 +120,33 @@ void create_component_config_data() {
       std::string(string_buffer.GetString(), string_buffer.GetSize());
 }
 
+bool read_server_uuid() {
+  if (xtrabackup_stats) return true;
```
satya-bodapati: whats special about stats? :)

satya-bodapati (changes requested):

storage/innobase/dict/dict0dd.cc
altmannmarcelo: @satya-bodapati 
```
dict_sys_t::s_dd_space_id -> dict_sys_t::s_dd_dict_space_id
dict_sys_t::s_space_id -> dict_sys_t::s_dict_space_id
```

Seems like the line in the if branch is correct, however, the sdi attribution needs to be to `sid = dict_sys_t::s_dict_space_id;`

The one that was causing the problem and @rahulmalik87 figured out was inside `dict_load_from_spaces_sdi` at `xtrabackup.cc`

storage/innobase/srv/srv0start.cc
altmannmarcelo: correct, `xb_fetch_tablespace_key` reads the key from xtrabackup_keys file.

> Also, we were inserting the key to some hash before? is that not necessary now?

I don't remember a hash, `fil_set_encryption` will set the encryption. We had a hash (a map actually) before for the tablespaces missing encryption. And we would check them again in the end of backup.

storage/innobase/xtrabackup/src/keyring_components.cc
altmannmarcelo: Stats its been my nightmare :D . It can be run against a live server / offline server / backup. For the first two, we don't have backup-my.cnf

storage/innobase/dict/dict0dd.cc
satya-bodapati: Right @altmannmarcelo. Only sid needs to be corrected in the above two lines.

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -953,13 +953,15 @@ static bool reencrypt_redo_header(const char *dir, const char *filepath,
   size_t len = my_read(fd, log_buf, UNIV_PAGE_SIZE_MAX, MYF(MY_WME));
 
   if (len < UNIV_PAGE_SIZE_MIN) {
+    ut_free(log_buf);
```
satya-bodapati: an optional comment:
robert introduced this in PS. Maybe we can copy them to PXB sources :)

ut0new.h (check PS source)
inline void ut_free_func(byte *buf) { ut_free(buf); }

using ut_unique_ptr = std::unique_ptr<byte, std::function<void(byte *)>>;

inline ut_unique_ptr ut_make_unique_ptr_null() {
  return ut_unique_ptr(nullptr, ut_free_func);
}

inline ut_unique_ptr ut_make_unique_ptr_nokey(const size_t size) {
  return ut_unique_ptr(static_cast<byte *>(ut_malloc_nokey(size)),
                       ut_free_func);
}

inline ut_unique_ptr ut_make_unique_ptr(const size_t size,
                                        PSI_memory_key memory_key) {
  return ut_unique_ptr(static_cast<byte *>(ut_malloc(size, memory_key)),
                       ut_free_func);
}

inline ut_unique_ptr ut_make_unique_ptr_zalloc(const size_t size,
                                               PSI_memory_key memory_key) {
  return ut_unique_ptr(static_cast<byte *>(ut_zalloc(size, memory_key)),
                       ut_free_func);
}

inline ut_unique_ptr ut_make_unique_ptr_zalloc_nokey(const size_t size) {
  return ut_unique_ptr(static_cast<byte *>(ut_zalloc_nokey(size)),
                       ut_free_func);
}

altmannmarcelo: @satya-bodapati makes sense, saves us from calling ut_free on a lot of places. Do you think we should make a separate task and rework all places on xtrabackup.cc or only new functions?

altmannmarcelo: ported de code and used for new functions on this PR

storage/innobase/xtrabackup/src/keyring_components.cc
satya-bodapati: Please add a comment before the return true :)

storage/innobase/srv/srv0start.cc
satya-bodapati: So the old xb_insert_tablespace_key() isn't necessary?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6915,6 +6917,14 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
     if (innodb_init(false, false)) goto error;
 
     if (innodb_end()) goto error;
+    /*
+     * we cannot generate encrypted redo log without keyring access.
+     * For redo log, we only have un-encrypted key/iv but don't have original
+     * encrypted version. Copy it from xtrabackup_logfile.
```
satya-bodapati: Copy it from xtrabackup_logfile to the newly created redo log file header

satya-bodapati (approved):
Please check the binaries created in the package or tarballs. (To ensure SKIP_INSTALL is used for all old and new unnecessary binaries).

storage/innobase/xtrabackup/test/t/bug1254227.sh
```diff
@@ -25,6 +25,10 @@ XA END 'xatrx';
 XA PREPARE 'xatrx';
 EOF
 
+# Let the client complete the above set of statements
```
rahulmalik87: do we need to run this in another thread. Can't we just run this sql in before taking backup?

storage/innobase/srv/srv0start.cc
altmannmarcelo: Not anymore, before we were first calling `fil_set_encryption` without key & iv, this would cause InnoDB to generete a new pair of key&iv and encrypt it using the default master key, thus we were required to call `xb_insert_tablespace_key` to insert the new keys to `xtrabackup_keys`. However, we changed the approach to copy the key we already have in `xb_insert_tablespace_key` reutilizing the same key. We now call `fil_set_encryption` passing a pair of key&iv making it not generate a new one.

storage/innobase/xtrabackup/test/t/bug1254227.sh
altmannmarcelo: That is the way it was there before. The failure itself is not related to 8.0.26. It's just because I set the server folder as a symblink to a binary that is located in a NFS in my setup, thus it's a bit slow compared with xtrabackup / datadir / backup dir that are all located on SSD's and in some cases the backup completes before mysql client was invoked.
I just added some (poor) synchronization to overcome the issue on my setup.

altmannmarcelo:
> Please check the binaries created in the package or tarballs. (To ensure SKIP_INSTALL is used for all old and new unnecessary binaries).

@satya-bodapati good point:

```
marcelo@marce-bld:/work/pxb/ins/8.0$ ls bin/
xbcloud  xbcloud_osenv  xbcrypt  xbstream  xtrabackup

marcelo@marce-bld:/work/pxb/ins/8.0$ ls lib/plugin/
component_keyring_file.so  keyring_file.so  keyring_vault.so
```

Just the files we need. Thanks for asking

storage/innobase/srv/srv0start.cc
satya-bodapati: Ok, this much happened with the change of two lines :smile:

storage/innobase/xtrabackup/test/t/bug1254227.sh
rahulmalik87: ack. thanks for clarification.

rahulmalik87 (approved):
LGTM.

PR #1155, opened 2021-08-24
PXB-2574 - Adjust test to make usage of more cores without timing-out

altmannmarcelo:
https://jira.percona.com/browse/PXB-2574

Problem:
Some tests run the same set of steps over and over in a loop changing
some configuration to explore edge cases. Those tests are time
consuming, and often timeout on low end hardware. Splitting them into
multiple tests allow us to divide and conquer and speed up things by
running it in parallel.
Also, some tests restore a datadir with a lower DD version, causing the
next start-up to require to upgrade DD, which sometimes can take more
than 200 seconds (current hard-coded timeout).

Fix:
 - Binlog - Split binlog.sh into multiple tests, created a separate
suite for it and moved the relevant tests into it. Suite default is ON.
 - bug1590351.sh - Added --target-dir so it will have a proper folder
and wont save the backup into xtrabackup_backupfiles which makes it
impossible to run the same test twice.
 - Keyring - split the for loop of KEYRING type into it's own test.
 - Introduced MYSQLD_START_TIMEOUT variable that allows to control the
timeout we wait for a server to be ready for connections.
 - xb_duplicate_sdi.sh / pxb-2455.sh - Those tests restore a lower DD
version datadir. Adjusted the test start_server timeouts to 20 minutes.
 - Adjusted the order of suites, so long running suites start before the
main suite improving overall test time completion.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/104/

storage/innobase/xtrabackup/test/suites/binlog/binlog_10.sh
```diff
@@ -0,0 +1,6 @@
+. inc/common.sh
+. inc/binlog_common.sh
+vlog "------- TEST 10 -------"
+INDEX_FILE="$topdir/binlog-dir1/bin.index"
+FILES="$topdir/binlog-dir1/bin.000003"
```
satya-bodapati, line 5: how do we know that  binlog with .00003 will be created in this test?

storage/innobase/xtrabackup/test/suites/binlog/binlog_101.sh
```diff
@@ -0,0 +1,13 @@
+. inc/common.sh
+. inc/binlog_common.sh
+# do the same, but with updating my.cnf
+
+vlog "------- TEST 101 -------"
+MYSQLD_EXTRA_MY_CNF_OPTS="
+skip-log-bin
+"
+INDEX_FILE=""
```
satya-bodapati, line 9: if index_file is empty, how do we know which binlog index file to use for verification?

storage/innobase/xtrabackup/test/suites/binlog/binlog_10.sh
altmannmarcelo: Hi @satya-bodapati I've not modified the original binlog.sh test. I just split it into single files.
FILES and INDEX_FILE is tested on restore, to validate if the file exists. By looking into all tests, I can see that all that specify a file with the numeric part uses .000003 or .000004 - Most likely those are the starting point on MTR (1 at --initialize, 1 at startup, 1 as part of flush engine logs 1 as part of startup after the restore) = 4

storage/innobase/xtrabackup/test/suites/binlog/binlog_101.sh
altmannmarcelo: INDEX_FILE is tested as `-n` which will be skipped on empty values or null

satya-bodapati (approved):

storage/innobase/xtrabackup/test/suites/binlog/binlog_10.sh
satya-bodapati: ok, test knows upfront that it will rotate the binlogs :+1:

PR #1178, opened 2021-09-23
PXB-2543  Add IB_EXPORT_CFG_VERSION_V6 for exporting cfg file for a t…

rahulmalik87:
…able

https://jira.percona.com/browse/PXB-2543

Problem:
PXB can't export single table with instant column, compression
information and default columns.

Analysis:
PXB export IB_EXPORT_CFG_VERSION_V2 whereas mysql 8.0.23 is using
IB_EXPORT_CFG_VERSION_V6.

Fix:
Generate IB_EXPORT_CFG_VERSION_V6 if using mysql 8.0.23 or later

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -61,6 +61,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "rpl_log_encryption.h"
 #include "space_map.h"
 #include "typelib.h"
+#include "utils.h"
 #include "xb0xb.h"
 #include "xtrabackup.h"
 #include "xtrabackup_version.h"
@@ -365,24 +366,6 @@ void parse_show_engine_innodb_status(MYSQL *connection) {
   mysql_free_result(mysql_result);
 }
 
-/* find the pxb base version */
-static unsigned long pxb_base_version() {
-  std::string pxb_base = MYSQL_SERVER_VERSION;
-  unsigned long major = 0, minor = 0, version = 0;
-  std::size_t major_p = pxb_base.find(".");
-  if (major_p != std::string::npos) major = stoi(pxb_base.substr(0, major_p));
-
-  std::size_t minor_p = pxb_base.find(".", major_p + 1);
-  if (minor_p != std::string::npos)
-    minor = stoi(pxb_base.substr(major_p + 1, minor_p - major_p));
-
-  std::size_t version_p = pxb_base.find(".", minor_p + 1);
-  if (version_p != std::string::npos)
-    version = stoi(pxb_base.substr(minor_p + 1, version_p - minor_p));
-  else
-    version = stoi(pxb_base.substr(minor_p + 1));
-  return major * 10000 + minor * 100 + version;
-}
 
 static bool check_server_version(unsigned long version_number,
                                  const char *version_string,
@@ -429,8 +412,8 @@ static bool check_server_version(unsigned long version_number,
     }
   }
 
-
-  auto pxb_base_ver = pxb_base_version();
+  auto pxb_base_ver =
+      xtrabackup::utils::get_version_number(MYSQL_SERVER_VERSION);
 
   DBUG_EXECUTE_IF("simulate_lower_version", pxb_base_ver = 80014;);
 
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -46,6 +46,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <mysql_version.h>
 #include <mysqld.h>
 #include <sql_bitmap.h>
+#include "row0quiesce.h"
 
 #include <fcntl.h>
 #include <signal.h>
@@ -194,6 +195,8 @@ static hash_table_t *tables_exclude_hash = NULL;
 
 /* map of schema name and schema id */
 static std::map<std::string, uint64> dd_schema_map;
+
+static uint32_t cfg_version = IB_EXPORT_CFG_VERSION_V6;
 
 /* map of <schema id, name> and SDI id*/
 static std::map<std::pair<int, std::string>, uint64> dd_table_map;
@@ -2537,6 +2540,48 @@ static bool innodb_end(void) {
   internal_innobase_data_file_path = NULL;
 
   return (FALSE);
+}
+
+/* Read backup meta info.
+@param[in] filename filename to read mysql version
+@return TRUE on success, FALSE on failure. */
+static bool xtrabackup_read_info(char *filename) {
+  FILE *fp;
+  bool r = TRUE;
+  char mysql_server_version_str[30] = ""; /* 8.0.20.debug| 8.0.20 */
+  unsigned long mysql_server_version;
+
+  fp = fopen(filename, "r");
+  if (!fp) {
+    msg("xtrabackup: Error: cannot open %s\n", filename);
+    return (FALSE);
+  }
+  /* skip uuid, name, tool_name, tool_command, tool_version, ibbackup_version */
+  for (int i = 0; i < 6; i++) {
+    char c;
+    do {
+      c = fgetc(fp);
+    } while (c != '\n');
+  }
+
+  if (fscanf(fp, "server_version = %29s\n", mysql_server_version_str) != 1) {
+    r = FALSE;
+    goto end;
+  }
+  mysql_server_version =
+      xtrabackup::utils::get_version_number(mysql_server_version_str);
+
+  ut_ad(mysql_server_version > 80000 && mysql_server_version < 90000);
+  if (mysql_server_version < 80019) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V3;
+  } else if (mysql_server_version < 80020) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V4;
+  } else if (mysql_server_version < 80023) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V5;
+  }
+end:
+  fclose(fp);
+  return (r);
 }
 
 /* ================= common ================= */
@@ -5978,7 +6023,10 @@ static bool xb_export_cfg_write_index_fields(
                                this index */
     FILE *file)                /*!< in: file to write to */
 {
-  byte row[sizeof(ib_uint32_t) * 2];
+  /* This row will store prefix_len, fixed_len,
+  and in IB_EXPORT_CFG_VERSION_V4, is_ascending */
+  byte row[sizeof(ib_uint32_t) * 3];
+  size_t row_len = sizeof(row);
 
   for (ulint i = 0; i < index->n_fields; ++i) {
     byte *ptr = row;
@@ -5989,7 +6037,16 @@ static bool xb_export_cfg_write_index_fields(
 
     mach_write_to_4(ptr, field->fixed_len);
 
-    if (fwrite(row, 1, sizeof(row), file) != sizeof(row)) {
+    if (cfg_version >= IB_EXPORT_CFG_VERSION_V4)
+    {
+      ptr += sizeof(ib_uint32_t);
+      /* In IB_EXPORT_CFG_VERSION_V4 we also write the is_ascending boolean. */
+      mach_write_to_4(ptr, field->is_ascending);
+    } else {
+      row_len = sizeof(ib_uint32_t) * 2;
+    }
+
+    if (fwrite(row, 1, row_len, file) != row_len) {
       msg("xtrabackup: Error: writing index fields.");
 
       return (false);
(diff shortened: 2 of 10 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2574,6 +2578,10 @@ static bool xtrabackup_read_metadata(char *filename) {
     backup_redo_log_flushed_lsn = 0;
   }
 
+  if (fscanf(fp, "server_version = %lu \n", &mysql_server_version) != 1) {
```
altmannmarcelo: To be honest, I don't like the approach of using the metadata file. We already have this information in the backup info file and we don't need to care about missing entry in file (preparing old backups).

The way this PR implements it, we will have a problem if:
Backup was taken on server 8.0.26 by PXB 8.0.26
Backup was prepared on PXB 8.0.27 . We won't have the matching line on metadata file and we would asume mysql server version 8.0.21 then.

Can you check what is necessary to read server version from info file?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6544,6 +6596,12 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
     exit(EXIT_FAILURE);
   }
 
+  if (mysql_server_version < 80023) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V4;
+  } else if (mysql_server_version < 80026) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V5;
+  }
+
```
altmannmarcelo: We don't need to assign a value for `mysql_server_version` on reading xtrabacckup info file, once we read the line with the server version we can directly assign cfg_version there.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -479,6 +482,7 @@ extern TYPELIB innodb_flush_method_typelib;
 bool mdl_taken = FALSE;
 extern struct rand_struct sql_rand;
 extern mysql_mutex_t LOCK_sql_rand;
+extern unsigned long mysql_server_version;
```
altmannmarcelo: Do we have to `extern` this variable?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6274,6 +6302,17 @@ xb_export_cfg_write_header(
     return (false);
   }
 
+  if (cfg_version >= IB_EXPORT_CFG_VERSION_V5) {
```
altmannmarcelo, line 6342: Can you help me understand this one?
INSTANT is a method of ALTER, why will this be related to EXPORT?

storage/innobase/xtrabackup/test/t/xb_export.sh
```diff
@@ -138,22 +145,32 @@ col5 timestamp, col6 long, col7 date, col8 time, col9 datetime, col10 year,
...
 
 xtrabackup --datadir=$mysql_datadir --prepare --export \
     --target-dir=$backup_dir
 
 ls -tlr $backup_dir/incremental_sample/*cfg
 cfg_count=`find $backup_dir/incremental_sample -name '*.cfg' | wc -l`
-vlog "Verifying .cfg files in backup, expecting 6."
-if [ $cfg_count -ne 6 ]
+vlog "Verifying .cfg files in backup, expecting 10."
+if [ $cfg_count -ne 10 ]
```
altmannmarcelo, line 171: My understanding with the issue is that the import tablespace operation will fail, however, I don't see us covering this on the test case.
Can you enhance the test to validate if the import operation will suceed?

altmannmarcelo (changes requested):
@rahulmalik87 Thanks for picking up this issue. Seems like we lacked full support for a basic functionality (export tablespaces). So kudos to you.

I would like to ask you two things:
1) Address the review comments
2) This issue is clear something we missed on merge process. This (and there are many other things) that won't trown an error on compilation nor on current test cases, so we will miss it. I want us to start to document those items that we know might cause PXB to not have 100% full compatibility with PS/MySQL server and add them to a checklist for the merge process. Can you please add this CFG_VERSION (and any other that you might remember) to https://confluence.percona.com/display/PXB/PXB+notes#PXBnotes-Noneasytospotissues ?

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: fixed, now reading from xtrabackup_info file.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: doing it for better readability of code.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: fixed.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: fixed.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: After adding INSTANT column in a table. Server add that information during export and verify that during import. So replicating similar behaviour.

storage/innobase/xtrabackup/test/t/xb_export.sh
rahulmalik87: My bad, missed the actual import tablespace. fixed it.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: what about versions 3  & 4 ? before the fix, we are handling until 2?

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
```diff
@@ -0,0 +1,43 @@
...
+else
+	vlog "wrong version of CFG $CFG_VERSION"
+	exit -1
+fi
+
+# Terminate the background client
+echo "exit" >&3
+exec 3>&-
+wait $client_pid
+
```
satya-bodapati, line 44: why do we need a background client at all? connection disconnection will delete the .cfg file?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -366,21 +366,21 @@ void parse_show_engine_innodb_status(MYSQL *connection) {
...
-  std::size_t version_p = pxb_base.find(".", minor_p + 1);
+  std::size_t version_p = version_str.find(".", minor_p + 1);
   if (version_p != std::string::npos)
-    version = stoi(pxb_base.substr(minor_p + 1, version_p - minor_p));
+    version = stoi(version_str.substr(minor_p + 1, version_p - minor_p));
   else
-    version = stoi(pxb_base.substr(minor_p + 1));
+    version = stoi(version_str.substr(minor_p + 1));
   return major * 10000 + minor * 100 + version;
 }
```
satya-bodapati: this should be in some util file. Please move it

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2541,6 +2540,45 @@ static bool innodb_end(void) {
...
+    goto end;
+  }
+  mysql_server_version = get_version_number(mysql_server_version_str);
+
+  ut_ad(mysql_server_version > 80000 && mysql_server_version < 90000);
+  if (mysql_server_version < 80023) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V4;
+  } else if (mysql_server_version < 80026) {
+    cfg_version = IB_EXPORT_CFG_VERSION_V5;
+  }
```
satya-bodapati: when 8027 is released, what will cfg_version will be?
and is cfg_version global variable?

Can you please add safety checks. static asserts in PXB code to detect changing CFG versions?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
rahulmalik87: yeah  disconnection will unlock table delete the cfg file.

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: as safety check I have added new test case xb_export_check_cfg.sh which checks what is version of CFG. if version is greater than 6 . it will fail so we will come to know.

rahulmalik87: with PXB 8.0.27. we will have CFG 6.

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
satya-bodapati: ok, I think instead of saying WRONG VERSION, you should say "PXB should have changed the source code to handle the new version"

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: @rahulmalik87 please answer this

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: ideally, if the server maintained current_cfg_version variable, we could have utilized it to assert.
I guess your test case is the only way now

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
```diff
@@ -0,0 +1,43 @@
...
+EOF
+
+
+# Let the client complete the above set of statements
+vlog "waiting for 3 seconds to ensure cfg file is generated"
+sleep 3
+
+CFG_VERSION=`od -j3 -N1 -t x1 -An $mysql_datadir/test/t1.cfg  | head -30 | tr -d ' ' | tr -d '\n'`
+
+if [ "$CFG_VERSION" = "06" ]; then
```
satya-bodapati: are we doing string comparison?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
satya-bodapati: you need a big endian to little conversion.

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
rahulmalik87: added

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: we are also handling version 3, 4 and 6. 3 introduced column default value, 4  introduced asc index and 6 introduced compress tables. all are handled in xtrabackup.cc

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
rahulmalik87: yes it string comparison. I tried playing with this variable. like changing the server version to 07 and test fails with proper error message.

satya-bodapati: @rahulmalik87 I would use the --endian feature of od tool (recent versions) or the manual conversion to a decimal using xxd tool
Examples: SPACE_ID=$(od -An -j38 -D -N4 --endian=big $FILE)

MASTER_KEY_ID=$(xxd -b -s$OFFSET -l4 $FILE |  awk '{print $2 $3 $4 $5}')
echo "MASTER_KEY_ID: $MASTER_KEY_ID ( $((2#$MASTER_KEY_ID)) )"

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_export_check_cfg.sh
rahulmalik87: fixed. using xxd now.

satya-bodapati (commented):
Looks good. Please squash and submit to Jenkins

rahulmalik87:
@satya-bodapati jenkin results https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/125/

PR #1182, opened 2021-09-30
Fixed PXB-2614 (xbstream sparse files support doesn't work on XFS)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2614

Problem:
A whole in a file consists in a free space between data. A whole (at the
time of a write) is done by moving the file pointer (lseek) to a
specific offset beyond the end of the file and then write some data
after that.

Xbstream sparse file support works by having a map of those wholes in
the original file, when extracting each chunk, we re-create those wholes
by doing lseek to advance the file pointer to pass the whole and then
we write the actual data.

XFS has a feature called speculative preallocation
https://linux-xfs.oss.sgi.narkive.com/jjjfnyI1/faq-xfs-speculative-preallocation
Basically when you write some data, it allocates some extra blocks after
the end of the file for the next write.

The problem lies in the fact that for XFS, if the lseek doesn't cross
the space preallocated from the previous write, the next time we write
data to the file this allocation will be made permanent and no
space/whole will be created on the file.

Fix:
Manually punch a whole at the desired offset using fallocate.

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_local.cc
+++ b/storage/innobase/xtrabackup/src/ds_local.cc
@@ -123,17 +123,32 @@ static int local_write(ds_file_t *file, const void *buf, size_t len) {
   return 1;
 }
 
+/**
+  Checks if punch hole via fallocate is supported
+
+  @return true if punch hole is supported
+*/
+static bool is_fallocate_punch_hole_supported() {
+#if defined(HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE)
+  return (true);
+#else
+  return (false);
+#endif /* HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE */
+}
+
 static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
                               size_t sparse_map_size,
                               const ds_sparse_chunk_t *sparse_map) {
   auto local_file = ((ds_local_file_t *)file->ptr);
   File fd = local_file->fd;
+  int seek = 0;
 
   const uchar *ptr = static_cast<const uchar *>(buf);
 
   for (size_t i = 0; i < sparse_map_size; ++i) {
     my_off_t rc;
 
+    seek = my_tell(fd, MYF(MY_WME));
     rc = my_seek(fd, sparse_map[i].skip, MY_SEEK_CUR, MYF(MY_WME));
     if (rc == MY_FILEPOS_ERROR) {
       return 1;
@@ -142,6 +157,11 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
     rc = my_write(fd, ptr, sparse_map[i].len, MYF(MY_WME | MY_NABP));
     if (rc != 0) {
       return 1;
+    }
+
+    if (is_fallocate_punch_hole_supported()) {
+      fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, seek,
+                sparse_map[i].skip);
     }
 
     ptr += sparse_map[i].len;
--- a/storage/innobase/xtrabackup/src/ds_local.cc
+++ b/storage/innobase/xtrabackup/src/ds_local.cc
@@ -28,6 +28,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "common.h"
 #include "datasink.h"
 
+#define PUNCH_HOLE_PLACEHOLDER_FILE "xtrabackup_punch_hole"
 typedef struct {
   File fd;
   size_t last_seek;  // to track last page sparse_file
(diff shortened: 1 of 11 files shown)


altmannmarcelo:
This is hard to test on Jenkins. Here is a link to make sure it doesn't introduce any regression:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/117/

And manual test as per bug report:
```
marcelo@marce-bld:~$ xbstream -C /mnt/xfs/backup -x < /tmp/a.xbs

# Original Datadir
marcelo@marce-bld:~$ du -sh --apparent-size /work/ps/ins/8.0/datadir1/test/a.ibd
520M	/work/ps/ins/8.0/datadir1/test/a.ibd
marcelo@marce-bld:~$ du -sh /work/ps/ins/8.0/datadir1/test/a.ibd
136M	/work/ps/ins/8.0/datadir1/test/a.ibd

# Restored backup at XFS
marcelo@marce-bld:~$ du -sh --apparent-size /mnt/xfs/backup/test/a.ibd
520M	/mnt/xfs/backup/test/a.ibd
marcelo@marce-bld:~$ du -sh  /mnt/xfs/backup/test/a.ibd
136M	/mnt/xfs/backup/test/a.ibd
marcelo@marce-bld:~$

```

rahulmalik87:
Do we need a fix for 2.4? also It should be hole and not whole?

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -144,6 +159,11 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
       return 1;
     }
 
+    if (is_fallocate_punch_hole_supported()) {
```
rahulmalik87: Can you check if we can use server method os_file_punch_hole

altmannmarcelo: @rahulmalik87 - Fixed the commit message (Since there were no code changes, I force-pushed it).

Regarding the os_file_punch_hole. I tried that and also using is_punch_hole_supported from innodb part. However, that would mean that we will need to start compiling the whole InnoDB suite for xbstream and xbcrypt.
Currently, both tools are standalone.

altmannmarcelo: @rahulmalik87 regarding 2.4 - sparse file support at xbstream / incremental is not supported.

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -144,6 +159,11 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
...
+      fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, seek,
+                sparse_map[i].skip);
```
satya-bodapati, line 181: what if the backup is streamed to another host? Even though the host filesystem (XFS/EXT4) supports punch hole, we cannot do punch hole?
For example in this case:
xtrabackup --backup --compress --stream=xbstream --target-dir=./ | ssh user@otherhost "xbstream -x" ??

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ds_local.cc
altmannmarcelo: @satya-bodapati
`xtrabackup --backup --compress --stream=xbstream --target-dir=./` This part will be building the sparse_map
`ssh user@otherhost "xbstream -x" ` This part will be running locally in `otherhost` machine and as long fallocate supports punch hole there we will be calling fallocate.

satya-bodapati: :+1:

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -123,17 +123,32 @@ static int local_write(ds_file_t *file, const void *buf, size_t len) {
...
+  @return true if punch hole is supported
+*/
+static bool is_fallocate_punch_hole_supported() {
+#if defined(HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE)
+  return (true);
+#else
+  return (false);
+#endif /* HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE */
+}
+
```
satya-bodapati: I think this is wrong way to determine if punch hole is supported. Please check os_is_sparse_file_supported()

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ds_local.cc
altmannmarcelo: @satya-bodapati It's the same code as `os_file_punch_hole_posix ` were we check if `HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE` is defined. I just mirrored  it in order to keep xbstream/xbcrypt as standalone tools (not require to compile InnoDB in order for it to work):

https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/os/os0file.cc#L2091

satya-bodapati: we need to do actual fallocate() with punch hole flag. I don't see it in your code.

altmannmarcelo: @satya-bodapati at https://github.com/percona/percona-xtrabackup/pull/1182/files#diff-1affa0169c07f724dbbf3000bde191aa2ee45a03aec011167642618e8c90baa2R162 we call fallocate if HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE returns true:

```
if (is_fallocate_punch_hole_supported()) {
    fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, seek,
              sparse_map[i].skip);
}
```

satya-bodapati: @altmannmarcelo here me out. The way it works is in regular server is 
1. for each tablespace, we do a real fallocate() test. This involves doing a system call falloc(FALLOC_FL_PUNCH_HOLE), check return type, and save that into in-memory tablespace object member of fil_space_t
2. Future IO operations on this tablespace, will consult the above saved result and then do punch hole or not

In PXB
----------
1. Please do real fallocate() system call/test once. 
2. Save the result in global member variable ( we don't need to store per file/tablespace). Stream is always in one destination
3. Use the global variable

altmannmarcelo: @satya-bodapati as per our slack conversation adjusted the punch_hole detection.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3254,6 +3256,7 @@ static void xtrabackup_init_datasinks(void) {
     /* Local filesystem */
     ds_data = ds_meta = ds_redo =
         ds_create(xtrabackup_target_dir, DS_TYPE_LOCAL);
+    punch_hole_supported = ds_data->fs_support_punch_hole;
```
satya-bodapati, line 3259: Can you plz show where this global variable is used?

altmannmarcelo: that is used on write_filt. We don't have access to ds_ctxt_t there:

https://github.com/percona/percona-xtrabackup/pull/1182/commits/8faf47cc4f597cad1eef85ada990143e409fa11c#diff-b2756406f7e5b3b22ae43ae543f1683431534448712d9831fe1c3846d1062b3eL205

satya-bodapati: Thanks, I couldn't make out from code read :D Should we instead use xb_fil_cur_t to store the punch_hole metadata? Because it seems to contain all file-related metadata.

satya-bodapati (approved):
LGTM!

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: As per our slack conversation. We will keep the current path as xb_fil_cur_t is on a per-file basis.

PR #1192, opened 2021-10-28
PXB-2614 Followup-fix xbstream sparse files support doesn't work on XFS

rahulmalik87:
https://jira.percona.com/browse/PXB-2614

fixed the compilation error on macos

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_local.cc
+++ b/storage/innobase/xtrabackup/src/ds_local.cc
@@ -54,6 +54,7 @@ datasink_t datasink_local = {&local_init,         &local_open,  &local_write,
   @return true if punch hole is supported
 */
 static void is_fallocate_punch_hole_supported(ds_ctxt_t *ctxt) {
+#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE
   char fullpath[FN_REFLEN];
   File fd;
   int ret;
@@ -71,6 +72,9 @@ static void is_fallocate_punch_hole_supported(ds_ctxt_t *ctxt) {
     my_close(fd, MYF(MY_WME));
     my_delete(fullpath, MYF(MY_WME));
   }
+#else
+  ctxt->fs_support_punch_hole = false;
+#endif
 }
 
 static ds_ctxt_t *local_init(const char *root) {
@@ -155,7 +159,7 @@ static int local_write(ds_file_t *file, const void *buf, size_t len) {
 static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
                               size_t sparse_map_size,
                               const ds_sparse_chunk_t *sparse_map,
-                              bool punch_hole_supported) {
+                              [[maybe_unused]] bool punch_hole_supported) {
   auto local_file = ((ds_local_file_t *)file->ptr);
   File fd = local_file->fd;
   ulonglong seek = 0;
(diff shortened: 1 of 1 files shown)


storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -70,7 +71,10 @@ static void is_fallocate_punch_hole_supported(ds_ctxt_t *ctxt) {
       ctxt->fs_support_punch_hole = false;
     my_close(fd, MYF(MY_WME));
     my_delete(fullpath, MYF(MY_WME));
+    return;
```
altmannmarcelo: we should not have a return on a void function.

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -70,7 +71,10 @@ static void is_fallocate_punch_hole_supported(ds_ctxt_t *ctxt) {
...
   }
+#endif
```
altmannmarcelo:
```suggestion
#else
  ctxt->fs_support_punch_hole = false;
#endif
```

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -176,10 +180,12 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
       return 1;
     }
 
+#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE
```
altmannmarcelo, line 183: just wondering if this can cause warnings because `punch_hole_supported` is not used in the function?

altmannmarcelo (changes requested):
few minor comments.

Also, please send a jenkins test run.

storage/innobase/xtrabackup/src/ds_local.cc
rahulmalik87: variables FALLOC_FL_PUNCH_HOLE and FALLOC_FL_KEEP_SIZE is not defined in macos.

storage/innobase/xtrabackup/src/ds_local.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/src/ds_local.cc
rahulmalik87: fixed

rahulmalik87:
jenkin results, https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/134/testReport/

storage/innobase/xtrabackup/src/ds_local.cc
altmannmarcelo: What I mean is that on function signature we have a parameter `punch_hole_supported` however on macos, due to preprocessor `HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE` we will never use it. I would prefer that we also extend the `#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE` to function signature. Something like:

```
static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
                              size_t sparse_map_size,
                              const ds_sparse_chunk_t *sparse_map
#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE
                              , bool punch_hole_supported
#endif
){

```

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/ds_local.cc
rahulmalik87: now using  __attribute__((unused))

altmannmarcelo (changes requested):
@rahulmalik87 thanks for your suggestion regarding the unused attribute.
Last request - can you use `[[maybe_unused]]` instead?

rahulmalik87:
> maybe_unused

Thanks for suggestion . Using maybe_unused now.

altmannmarcelo (approved):
LGTM

PR #1193, opened 2021-10-29
PXB-2566 Missing parameter on advanced encryption documentation (8.0)

patrickbirch:
Merged 2.4 version
modified:   source/advanced/encrypted_innodb_tablespace_backups

### DIFF
--- a/storage/innobase/xtrabackup/doc/source/advanced/encrypted_innodb_tablespace_backups.rst
+++ b/storage/innobase/xtrabackup/doc/source/advanced/encrypted_innodb_tablespace_backups.rst
@@ -392,8 +392,8 @@ Backup
 
 .. code-block:: bash
 
-   $ xtrabackup --backup --user=root -p --target-dir=/data/backup \
-   --generate-transition-key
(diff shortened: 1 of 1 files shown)


altmannmarcelo (changes requested):
The actual issue is a few lines above at https://github.com/percona/percona-xtrabackup/blame/percona-xtrabackup-8.0.26-18/storage/innobase/xtrabackup/doc/source/advanced/encrypted_innodb_tablespace_backups.rst#L392


```
The same passphrase should be specified for the `prepare` command:

.. code-block:: bash
   $ xtrabackup --prepare --target-dir=/data/backup
```

We are saying we should use the same passphrase but we are not adding it to the backup command. Please adjust as below:

From:
```
$ xtrabackup --prepare --target-dir=/data/backup
```
To:
```
$ xtrabackup --prepare --target-dir=/data/backup \
--transition-key=MySecetKey
```

altmannmarcelo:
2.4 is also wrong

patrickbirch:
Fixed.

altmannmarcelo (approved):
LGTM

PR #1194, opened 2021-11-03
PXB-2466 Xbcrypt crashes when overwriting an existing file

rahulmalik87:
https://jira.percona.com/browse/PXB-2466

Problem:
Xbcrypt crashes when Encrypt the file again in the same location

Analysis:
Incase of error, Xbcrypt is using my_free to clear object memory
which is created with new operator

Fix
Use the delete operator instead of my_free

### DIFF
--- a/storage/innobase/xtrabackup/test/t/xbcrypt.sh
+++ b/storage/innobase/xtrabackup/test/t/xbcrypt.sh
@@ -18,6 +18,13 @@ run_cmd xbcrypt -i inc/decrypt_v1_test_file.txt \
     -o ${test_file}.xbcrypt \
     -a ${encrypt_algo} -k ${encrypt_key}
 
+vlog "check double encryption should fail"
+run_cmd_expect_failure xbcrypt -i inc/decrypt_v1_test_file.txt \
+    -o ${test_file}.xbcrypt \
+    -a ${encrypt_algo} -k ${encrypt_key}
+
+grep -qi "assert" $OUTFILE && die "Unexpected assertion instead of graceful exit on double encryption"
+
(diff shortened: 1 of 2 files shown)


altmannmarcelo (approved):
LGTM, please check if it happens on 2.4 too

storage/innobase/xtrabackup/test/t/xbcrypt.sh
```diff
@@ -18,6 +18,13 @@ run_cmd xbcrypt -i inc/decrypt_v1_test_file.txt \
     -o ${test_file}.xbcrypt \
     -a ${encrypt_algo} -k ${encrypt_key}
 
+vlog "check double encryption should fail"
+run_cmd_expect_failure xbcrypt -i inc/decrypt_v1_test_file.txt \
+    -o ${test_file}.xbcrypt \
+    -a ${encrypt_algo} -k ${encrypt_key}
+
+grep -qi "assert" $OUTFILE && die "assert during testing double encryption "
```
satya-bodapati: "Unexpected assertion instead of graceful exit on double encryption"

satya-bodapati (changes requested):
Please provide jenkins link

rahulmalik87:
> LGTM, please check if it happens on 2.4 too

It doesn't happen in 2.4. We are not using new, instead my_malloc

storage/innobase/xtrabackup/test/t/xbcrypt.sh
rahulmalik87: fixed

satya-bodapati (approved):

PR #1197, opened 2021-11-09
PXB-2625 - Error in the HTTP2 framing layer / Large backup fails on D…

altmannmarcelo:
…ebian Buster

https://jira.percona.com/browse/PXB-2625

Problem:
In some version of curl after 7.52.1 (default for Debian 9/Stretch) and
version 7.64.0 (default for Debian 10/Buster) a regression was
introduced that caused curl fail to identify a TLS HTTP2 connection was
closed and it tried to re-use the closed connection causing xbcloud
fail to download / upload chunks with 'Error in the HTTP2 framing layer'
message / Curl error code 16 - CURLE_HTTP2.

Fix:
Curl has fixed it at
https://github.com/curl/curl/commit/3f5da4e59a556fc68272a9857a38dd75234d0c04
and delivered it on 7.65.0. For users that cannot update the curl
version and to provide a better user experience by making the backup to
complete we are adding error CURLE_HTTP2 retriable by default.

Thanks to Johan Andersson for reporting and helping troubleshooting.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/http.h
+++ b/storage/innobase/xtrabackup/src/xbcloud/http.h
@@ -341,10 +341,10 @@ class Http_client {
   bool verbose{false};
   std::string cacert;
   std::vector<CURLcode> curl_retriable_errors{
-      CURLcode::CURLE_GOT_NOTHING,      CURLcode::CURLE_OPERATION_TIMEDOUT,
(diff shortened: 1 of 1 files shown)


rahulmalik87 (approved):
LGTM

satya-bodapati (approved):

PR #1201, opened 2021-11-16
Support mysql 5.5.0-m2 database backup

LI8023NING:
MySQL officially supports the 5.5.0-m2 version of the database, and xtrabackup also promises to support the 5.5 version of the database, but after verification, the 5.5.0-m2 database does not support backup, by modifying the source code, I have passed the verification


altmannmarcelo:
Hi @LI8023NING Thanks for your PR, however at this time we will not accept it as version 5.5.0 was not a GA version at the time, furthermore 5.5 is already EOL.

We encourage you to keep contributing to xtrabackup.
Thanks.

LI8023NING:
OK, I understand. Thank you for your reply

PR #1213, opened 2021-12-01
PXB-61 - Enable --strict by default / PXB-2644 - Cannot perform backup with user created with REQUIRE x509 

altmannmarcelo:
https://jira.percona.com/browse/PXB-2644

Problem:
Upstream at commit 1910e12 removed #ifdef HAVE_OPENSSL preprocessor.
However, PXB was still using it in order to decide if we would pass SSL
options to stablished connection.

Fix:
Removed preprocessor from PXB code.
Modified sslopt-vars.h in order for PXB to use the same function that
configure a connection to use SSL. Before we were mirroring the code at
backup_mysql.cc, which makes it hard to notice if something on that code
changed on the include file.

https://jira.percona.com/browse/PXB-61

Problem:
PXB accepts any command line parameters and only produces a warning.
This can be dangerous.

Fix:
Enable --strict by default.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -141,7 +141,6 @@ const char reserved_system_space_name[] = "innodb_system";
 /* This tablespace name is reserved by InnoDB for the predefined temporary
 tablespace. */
 const char reserved_temporary_space_name[] = "innodb_temporary";
-ulong opt_ssl_fips_mode = SSL_FIPS_MODE_OFF;
 
 /* === xtrabackup specific options === */
 char xtrabackup_real_target_dir[FN_REFLEN] = "./xtrabackup_backupfiles/";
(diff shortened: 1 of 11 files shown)


altmannmarcelo:
@satya-bodapati @rahulmalik87 is not ideal, but I've tackled two issues on 1 PR. Without PXB-2644 pxb will thrown an error due to any `--ssl*` options being invalid.

https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/146/#showFailuresLink - suites/keyring/reencrypt.sh test is fixed, option `reencrypt-for-server-id` is only available on 2.4 .

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1826,6 +1827,9 @@ bool xb_get_one_option(int optid, const struct my_option *opt, char *argument) {
     case OPT_XTRA_ENCRYPT_KEY:
       hide_option(argument, &xtrabackup_encrypt_key);
       break;
+
+#include "sslopt-case.h"
+
```
satya-bodapati, line 1832: include in middle of file, intentional?? :thinking:

altmannmarcelo: yes, this file contains the `case OPT_` from SSL options. I've added it on the end of the block, right before print help following the location where client binaries do - https://github.com/percona/percona-server/blob/8.0/client/mysql.cc#L2086

satya-bodapati (approved):

PR #1214, opened 2021-12-01
PXB-2434 Use MySQL page tracking for incremental backup

rahulmalik87:
https://jira.percona.com/browse/PXB-2434

follow up fix. Clang format for read_filt.cc file


altmannmarcelo (approved):
LGTM - circleci is passing too

PR #1222, opened 2021-12-09
PXB-2629 Backup download using xbcloud on Debian-10 not working. (2.4)

patrickbirch:
modified:   source/installation/apt_repo.rst - added section to upgrade curl to 7.74.0

### DIFF
--- a/storage/innobase/xtrabackup/doc/source/installation/apt_repo.rst
+++ b/storage/innobase/xtrabackup/doc/source/installation/apt_repo.rst
@@ -108,6 +108,41 @@ Now you can install *Percona XtraBackup* by running:
   Installing the packages manually like this, you must
   resolve all dependencies and install the missing packages yourself.
 
+Update the Curl utility in Debian 10
+=============================================
+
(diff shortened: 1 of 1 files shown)


storage/innobase/xtrabackup/doc/source/installation/apt_repo.rst
```diff
@@ -108,6 +108,46 @@ Now you can install *Percona XtraBackup* by running:
...
+
+For more details, see `curl #3750 <https://github.com/curl/curl/issues/3750>`__ or `curl #3763 <https://github.com/curl/curl/pull/3763>`__. 
+
+Follow these steps to upgrade curl to version 7.74.0: 
+
+#. Remove the current version of curl:
+
+   .. code-block:: bash
+
+      $ sudo apt remove curl
```
altmannmarcelo: I think that by doing this, we will also remove other software that has curl as a dependency

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/doc/source/installation/apt_repo.rst
patrickbirch: Tested and verified that I can drop this step and curl is updated without issues.

altmannmarcelo (approved):
LGTM

PR #1223, opened 2021-12-13
PXB-2664 --no-server-version-check is not mentioned in option referen…

patrickbirch:
…ce (8.0)

modified:   source/xtrabackup_bin/xbk_option_reference.rst
removed: soure/.res/contents/option.no-version-check.txt


nastena1606 (approved):
LGTM

george-lorch (approved):

PR #1224, opened 2021-12-16
PXB-2662 : pxb with page-tracking crashes during prepare, cannot apply an insert redo on page

rahulmalik87:
PXB-2662 : pxb crashes during prepare, cannot apply an insert redo on page
    https://jira.percona.com/browse/PXB-2662

    Background:
    -----------
    With --page-tracking enabled, incremental backups rely on pages given
    by server's page-tracking component. Another requirement is that PXB
    should copy redo logs in background

    Analysis
    --------
    There are inplace DDLs (like ADD INDEX, OPTIMIZE TABLE etc) which create
    secondary indexes using the BulkLoad (bottom-up build). These DDLs do not
    write redo-log for the secondary index changes.

    A change done in server is either retrieved via page-tracking component or via
    change in redo. But for the above DDLs, which happen after the checkpoint_lsn,
    appear only in page-tracking only at the next checkpoint.

    So for these inplace DDLs that happen after full-backup, PXB during incremental
    backup, cannot find all the changes done to tablespace using page-tracking and
    redo.

    Fix:
    ----
    Server writes a MLOG_INDEX_LOAD redo for the secondary indexes built using BulkLoad.
    So we fall-back to copying the entire IBD file if we find this redo log record.

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -96,6 +96,12 @@ otherwise.  Note that this is false while a background thread is
 rolling back incomplete transactions. */
 volatile bool recv_recovery_on;
 volatile lsn_t backup_redo_log_flushed_lsn;
+extern bool opt_page_tracking;
+extern char *xtrabackup_incremental;
+extern lsn_t incremental_start_checkpoint_lsn;
+
+/**  map of sapce_id, that experienced an inplace DDL during a backup op */
+std::map<space_id_t, bool> index_load_map;
 
 #ifdef UNIV_HOTBACKUP
 std::list<std::pair<space_id_t, lsn_t>> index_load_list;
@@ -1754,7 +1760,7 @@ static byte *recv_parse_or_apply_log_rec_body(
       Error out in case of online backup and emit a warning in case
       of offline backup and continue. */
       if (!recv_recovery_on) {
-        if (mdl_taken) {
+        if (redo_catchup_completed) {
           if (backup_redo_log_flushed_lsn < recv_sys->recovered_lsn) {
             ib::info() << "Last flushed lsn: " << backup_redo_log_flushed_lsn
                        << " load_index lsn " << recv_sys->recovered_lsn;
@@ -1781,6 +1787,24 @@ static byte *recv_parse_or_apply_log_rec_body(
           /** else the index is flushed to disk before
           backup started hence no error */
         } else {
+          // While scaning redo logs during a backup operation a
+          // MLOG_INDEX_LOAD type redo log record indicates, that a DDL
+          // (create index, alter table...) is performed with
+          // 'algorithm=inplace'. If using pagetracking, the affected tablespace
+          // must be copied using full scan. Record it in the index_load_list
+          // We do this during the first scan of redo before starting file copy
+          // thread. This copy is required because Page-tracking relies on
+          // changes from redo as well. ie PXB has to copy redo as well apart
+          // from copying pages from the pages given by page-tracking list.
+          // Since in-place DDL skips redo logging and the pages that are not
+          // yet flushed, will not be tracked by page-tracking. And since the
+          // redo logging is skipped, PXB will fail to get those changed pages
+          // (currently only in buffer pool). hence, we rely on re-copying the
+          // datafiles.
+          if (opt_page_tracking && xtrabackup_incremental != nullptr &&
+              recv_sys->recovered_lsn > incremental_start_checkpoint_lsn) {
+            index_load_map[space_id] = true;
+          }
           /* offline backup */
           ib::info() << "Last flushed lsn: " << backup_redo_log_flushed_lsn
                      << " load_index lsn " << recv_sys->recovered_lsn;
@@ -1788,7 +1812,7 @@ static byte *recv_parse_or_apply_log_rec_body(
           ib::warn(ER_IB_MSG_717);
         }
       }
-#endif /* UNIV_HOTBACKUP */
+#endif /* UNIV_HOTBACKUP || XTRABACKUP */
       if (end_ptr < ptr + 8) {
         return (nullptr);
       }
--- a/storage/innobase/xtrabackup/src/read_filt.cc
+++ b/storage/innobase/xtrabackup/src/read_filt.cc
@@ -28,6 +28,9 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "fil_cur.h"
 #include "xtrabackup.h"
 
+/** map of tablespace_id, that experienced an inplace DDL during a backup op */
+extern std::map<space_id_t, bool> index_load_map;
+
 /****************************************************************/ /**
  Perform read filter context initialization that is common to all read
  filters.  */
@@ -229,7 +232,7 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
       for (i = 0, page = cursor->buf; i < cursor->buf_npages;
            i++, page += page_size) {
         lsn_t page_lsn = mach_read_from_8(page + FIL_PAGE_LSN);
-        ut_ad(page_lsn < incremental_lsn ||
+        ut_ad(page_lsn <= incremental_lsn ||
               page_lsn >= incremental_start_checkpoint_lsn);
       }
     }
@@ -245,7 +248,12 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
    * since the last backup. Hence we copy all changes to mysql.ibd since last
    * backup start_lsn instead of last backup end_lsn.
    * We read all pages by setting read_batch_len to the size of file */
-  if (ctxt->space_id == dict_sys_t::s_dict_space_id) {
+
+  /* if DDL happened on the table after the checkpoint LSN we do full scan on
+  that table. index_load_map is poped during the first scan of redo */
+
+  if (ctxt->space_id == dict_sys_t::s_dict_space_id ||
+      index_load_map[ctxt->space_id] == true) {
     *read_batch_start = ctxt->offset;
     if (ctxt->offset >= ctxt->data_file_size) {
       *read_batch_len = 0;
@@ -307,7 +315,6 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
 
     *read_batch_len = ctxt->filter_batch_end * ctxt->page_size - ctxt->offset;
   }
-
   /* If the page block is larger than the buffer capacity, limit it to
   buffer capacity.  The subsequent invocations will continue returning
   the current block in buffer-sized pieces until ctxt->filter_batch_end
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -485,7 +485,7 @@ extern TYPELIB innodb_flush_method_typelib;
 #include "caching_sha2_passwordopt-vars.h"
 #include "sslopt-vars.h"
 
-bool mdl_taken = FALSE;
+bool redo_catchup_completed = false;
 extern struct rand_struct sql_rand;
 extern mysql_mutex_t LOCK_sql_rand;
 
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -25,7 +25,7 @@ extern bool innodb_log_checksums_specified;
 extern bool innodb_checksum_algorithm_specified;
 
 extern bool opt_lock_ddl_per_table;
-extern bool mdl_taken;
+extern bool redo_catchup_completed;
 
 extern bool use_dumped_tablespace_keys;
 
--- a/storage/innobase/xtrabackup/src/changed_page_tracking.cc
+++ b/storage/innobase/xtrabackup/src/changed_page_tracking.cc
@@ -50,7 +50,7 @@ xb_space_map *init(lsn_t checkpoint_lsn_start, MYSQL *connection)
 
   if (page_start_lsn == page_end_lsn) {
     msg("xtrabackup: pagetracking: incremental backup LSN is same as last "
-        "checkpoint LSN " LSN_PF " not calling mysql component",
+        "checkpoint LSN " LSN_PF " not calling mysql component\n",
         page_start_lsn);
     space_map = new xb_space_map;
     return space_map;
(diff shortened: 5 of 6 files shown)


storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -201,13 +203,41 @@ static void rf_page_tracking_init(xb_read_filt_ctxt_t *ctxt,
...
+
+  /* if DDL happened on the table after the checkpoint LSN we do full scan on
+  that table. index_load_list is poped during the first scan of redo */
+  for (auto space_lsn_pair : index_load_list) {
+    if (space_lsn_pair.second > incremental_start_checkpoint_lsn)
+      spaces[space_lsn_pair.first] = true;
+  }
+
+  return (spaces);
+}
```
satya-bodapati: the redo processing is in background thread and page-copying in another thread. What if the redo thread didn't reach the MLOG_INDEX_LOAD record and page-tracking is processing that space?

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
```diff
@@ -95,10 +95,15 @@ recv_sys_t *recv_sys = nullptr;
 otherwise.  Note that this is false while a background thread is
 rolling back incomplete transactions. */
 volatile bool recv_recovery_on;
+
+#if defined(UNIV_HOTBACKUP) || defined(XTRABACKUP)
```
altmannmarcelo: Do we really need `UNIV_HOTBACKUP` here ?

storage/innobase/xtrabackup/src/read_filt.cc
altmannmarcelo: Can we avoid this function?
Why not read `index_load_list` at `rf_page_tracking_get_next_batch` and check LSN calculation there ?

altmannmarcelo (changes requested):

storage/innobase/log/log0recv.cc
rahulmalik87: Aligning it with upstream code.

altmannmarcelo: I would say we should not add this since we do not add logic behind UNIV_HOTBACKUP.

Also, here you are making `volatile lsn_t backup_redo_log_flushed_lsn;` definition only if XTRABACKUP or UNIV_HOTBACKUP macro is true. Later we use this variable at `recv_parse_or_apply_log_rec_body` inside a `#ifndef UNIV_HOTBACKUP
` macro block, which is invalid. 

In summary, we should not care about `UNIV_HOTBACKUP` as this is upstream way of identifying MEB. We will find places where the preprocessor checks both but that is because upstream has added UNIV_HOTBACKUP, not us.

storage/innobase/xtrabackup/src/read_filt.cc
rahulmalik87: rf_page_tracking_get_next_batch is called multiple times for each space. Doing calculation each time could be expensive operation. Building map would be one time operation.

satya-bodapati: @rahulmalik87 marce's point is we don't need a list and a map. You can make index_load_list as a map and add entries for all MLOG_INDEX_LOAD redo entries.

Then just use the map here.

satya-bodapati: instead of index_load_list, I would say map_full_scan/full_scan_spaces/

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
rahulmalik87: fixed. moved volatile lsn_t backup_redo_log_flushed_lsn where it was previously.

storage/innobase/log/log0recv.cc
```diff
@@ -1781,14 +1787,25 @@ static byte *recv_parse_or_apply_log_rec_body(
           /** else the index is flushed to disk before
           backup started hence no error */
         } else {
+          // While scaning redo logs during a backup operation a
+          // MLOG_INDEX_LOAD type redo log record indicates, that a DDL
+          // (create index, alter table...) is performed with
+          // 'algorithm=inplace'. If using pagetracking, the affected tablespace
+          // must be copied using full scan. Record it in the index_load_list
+          // We do this during the first scan of redo before starting file copy
+          // thread
```
satya-bodapati: Can you please explain the 'must be copied' part?

Page-tracking relies on changes from redo as well. ie PXB has to copy redo as well apart from copying pages from the pages given by page-tracking list. Since in-place DDL skips redo logging and the pages that are not yet flushed,  will not be tracked by page-tracking. And since the redo logging is skipped, PXB will fail to get those changed pages (currently only in buffer pool).

Hence, we rely on re-copying the datafiles.

satya-bodapati (changes requested):
Looks good, minor comment. Can you please add testcase. Atleast for debug server?

storage/innobase/log/log0recv.cc
```diff
@@ -96,6 +96,12 @@ otherwise.  Note that this is false while a background thread is
 rolling back incomplete transactions. */
 volatile bool recv_recovery_on;
 volatile lsn_t backup_redo_log_flushed_lsn;
+extern bool opt_page_tracking;
+extern char *xtrabackup_incremental;
+extern lsn_t incremental_start_checkpoint_lsn;
```
altmannmarcelo, line 101: Where are we initializing those variables? shouldn't those be part of xb0xb.h?

storage/innobase/log/log0recv.cc
```diff
@@ -1781,14 +1787,25 @@ static byte *recv_parse_or_apply_log_rec_body(
...
+          }
           /* offline backup */
           ib::info() << "Last flushed lsn: " << backup_redo_log_flushed_lsn
                      << " load_index lsn " << recv_sys->recovered_lsn;
 
           ib::warn(ER_IB_MSG_717);
         }
       }
-#endif /* UNIV_HOTBACKUP */
+#endif /* UNIV_HOTBACKUP || XTRABACKUP */
```
altmannmarcelo, line 1815: where are we adding the `#ifdef XTRABACKUP` that we are closing here?

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/read_filt.cc
rahulmalik87: changed the name.

storage/innobase/log/log0recv.cc
rahulmalik87: added the comment.

storage/innobase/log/log0recv.cc
rahulmalik87: we are intializing them in xtrabackup.cc. All related variables are initialise
>>lsn_t incremental_lsn;
>>lsn_t incremental_start_checkpoint_lsn;
  lsn_t incremental_to_lsn;
  lsn_t incremental_last_lsn;
  lsn_t incremental_flushed_lsn;

storage/innobase/log/log0recv.cc
rahulmalik87: It is already added 
`      case MLOG_INDEX_LOAD:
  #if defined(UNIV_HOTBACKUP) || defined(XTRABACKUP)`

satya-bodapati:
> Looks good, minor comment. Can you please add testcase. Atleast for debug server?

@rahulmalik87 did you try for testcase against debug server?

rahulmalik87:
> > Looks good, minor comment. Can you please add testcase. Atleast for debug server?
> 
> @rahulmalik87 did you try for testcase against debug server?

Yes, spend lot of time but could not create the DML which causes the crash.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -485,7 +485,7 @@ extern TYPELIB innodb_flush_method_typelib;
 #include "caching_sha2_passwordopt-vars.h"
 #include "sslopt-vars.h"
 
-bool mdl_taken = FALSE;
+bool redo_catchup_completed = FALSE;
```
satya-bodapati: use 'false' not CAPITAL FALSE or TRUE (these are #defines and not standard)

satya-bodapati (approved):
please address minor comment

satya-bodapati:
@rahulmalik87 please submit to jenkins

rahulmalik87:
> @rahulmalik87 please submit to jenkins

https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/160/

storage/innobase/log/log0recv.cc
altmannmarcelo: @rahulmalik87 this is still wrong to me. `xb0xb.h` is the place where we should be externing PXB variables used on shared (pxb / mysql) code.

storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -28,6 +28,9 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "fil_cur.h"
 #include "xtrabackup.h"
 
+/** map of tablespace_id, that experienced an inplace DDL during a backup op */
+extern std::map<space_id_t, bool> index_load_map;
+
```
altmannmarcelo, line 33: Same here, this is confusing and we are following no logic. We should be externing the variable on the header file where it's been declared, not where you will use it. If you use it in two external places, which logic to follow in regards to where to extern it?

You declared this on `log0recv.cc` so the logical place for this extern is on `include/log0recv.h`.

Take for example `index_load_list`:
 - Declared https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/log/log0recv.cc#L107
 - Extern on it's corresponding header file - https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/include/log0recv.h#L56

storage/innobase/log/log0recv.cc
```diff
@@ -1781,14 +1787,32 @@ static byte *recv_parse_or_apply_log_rec_body(
           /** else the index is flushed to disk before
           backup started hence no error */
         } else {
+          // While scaning redo logs during a backup operation a
+          // MLOG_INDEX_LOAD type redo log record indicates, that a DDL
+          // (create index, alter table...) is performed with
+          // 'algorithm=inplace'. If using pagetracking, the affected tablespace
+          // must be copied using full scan. Record it in the index_load_list
```
altmannmarcelo, line 1794: we mention index_load_list here but then add it to index_load_map

storage/innobase/log/log0recv.cc
```diff
@@ -1781,14 +1787,32 @@ static byte *recv_parse_or_apply_log_rec_body(
...
+          // We do this during the first scan of redo before starting file copy
+          // thread. This copy is required because Page-tracking relies on
+          // changes from redo as well. ie PXB has to copy redo as well apart
+          // from copying pages from the pages given by page-tracking list.
+          // Since in-place DDL skips redo logging and the pages that are not
+          // yet flushed, will not be tracked by page-tracking. And since the
+          // redo logging is skipped, PXB will fail to get those changed pages
+          // (currently only in buffer pool). hence, we rely on re-copying the
```
altmannmarcelo, line 1802: > (currently only in buffer pool)

Is this true? if they are not redo logged and only reside on BP, what happens in case of a crash?
I believe they are already flushed to the tablespace, but we don't know which ones, otherwise what is the point of copying the .ibd if the changes are only on BP?

cc: @satya-bodapati

storage/innobase/log/log0recv.cc
```diff
@@ -1781,14 +1787,32 @@ static byte *recv_parse_or_apply_log_rec_body(
...
           backup started hence no error */
         } else {
+          // While scaning redo logs during a backup operation a
+          // MLOG_INDEX_LOAD type redo log record indicates, that a DDL
+          // (create index, alter table...) is performed with
+          // 'algorithm=inplace'. If using pagetracking, the affected tablespace
+          // must be copied using full scan. Record it in the index_load_list
+          // We do this during the first scan of redo before starting file copy
+          // thread. This copy is required because Page-tracking relies on
+          // changes from redo as well. ie PXB has to copy redo as well apart
```
altmannmarcelo, line 1797: remove one of the `as well`.

storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -245,7 +248,12 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
    * since the last backup. Hence we copy all changes to mysql.ibd since last
    * backup start_lsn instead of last backup end_lsn.
    * We read all pages by setting read_batch_len to the size of file */
-  if (ctxt->space_id == dict_sys_t::s_dict_space_id) {
+
+  /* if DDL happened on the table after the checkpoint LSN we do full scan on
```
altmannmarcelo, line 252: Not all DDL, can we be more specific here that is only MLOG_INDEX_LOAD. Also, add a note about `s_dict_space_id`

storage/innobase/log/log0recv.cc
satya-bodapati: I think I gave this comment to Rahul. So there is more to this story.
For all the changes skipped to redo log, they are only in BP. Only after complete flush to disk, we see the redo log record. So copying the IBD on seeing this redo is correct.

So the problematic case is this.  ALTER ADD INDEX  happens after the current checkpoint_lsn (backup has already started copying from this checkpoint_lsn). and the checkpoint_lsn doesn't move

In this case, all data is flushed to disk, redo is written, but the page-tracking data is outdated. We see the changes done by ALTER ADD INDEX from page-tracking, only on next checkpoint.

The main culprit is..well, no one.. actually a design choice. page-tracking data is flushed at checkpoints.

To look at another way, we don't need to copy IBDs for all the MLOG_INDEX_LOAD seen before the current checkpoint_lsn as seen by backup. Because those changed-pages are available at the start of backup (page-tracking flushed those at the checkpoint_lsn seen by us)

The problematic DDLs are only those MLOG_INDEX_LOADs after the checkpoint_lsn seen by backup and when checkpoint_lsn doesn't move.

It took me a while to understand this whole thing but I think I got it right :wink: Hope it is clear to you as well, @altmannmarcelo

satya-bodapati:
@rahulmalik87 Please address comments from @altmannmarcelo as followup fix and create new PR

storage/innobase/log/log0recv.cc
altmannmarcelo: @satya-bodapati understood.

@rahulmalik87 can we adjust the `(parenthesis)` comment to reflect that they are indeed flushed to .idb file.

storage/innobase/xtrabackup/src/read_filt.cc
rahulmalik87: added inplace DDL. note about s_dict_space_id is mentioned above.

PR #1227, opened 2021-12-22
PXB-8.0-2666 Azure changes

rahulmalik87:
https://jira.percona.com/browse/PXB-2666
Follow up fix PXB-1883 support for azure storage in xbcloud
https://jira.percona.com/browse/PXB-1883

Addded environment variable for storage class.
Rename the ENDPOINT environment variable to AZURE_ENDPOINT


altmannmarcelo:
aren't we doing the same changes on 2.4?

PR #1232, opened 2022-01-07
Fixed PXB-2673 - Rocksdb backup fails

altmannmarcelo:
https://jira.percona.com/browse/PXB-2673

Problem:
After the introduction of Page tracking component, PXB might require to
query log_status table more than once in case InnoDB has not passed page
tracking start checkpoint.
Inside backup_mysql module, we have a global variable log_status, which
is defined by the struct log_status_t. This struct currently has two
vectors, one for replication channels and one for rocksdb files. When
parsing log_status rows for those two members, we use push_back method,
which causes the the data to be appended instead of replaced.
As a result, PXB was trying to copy the same file more than once making
the backup to fail.
Modules backup_copy and backup_mysql where both instantiating a variable
with the same name (log_status), which was a bit confusing.

Fix:
Removed log_status variable from backup_copy to make the code easier to
read, as this is redundant to context.log_status.
Added a clear method to log_status_t to remove entries from both
vectors. This new method is called before adding more data to log_status
global variable.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1495,7 +1495,7 @@ static bool backup_rocksdb_checkpoint(Backup_context &context, bool final) {
       copy = std::bind(&backup_rocksdb_files, _1, _2, _3, &result);
 
   auto checkpoint_files =
-      final ? context.myrocks_checkpoint.checkpoint_files(context.log_status)
+      final ? context.myrocks_checkpoint.checkpoint_files(log_status)
             : context.myrocks_checkpoint.data_files();
 
   checkpoint_files.erase(
@@ -1584,7 +1584,7 @@ bool backup_start(Backup_context &context) {
   xb_mysql_query(mysql_connection, "FLUSH NO_WRITE_TO_BINLOG BINARY LOGS",
                  false);
 
-  auto log_status = log_status_get(mysql_connection);
+  log_status_get(mysql_connection);
 
   /* Wait until we have checkpoint LSN greater than the page tracking start LSN.
   Page tracking start LSN is system LSN (lets say 105) and Backup End LSN is
@@ -1599,6 +1599,9 @@ bool backup_start(Backup_context &context) {
         pagetracking::get_pagetracking_start_lsn(mysql_connection);
     debug_sync_point("xtrabackup_after_wait_page_tracking");
     while (true) {
+      DBUG_EXECUTE_IF("page_tracking_checkpoint_behind",
+                      log_status.lsn_checkpoint = 1;
+                      DBUG_SET("-d,page_tracking_checkpoint_behind"););
       if (log_status.lsn_checkpoint >= page_tracking_start_lsn) {
         msg("xtrabackup: pagetracking: Checkpoint lsn is " LSN_PF
             " and page tracking start lsn is " LSN_PF "\n",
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/155/

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1584,7 +1584,7 @@ bool backup_start(Backup_context &context) {
   xb_mysql_query(mysql_connection, "FLUSH NO_WRITE_TO_BINLOG BINARY LOGS",
                  false);
 
-  auto log_status = log_status_get(mysql_connection);
+  context.log_status = log_status_get(mysql_connection);
```
satya-bodapati: Do we need this static global variable at all? I think log_status_get() is disgrace. Returning a global variable :)
AFAIK, even after this change, there will be two log_status. backup_context.log_status and the global log_status.

Will there be two copies of files in these two variables?

storage/innobase/xtrabackup/test/suites/rocksdb/log_status.sh
```diff
@@ -0,0 +1,21 @@
...
+start_server
+init_rocksdb
+
+mysql -e "INSTALL COMPONENT \"file://component_mysqlbackup\""
+mysql -e "CREATE TABLE t_logstatus (a INT AUTO_INCREMENT PRIMARY KEY) ENGINE=ROCKSDB" test
+mysql -e "INSERT INTO t_logstatus (a) VALUES (1), (2), (3)" test
+
+while true ; do
+    echo "INSERT INTO t_logstatus () VALUES ();"
+done | mysql test &
```
satya-bodapati: how do we stop this insert loop (at the end of the test) that is running in background?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: Yes, in the end we had two copies. Adjusted now to use log_status global variable.

storage/innobase/xtrabackup/test/suites/rocksdb/log_status.sh
altmannmarcelo: Good catch, it has bitten me already in the past. Fixed.

altmannmarcelo:
@satya-bodapati sorry, old habits die hard. Did a force push instead of a new commit.

satya-bodapati:
> @satya-bodapati sorry, old habits die hard. Did a force push instead of a new commit.

@altmannmarcelo its ok marce

storage/innobase/xtrabackup/src/backup_mysql.h
```diff
@@ -58,6 +58,10 @@ struct log_status_t {
   lsn_t lsn_checkpoint;
   std::vector<replication_channel_status_t> channels;
   std::vector<rocksdb_wal_t> rocksdb_wal_files;
+  void clear() {
+    channels.clear();
+    rocksdb_wal_files.clear();
```
satya-bodapati: is it OK to clear all fields? :)

altmannmarcelo: yes, they will be populated once again by `log_status_storage_engines_parse` and `log_status_replication_parse` using the new information queried by the log_status table. The main problem with those two is that they are vectors that receive the new data via append mode (push_back). The other info, like for InnoDB we just replace by a normal assignment (var = new_value).

satya-bodapati: Ok then, lets clear everything for clarity purpose

altmannmarcelo: make sense. Done

satya-bodapati (approved):

PR #1243, opened 2022-01-28
PXB-2692 Fix build

EvgeniyPatlan:


vorsel (approved):

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -25,6 +25,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include <mysql/service_mysql_alloc.h>
 #include <mysql_version.h>
 #include <mysys_err.h>
+#include <linux/falloc.h>
```
Bo98, line 28: Shouldn't this be wrapped in a `#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE`?

PR #1261, opened 2022-02-16
PXB-2690 - MySQL 8.0.28 merge 

altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/169/


storage/innobase/fil/fil0fil.cc
```diff
@@ -11872,6 +11872,11 @@ dberr_t fil_open_for_xtrabackup(const std::string &path,
 
   ut_a(space != NULL);
 
+  char *fn = fil_node_create(file.filepath(), n_pages, space, false, false);
+  if (fn == nullptr) {
+    return (DB_ERROR);
+  }
+
```
satya-bodapati, line 11879: any reason to move this code up? I see that fil_space_t* is also available later too

satya-bodapati (commented):
Looks good.  keyring tests show that there is some memory leak? can you please check?

storage/innobase/fil/fil0fil.cc
altmannmarcelo: @satya-bodapati as part of b184bd3 a new assertion was added in get_space_by_id_from_map - https://github.com/mysql/mysql-server/blob/mysql-8.0.28/storage/innobase/fil/fil0fil.cc#L736 . Now the fil_space_t requires to have a file attached to it.

A few lines below, we will call fsp_flags_set_encryption, which calls Fil_shard::get_space_by_id which calls Fil_shard::get_space_by_id_from_map, resulting on failing at the new assertion. 

calling fil_node_create to attach the file to the tablespace earlier fixes the issue.

altmannmarcelo:
> Looks good. keyring tests show that there is some memory leak? can you please check?

@satya-bodapati that is something introduced on rework of keyring components load from KMIP. 8.0 trunk is also failing I need to investigate.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -65,9 +65,9 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
       mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_NO);
 
   while (true) {
-    err = fil_io(IORequest(IORequest::READ), true,
-                 page_id_t(log_sys->files_space_id, 0), univ_page_size, 0,
-                 LOG_FILE_HDR_SIZE, log_hdr_buf, nullptr);
+    err = fil_redo_io(IORequest(IORequest::LOG | IORequest::READ),
```
rahulmalik87, line 68: Why did we change this?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -65,9 +65,9 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
...
+                      page_id_t(log_sys->files_space_id, 0), univ_page_size, 0,
+                      LOG_FILE_HDR_SIZE, log_hdr_buf);
 
     if (err != DB_SUCCESS) {
       xb::error() << "fil_io() failed.";
```
rahulmalik87: can you change the error message to fil_redo_io.

rahulmalik87 (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: b184bd30f94df30a8bf178fc327590c5865d33bc added an assertion to enforce IO on redo is done via fil_redo_io:
https://github.com/altmannmarcelo/percona-xtrabackup/blob/PXB-2690-8.0/storage/innobase/fil/fil0fil.cc#L8475
```
. . .
dberr_t fil_io(const IORequest &type, bool sync, const page_id_t &page_id,
               const page_size_t &page_size, ulint byte_offset, ulint len,
               void *buf, void *message) {
  auto shard = fil_system->shard_by_id(page_id.space());
#ifdef UNIV_DEBUG
  if (!sync) {
    /* In case of async io we transfer the io responsibility to the thread which
    will perform the io completion routine. */
    static_cast<buf_page_t *>(message)->release_io_responsibility();
  }
#endif
  /* For redo IO the fil_redo_io must be used. */
  ut_ad(shard->id() != REDO_SHARD);
  ```

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: Done

rahulmalik87 (approved):
LGTM

satya-bodapati (approved):

PR #1263, opened 2022-02-22
Fixed HD-1549 validate decompression path

altmannmarcelo:
Problem:

Fix:
Quote those path's in order to make it more secure.

### DIFF
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -65,4 +65,5 @@ xb_fetch_tablespace_key(ulint space_id, byte *key, byte *iv);
 }
 #endif
 
+#define SQUOTE(str) "'" << str << "'"
 #endif
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -2219,10 +2219,16 @@ bool
 decrypt_decompress_file(const char *filepath, uint thread_n)
 {
 	std::stringstream cmd, message;
-	char *dest_filepath = strdup(filepath);
+	char buf[FN_LEN];
 	bool needs_action = false;
 
-	cmd << "cat " << SQUOTE(filepath);
+	if (escape_string_for_mysql(&my_charset_utf8_general_ci, buf, 0,
+			                           filepath, strlen(filepath)) == (size_t)-1) {
+		msg_ts("[%02u] Error escaping file: %s\n", thread_n, filepath);
+		return false;
+	}
+	char *dest_filepath = strdup(buf);
+	cmd << "cat " << SQUOTE(buf);
 
  	if (ends_with(filepath, ".xbcrypt") && opt_decrypt) {
  		cmd << " | xbcrypt --decrypt --encrypt-algo="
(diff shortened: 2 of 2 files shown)


rahulmalik87 (approved):
LGTM

storage/innobase/include/xb0xb.h
```diff
@@ -65,4 +65,5 @@ xb_fetch_tablespace_key(ulint space_id, byte *key, byte *iv);
 }
 #endif
 
+#define SQUOTE(str) "'" << str << "'"
```
satya-bodapati, line 68: Looks familiar :D

satya-bodapati (approved):

storage/innobase/include/xb0xb.h
altmannmarcelo: that is called backporting :D

Oneiroi (changes requested):
Please refer to the associated ticket that spawned this PR, the fix is noted as incomplete at this stage and needs further refinement to meet the intended purpose.

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2219,10 +2219,16 @@ bool
 decrypt_decompress_file(const char *filepath, uint thread_n)
 {
 	std::stringstream cmd, message;
-	char *dest_filepath = strdup(filepath);
+	char buf[FN_LEN];
```
Oneiroi, line 2222: Where is `FN_LEN` being defined ? 

I note this is present in https://github.com/percona/percona-xtrabackup/blob/da04f2363f6ed1f601a690cc56856131c6d574b8/include/my_io.h#L79 however this is not being included in this file, I've missed or am unable to identify where FN_LEN is being set (this is set as `256` in the my_io.h header file fwiw)

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2219,10 +2219,16 @@ bool
...
 	bool needs_action = false;
 
-	cmd << "cat " << SQUOTE(filepath);
+	if (escape_string_for_mysql(&my_charset_utf8_general_ci, buf, 0,
```
Oneiroi, line 2225: I presume this is the function that's being called here: https://github.com/percona/percona-xtrabackup/blob/d4cbe0e6d4b4895dda372353f3abab23aa17c96a/mysys/charset.cc#L733-L833

This is pretty exhaustive and even escapes multibyte chars afaict (read: this _may_ also intercept the use of unicode homoglyph's in filenames also). 👍

Oneiroi (approved):
Please see my notes on `FN_LEN`, right now this LGTM though I am also aware that my COVID brain may be causing me to miss something so to err on the side of caution, I will request validation from stakeholders and communicate their comments via the HD ticket, thanks :+1:

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: backup_copy.cc -> #include xtrabackup.h -> #include xbstream.h -> #include <my_io.h>

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: correct, also in that sense we might better use `my_charset_utf8mb4_general_ci` instead so we are covered on the 4th byte too.

PR #1265, opened 2022-02-25
Fix compliation error on systems without fallocate

Bo98:
This include should be guarded behind `HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE`, which is defined when `fallocate` is available.

```
/tmp/percona-xtrabackup-20220221-95072-10aesv0/percona-xtrabackup-8.0.27-19/storage/innobase/xtrabackup/src/ds_local.cc:28:10: fatal error: 'linux/falloc.h' file not found
#include <linux/falloc.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [storage/innobase/xtrabackup/src/CMakeFiles/xbstream.dir/ds_local.cc.o] Error 1
```


rahulmalik87:
LGTM. thank you @Bo98 for you patch.

PR #1267, opened 2022-02-25
Fixed PXB-2711 - Libgcrypt initialization warnings for xtrabackup

altmannmarcelo:
Libgcrypt initialization warnings for xtrabackup

https://jira.percona.com/browse/PXB-2711

Problem:
Version of libgcrypt used on centos7 requires proper Libgcrypt
initialization, otherwise warnings will be issued at syslog.
We currently use centos7 for tarballs and relink linked libraries to
copies we provide unde lib/private.
This causes the warning to appear on centos7 and all OS's using the
tarball.

Fix:
Properly initiate libgcrypt before derive keys.


rahulmalik87 (approved):
LGTM

satya-bodapati (approved):

PR #1269, opened 2022-03-01
Fixed PXB 2714 Memory leak on keyring component.

altmannmarcelo:
https://jira.percona.com/browse/PXB-2714

get_default_mr in keyring_common created a new instance of the default
memory resource every time the funciton was called instead of using a
single static instance.

This commit fixes this by making it a static variable.

Cherry-pick from percona/percona-server@da96718

Added the ability to add ASAN suppressions on PXB MTR
Adjusted plugin and component code to not close dynamic linked libraries
otherwise ASAN report will no be able to display symbols.

### DIFF
--- a/components/keyrings/common/data/global_default_mr.h
+++ b/components/keyrings/common/data/global_default_mr.h
@@ -19,7 +19,7 @@
 
 inline ::boost::container::pmr::memory_resource
     * ::boost::container::pmr::get_default_resource() BOOST_NOEXCEPT {
-  psi_memory_resource *global_default_mr = new psi_memory_resource{};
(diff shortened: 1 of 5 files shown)


dutow (approved):

satya-bodapati (commented):
@altmannmarcelo zsolt is saying we can avoid boost::pmr altogether. I am ok with patch as is. Please check with @dutow once

satya-bodapati (approved):

satya-bodapati:
> @altmannmarcelo zsolt is saying we can avoid boost::pmr altogether. I am ok with patch as is. Please check with @dutow once

as discussed on slack, we can cherry pick the new proposed change later

PR #1271, opened 2022-03-02
Fixed PXB-2717 - skip kmip test with no kmip server is up

altmannmarcelo:
https://jira.percona.com/browse/PXB-2717

Problem:
Current KMIP tests only validate if we are running on PS, however, we
are not validating if a KMIP server is up and running.
Doing it properly is hard as it will require external clients to be
compiled/installed on docker images.

Fix:
Added a ping method that at least validate that we can stablish a TCP
connection to ${KMIP_SERVER_ADDR}:${KMIP_SERVER_PORT}.

### DIFF
--- a/storage/innobase/xtrabackup/test/inc/keyring_kmip.sh
+++ b/storage/innobase/xtrabackup/test/inc/keyring_kmip.sh
@@ -51,3 +51,9 @@ function cleanup_keyring()
 {
   echo "place holder" > /dev/null
 }
+
+function ping_kmip()
+{
+  echo test | socat - TCP4:${KMIP_SERVER_ADDR}:${KMIP_SERVER_PORT} || \
+  return 1
+}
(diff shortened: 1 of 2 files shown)


rahulmalik87:
Is there a way to test it on jenkins or we validate it only local?

altmannmarcelo:
> Is there a way to test it on jenkins or we validate it only local?

@rahulmalik87 will also apply to jenkins. in order to KMIP to work we need to set `${KMIP_SERVER_ADDR}:${KMIP_SERVER_PORT}.` The include file automatically sets it to 127.0.0.1:5696 . Then on Jenkins we will try a tcp connection there and it won't have anything listening. Once we implement KMIP docker images on jenkins the ping will automatically suceed.

satya-bodapati:
> 

@altmannmarcelo Currently, Jenkins doesn't start KMIP server? how did any KMIP based test pass on Jenkins? Or did it fail from day 1 :smile:

altmannmarcelo:
@satya-bodapati it requires PS 8.0.27. Now that it is released it will try to run it.

satya-bodapati (approved):

PR #1274, opened 2022-03-10
Implemented PXB-1818 - Provide a new flag to improve performance by a…

altmannmarcelo:
…uto-tuning the value for --use-memory

https://jira.percona.com/browse/PXB-1818

Implementation:
For this implementation we keep resizing the buffer pool at prepare as
the parsed redo log records fill the buffer pool until we have enough
memory to hold all parsed records or we exceed the limit percentage of
free memory we can consume. A resize means we double the buffer pool
size. On each resize we also double the frames available to load
database pages.

Added a new flag to control the percentage of free memory we can
consume. It's important to note that we check the available
memory each time we are about to resize the buffer pool. To be on the
safe side, we don't consider the amount of memory we already hold on
buffer pool. For example:
If we have a BP of 1G we will go to 2G, if we selected
--use-free-memory-pct=50 to not pass 50% of free memory, this will
require at least 4G free of memory for this resize, not taking into
consideration that we already hold 1G.

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3436,8 +3436,9 @@ static bool xb_double_bp(ulint *max_mem) {
 one iteration or it hit threshold */
   if (srv_buf_pool_size * 2 <=
       (free_memory * xtrabackup_use_free_memory_pct / 100)) {
-    dict_ind_init();
-    if (!dict_boot()) return false;
+    if (dict_sys == nullptr) {
+      if (!dict_boot()) return false;
+    }
     srv_buf_pool_size *= 2;
 
     srv_buf_pool_size = buf_pool_size_align(srv_buf_pool_size);
@@ -3449,9 +3450,6 @@ one iteration or it hit threshold */
     *max_mem =
         UNIV_PAGE_SIZE * (srv_buf_pool_size / UNIV_PAGE_SIZE -
                           (recv_n_pool_free_frames * srv_buf_pool_instances));
-    dict_close();
-    ibuf_close();
-    dict_ind_init();
     return true;
   }
   xb::info() << "Not increasing Buffer Pool as it will exceed "
@@ -3908,6 +3906,11 @@ static void recv_recovery_begin(log_t &log, lsn_t *contiguous_lsn,
                                   &log.scanned_lsn, to_lsn);
 
     start_lsn = end_lsn;
+  }
+
+  if (dict_sys != nullptr) {
+    dict_close();
+    ibuf_close();
   }
 
   DBUG_PRINT("ib_log", ("scan " LSN_PF " completed", log.scanned_lsn));
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3423,6 +3423,45 @@ void recv_reset_buffer() {
   recv_sys->recovered_offset = 0;
 }
 
+#ifdef XTRABACKUP
+static bool xb_double_bp(ulint *max_mem) {
+  if (!xtrabackup_use_free_memory_pct_set) return false;
+
+  /* find availabe free memory */
+  auto free_memory = xtrabackup::utils::free_memory();
+  xb::info() << "Maximum memory for buffer pool:" << free_memory;
+  xb::info() << "Current Buffer Pool Size: " << srv_buf_pool_size;
+
+  /* gradually increase the buffer pool until it can processed redo log in
+one iteration or it hit threshold */
+  if (srv_buf_pool_size * 2 <=
+      (free_memory * xtrabackup_use_free_memory_pct / 100)) {
+    dict_ind_init();
+    if (!dict_boot()) return false;
+    srv_buf_pool_size *= 2;
+
+    srv_buf_pool_size = buf_pool_size_align(srv_buf_pool_size);
+
+    xb::info() << "Increasing the buffer pool to " << srv_buf_pool_size;
+
+    buf_pool_resize();
+    recv_n_pool_free_frames *= 2;
+    *max_mem =
+        UNIV_PAGE_SIZE * (srv_buf_pool_size / UNIV_PAGE_SIZE -
+                          (recv_n_pool_free_frames * srv_buf_pool_instances));
+    dict_close();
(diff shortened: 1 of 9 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -3423,6 +3423,45 @@ void recv_reset_buffer() {
...
+
+    buf_pool_resize();
+    recv_n_pool_free_frames *= 2;
+    *max_mem =
+        UNIV_PAGE_SIZE * (srv_buf_pool_size / UNIV_PAGE_SIZE -
+                          (recv_n_pool_free_frames * srv_buf_pool_instances));
+    dict_close();
+    ibuf_close();
+    dict_ind_init();
+    return true;
```
satya-bodapati: In server, we only do  resize, why do we to close these dict, ibuf structures?

storage/innobase/xtrabackup/src/utils.cc
```diff
@@ -100,5 +106,47 @@ unsigned long get_version_number(std::string version_str) {
...
+  meminfo();
+  return kb_main_total * 1024;
+}
+
+uint free_memory_pct() {
+  unsigned long free = free_memory();
+  unsigned long total = total_memory();
+
+  return (uint)(free * 100 / total);
+}
```
satya-bodapati: we are not using this API at all?

storage/innobase/log/log0recv.cc
satya-bodapati: do we need to clean/evict all the existing buf pool pages?

storage/innobase/xtrabackup/src/utils.cc
```diff
@@ -100,5 +106,47 @@ unsigned long get_version_number(std::string version_str) {
...
+    unused_mem = (int64_t)vm_stats.free_count * (int64_t)page_size;
+    return (unsigned long)unused_mem;
+  }
+  return 0;
+}
+
+unsigned long total_memory() {
+  unsigned long free_mem = sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGESIZE);
+  return free_mem;
+}
```
satya-bodapati: variable name and function names dont match

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
altmannmarcelo: The problem here is that on the server we have all those structures already initialized prior to allowing users to resize the buffer pool. Here we are on a stage pre DD initialization. We need to:
1) Initialize them prior to resize.
2) close them otherwise startup will fail when we reach the point where DD will be initialized.

We can either add a check at server startup code to validate if dic_sys is not a nullptr or adjust recv_recovery_begin to close it in case  xb_double_bp has initialized it. I would say the latter is better. 

I've added a new commit to adjust it

storage/innobase/xtrabackup/src/utils.cc
altmannmarcelo: Not in use at the moment. I had PXB-2710 in mind and decided to expose all memory APIs (free / total / free pct).
I can remove if it makes sense.

storage/innobase/log/log0recv.cc
altmannmarcelo: not that I know of. As part of the testing, we validate that DB state is the same as prior to backup to ensure we are not corrupting the database or missing records.

storage/innobase/xtrabackup/src/utils.cc
altmannmarcelo: fixed

storage/innobase/log/log0recv.cc
satya-bodapati: @altmannmarcelo These are not big structs, I would rather initialize them on startup unconditionally and free at the end.

storage/innobase/log/log0recv.cc
```diff
@@ -3423,6 +3423,43 @@ void recv_reset_buffer() {
...
+    if (dict_sys == nullptr) {
+      if (!dict_boot()) return false;
+    }
+    srv_buf_pool_size *= 2;
+
+    srv_buf_pool_size = buf_pool_size_align(srv_buf_pool_size);
+
+    xb::info() << "Increasing the buffer pool to " << srv_buf_pool_size;
+
+    buf_pool_resize();
```
satya-bodapati, line 3448: can this ever fail? lets say current free mem say 500M, and our new bp resize also requires 500MB memory and the above kernal call says its available but by the time we do resize, if such memory is not available, do we crash?

storage/innobase/log/log0recv.cc
```diff
@@ -3423,6 +3423,43 @@ void recv_reset_buffer() {
...
+#ifdef XTRABACKUP
+static bool xb_double_bp(ulint *max_mem) {
+  if (!xtrabackup_use_free_memory_pct_set) return false;
+
+  /* find availabe free memory */
+  auto free_memory = xtrabackup::utils::free_memory();
+  xb::info() << "Maximum memory for buffer pool:" << free_memory;
+  xb::info() << "Current Buffer Pool Size: " << srv_buf_pool_size;
+
+  /* gradually increase the buffer pool until it can processed redo log in
```
satya-bodapati, line 3435: crossing 80 characters?

storage/innobase/log/log0recv.cc
```diff
@@ -3423,6 +3423,43 @@ void recv_reset_buffer() {
...
+
+    srv_buf_pool_size = buf_pool_size_align(srv_buf_pool_size);
+
+    xb::info() << "Increasing the buffer pool to " << srv_buf_pool_size;
+
+    buf_pool_resize();
+    recv_n_pool_free_frames *= 2;
+    *max_mem =
+        UNIV_PAGE_SIZE * (srv_buf_pool_size / UNIV_PAGE_SIZE -
+                          (recv_n_pool_free_frames * srv_buf_pool_instances));
```
satya-bodapati, line 3452: does this mean, as we double BP, more memory from BP is allocated for recovery pages and not for redo data? :thinking:

satya-bodapati (changes requested):

PR #1276, opened 2022-03-14
PXB-2264 Auto increment information is removed after partial restore …

rahulmalik87:
…of a table

https://jira.percona.com/browse/PXB-2264

Problem:
If a table has next auto incremental value 2 then the table exported
through PXB has next auto incremental value 1 instead of 2.

Analysis:
To get the next autoinc, PXB fetches the current autoinc value from DD tables
and increase it by 1. Also it intialize the autoinc value as 1. There is a
addition check to not increase the autoinc when DD autoinc is same as current
table autoinc. This means autoinc is already correctected by metadata applier
through open table.
Subject table has autoinc in  1 in DD table and it is also initialize with value 1
So PXB do not increase the counter and with export this value is set as 1.

Fix:
PXB intialize autoinc_persisted with 0. So Increase the counter
if autoinc persisisted is not same as autoinc presisted from DD tables.

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -2546,13 +2546,20 @@ struct metadata_applier {
   @param[in]      table   table to visit */
   void operator()(dict_table_t *table) const {
     ut_ad(dict_sys->dynamic_metadata != nullptr);
+#ifndef XTRABACKUP
     uint64_t autoinc = table->autoinc;
+#endif /* XTRABACKUP */
     dict_table_load_dynamic_metadata(table);
     /* For those tables which were not opened by
     ha_innobase::open() and not initialized by
     innobase_initialize_autoinc(), the next counter should be
     advanced properly */
+#ifdef XTRABACKUP
+    /* PXB doesn't open tables via ha_innobase::open() */
+    if (table->autoinc != ~0ULL) {
+#else
     if (autoinc != table->autoinc && table->autoinc != ~0ULL) {
+#endif /* XTRABACKUP */
       ++table->autoinc;
     }
   }
--- a/storage/innobase/xtrabackup/test/t/xb_export.sh
+++ b/storage/innobase/xtrabackup/test/t/xb_export.sh
@@ -1,9 +1,13 @@
 . inc/common.sh
-if ! is_xtradb
+table_count=9
+NUMBER_OF_CFG_FILES=11
+if is_server_version_higher_than 8.0.28
 then
-  require_server_version_higher_than 8.0.31
+  NUMBER_OF_CFG_FILES=13
+  table_count=11
 fi
 
+vlog "total table to test $table_count "
 function init_schema() {
     local database_name=$1
     local table_name=$2
@@ -14,7 +18,7 @@ function init_schema() {
     cmd+="create table $database_name.$table_name ("
     for ((c=1; c<=$columns_count; ++c ))
     do
-        cmd+="\`c$c\` int,"
+        cmd+="c$c int,"
     done
     for ((c=1; c<=$columns_count; ++c ))
     do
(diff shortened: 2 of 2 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -2932,12 +2932,19 @@ struct metadata_applier {
...
+    ib_uint64_t autoinc_persisted = table->autoinc_persisted;
     dict_table_load_dynamic_metadata(table);
     /* For those tables which were not opened by
     ha_innobase::open() and not initialized by
     innobase_initialize_autoinc(), the next counter should be
-    advanced properly */
+    advanced properly.  */
+#if defined(XTRABACKUP)
+    if ((autoinc != table->autoinc ||
+         autoinc_persisted != table->autoinc_persisted) &&
```
altmannmarcelo: `autoinc_persisted != table->autoinc_persisted`
This condition will always be false as you are setting `autoinc_persisted=table->autoinc_persisted` on line 2935. Am I missing something?

storage/innobase/xtrabackup/test/t/xb_export.sh
```diff
@@ -136,59 +144,46 @@ vlog "Re-initializing database server"
...
+done
 
 xtrabackup --datadir=$mysql_datadir --prepare --export \
     --target-dir=$backup_dir
 
 ls -tlr $backup_dir/incremental_sample/*cfg
 cfg_count=`find $backup_dir/incremental_sample -name '*.cfg' | wc -l`
-vlog "Verifying .cfg files in backup, expecting 10."
-if [ $cfg_count -ne 10 ]
+vlog "Verifying .cfg files in backup, expecting 11."
```
altmannmarcelo: are we adding a new config on export ?

altmannmarcelo (changes requested):
Can you describe why the server doesn't face the same issue? Shouldn't we match the server behavior?

storage/innobase/srv/srv0start.cc
rahulmalik87: table->autoinc value and table->autoinc_persisted value would be updated inside dict_table_load_dynamic_metadata based on value DD table has.

storage/innobase/xtrabackup/test/t/xb_export.sh
rahulmalik87: we were exporting 10 cfg now we would be exporting 11 . new config added as test9

rahulmalik87:
> Can you describe why the server doesn't face the same issue? Shouldn't we match the server behavior?

Server open the table before doing FLUSH table for export. and as part of open It call MetadataRecover::apply() which fixes this with same logic.

satya-bodapati:
> > Can you describe why the server doesn't face the same issue? Shouldn't we match the server behavior?
> 
> Server open the table before doing FLUSH table for export. and as part of open It call MetadataRecover::apply() which fixes this with same logic.

@rahulmalik87 Can we call the dictmetadata apply during startup?

satya-bodapati (commented):
@rahulmalik87 did you trying loading the autoinc metadata at startup?

rahulmalik87:
> @rahulmalik87 did you trying loading the autoinc metadata at startup?

Hello @satya-bodapati , we load the dynamic metadata at the startup that is where the fix is inside metadata_applier.

satya-bodapati:
> > @rahulmalik87 did you trying loading the autoinc metadata at startup?
> 
> Hello @satya-bodapati , we load the dynamic metadata at the startup that is where the fix is inside metadata_applier.

@rahulmalik87 I mean, do same as server, call  MetadataRecover::apply(). Another option is to write back to DD on full backup prepare

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6952,6 +6953,8 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
       }
 
       /* Write MySQL 8.0 .cfg file */
+      /* load the dynamic metadata */
+      dict_table_load_dynamic_metadata(table);
```
satya-bodapati: if we load right value here, do we still need the above "autoinc_persisted != table->autoinc_persisted" change?

storage/innobase/srv/srv0start.cc
satya-bodapati: am not convinced with this change because this is the not the right way to determine autoinc or the way to load autoinc value

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: yes, we would still need these changes.  autoinc != table->autoinc doesn't work when table has just one row. 

If table has just one row. then table->autoinc would be 1. Also, MySQL initialise autoinc with 1. So this condition will not be satisfied. 

I proposed the fix of using autoinc_persisted because server initialise autoinc_persisted with 0. and after loading the dynamic metadata table-> autoinc_persisted would be 1. 

if you want  limit the scope of changes by introducing another condition autoinc =1 ?

storage/innobase/srv/srv0start.cc
```diff
@@ -2950,15 +2950,21 @@ struct metadata_applier {
   @param[in]      table   table to visit */
   void operator()(dict_table_t *table) const {
     ut_ad(dict_sys->dynamic_metadata != nullptr);
+#ifndef XTRABACKUP
     uint64_t autoinc = table->autoinc;
+#endif /* XTRABACKUP */
     dict_table_load_dynamic_metadata(table);
```
satya-bodapati, line 2552: we can make it simpler by incrementing right after dict_table_load_dynamic_metadata()
#ifdef XTRABACKUP
/*We don't open table by ha_innobase::open. So advance the counter for PXB
opened tables */
++table->autoinc
return;
#endif /*XTRABACKUP */

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
satya-bodapati: will this be done only once? :thinking:

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
rahulmalik87: I tried 2 scenarios normal insert and one where rollback is required(begin; insert into t1 values(null); )
 In both of the cases, it was done once. not sure if we have a case where will be called again. 
 but I added the case to check upper limit

storage/innobase/srv/srv0start.cc
```diff
@@ -2967,13 +2967,20 @@ struct metadata_applier {
...
     /* For those tables which were not opened by
     ha_innobase::open() and not initialized by
     innobase_initialize_autoinc(), the next counter should be
     advanced properly */
+#ifdef XTRABACKUP
+    /* For PXB we do no open tables */
```
satya-bodapati: /* PXB doesn't open tables via ha_innobase::open() */

satya-bodapati (changes requested):
Also next commit, please squash, fix commit message and also add jenkins link

satya-bodapati:
@rahulmalik87 waiting for Jenkins Link

satya-bodapati (changes requested):
@rahulmalik87 see if you can get jenkins run on 8.0.29 or do manual run.

rahulmalik87:
Jenkin reslut https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/285/#showFailuresLink

satya-bodapati (approved):

PR #1277, opened 2022-03-15
PXB-2722 password is written into xtrabackup_info

rahulmalik87:
https://jira.percona.com/browse/PXB-2722

Analysis:
password passed with -p option is written into backup tool_command in xtrabackup_info

Fix:
mask password before writting into xtrabackup_info

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -2088,6 +2088,9 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
 		if (strncmp(*argv, "--password", strlen("--password")) == 0) {
 			arg = "--password=...";
 		}
+		if (strncmp(*argv, "-p", strlen("-p")) == 0) {
+			arg = "-p...";
+		}
 		if (strncmp(*argv, "--encrypt-key",
 				strlen("--encrypt-key")) == 0) {
 			arg = "--encrypt-key=...";
--- a/storage/innobase/xtrabackup/test/inc/common.sh
+++ b/storage/innobase/xtrabackup/test/inc/common.sh
@@ -1150,5 +1150,31 @@ function backup_and_restore() {
   verify_db_state $db
 }
 
+### kill all running queries pattern is passed as argument or WHERE condition
+### example kill_condition_queries "time > 0 & INFO like '%SLEEP%'"
+function kill_query_pattern()
+{
+  local condition=$1
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where $condition into outfile '$MYSQLD_TMPDIR/killcondition.sql';
+  source $MYSQLD_TMPDIR/killcondition.sql;
+EOF
+  rm -f $MYSQLD_TMPDIR/killcondition.sql
+}
+
+### wait for file to be generated()
+### It is used to ensure backup has been started by checking if xtrabackup_logfile
+## is generated
+function wait_for_file_to_generated() {
+  local file=$1
+  vlog "Wait for $file to be generated"
+  while [ ! -r "$file" ]
+  do
+    sleep 1
+    i=$((i+1))
+    echo "Waited $i seconds for $file to be created"
+  done
+}
 # To avoid unbound variable error when no server have been started
 SRV_MYSQLD_IDS=
--- a/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
+++ b/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
@@ -1,0 +1,91 @@
+###############################################################################
+# check if for any password store in data dir thread info or in history table
+###############################################################################
+
+###############################################################################
+
+. inc/common.sh
+password_string=secretpwd
+
+MYSQLD_EXTRA_MY_CNF_OPTS="
+secure-file-priv=$TEST_VAR_ROOT
+"
+vlog "Preparing server"
+start_server
+mysql -e "CREATE USER pxb@localhost IDENTIFIED BY '$password_string'" mysql -e "GRANT ALL ON *.* TO pxb@localhost"
+### create table in MyISAM to let backup wait until there is pending update
+mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
+mysql -e "INSERT INTO t1 values(1)" test
+mkdir -p $topdir/backup
+
+#pass option like password -p to be run with
+
+function grep_in_datadir_ps() {
+
+  additional_options=$1
+  ### The running update on myisam table will hold access to the table.
+  ### PXB will do FLUSH TABLES on MyISAM tables which requires access
+  ### to the table and waits for the running update on MyISAM to finish.
+  mysql -e "SELECT SLEEP(1000) from t1 for update " test &
+
+  $XB_BIN $XB_ARGS --backup -u pxb $additional_options\
+  --transition-key=$password_string --target-dir=$topdir/backup \
+  2>&1 | tee $topdir/pxb.log &
+  job_pid=$!
+
+  echo "backup pid is $job_pid"
+
+  file=$topdir/backup/xtrabackup_logfile
+  if [[ $additional_options == *"encrypt"* ]]; then
+	  file+=".xbcrypt"
+  fi
+  echo "looking for $file"
+  wait_for_file_to_generated $file
+  vlog "check for $password_string string in ps with $job_pid "
+  if ps -ef | grep $job_pid | grep "xtrabackup.*$password_string" | grep -v grep
+  then
+     die "found $password_string string in ps output"
+  fi
+
+  ###kill running mysql query on MyISAM table to let backup resume
+  vlog "killing mysql query"
+  kill_query_pattern "time > 0 & INFO like '%SLEEP%'"
+
+  run_cmd wait $job_pid
+
+  vlog "check for $password_string string in directory $topdir"
+  if grep -rq $password_string $topdir
+  then
+     grep -r $password_string $topdir
+     die "found $password_string string in $topdir "
+  fi
+  rm -rf $topdir/backup
+}
+
+function grep_in_history_table() {
+  additional_options=$1
+  run_cmd $XB_BIN $XB_ARGS --history --backup -u pxb $additional_options\
+  --transition-key=$password_string --target-dir=$topdir/backup 2>&1 | tee $topdir/pxb.log
+
+  row_count=`$MYSQL $MYSQL_ARGS -Ns -e "select count(1) from PERCONA_SCHEMA.xtrabackup_history where tool_command like '%$password_string%'" | awk {'print $1'}`
+  if [ "$row_count" != 0 ]; then
+    mysql -e "select * from PERCONA_SCHEMA.xtrabackup_history where tool_command like '%$password_string%'"
+    die "found entry of $password_string in history table"
+  fi
+  rm -rf $topdir/backup
+}
+
+vlog "check for -p and  encryption string in history table"
+grep_in_history_table "-p$password_string --encrypt=AES256 --encrypt-key=r______$password_string"
+
+vlog "check for --password and encryption string in history table"
+grep_in_history_table "--password=$password_string --encrypt=AES256 --encrypt-key=r______$password_string"
+
+vlog "check for $password_string with -p option in data directory"
+grep_in_datadir_ps "-p$password_string"
+
+vlog "check for $password_string with --password option in data directory"
+grep_in_datadir_ps "--password=$password_string"
+
+vlog "check for encryption string in data directory"
+grep_in_datadir_ps "--password=$password_string --encrypt=AES256 --encrypt-key=r______$password_string"


altmannmarcelo:
@rahulmalik87 I see the test covering the history table, however, the commit mentions it has been written to xtrabackup_info file. Can you add that to the test case?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2088,7 +2088,10 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
 		if (strncmp(*argv, "--password", strlen("--password")) == 0) {
 			arg = "--password=...";
 		}
-		if (strncmp(*argv, "--encrypt-key",
+                if (strncmp(*argv, "-p", strlen("-p")) == 0) {
+                  arg = "-p...";
+                }
+                if (strncmp(*argv, "--encrypt-key",
```
satya-bodapati: plz check alignment. Also check new parameters that are passing cloud info credentials.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -0,0 +1,80 @@
...
+mysql -e "GRANT ALL ON *.* TO pxb@localhost "
+mkdir -p $topdir/backup
+
+# pass either --password or -p
+function grep_for_password() {
+  password=$1$password_string
+  $XB_BIN $XB_ARGS --backup -u pxb $password \
+  --encrypt=AES256 --encrypt-key=r______$password_string --transition-key=$password_string \
+  --target-dir=$topdir/backup \
+  --debug-sync="data_copy_thread_func"  2>&1 | tee $topdir/pxb.log &
```
altmannmarcelo: Do we need debug_sync here? This will make the test not run in half of platforms. 
Perhaps use some approach that will block FTWRL for non-innodb tables, like have SELECT *, SLEEP(6000) FOR UPDATE or something like that.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -0,0 +1,80 @@
...
+  while [ ! -r "$pid_file" ]
+  do
+      sleep 1
+      i=$((i+1))
+      echo "Waited $i seconds for $pid_file to be created"
+  done
+  xb_pid=`cat $pid_file`
+
+  vlog "check for $password_string string in ps "
+  if ps -ef | grep '$password_string' | grep -v grep
```
altmannmarcelo: not that this would be a big issue, but I think we have a potential for other tests running in parallel to interfere, like history_on_server.sh which uses the same password and being executed on a stage prior to stripping the password out. I would indicate use to limit `ps` for the current xtrabackup PID.

Also, I don't think we have any other test that uses `ps`. If I am not mistaken, `ps` is provided by procps library which is not installed by default on Jenkins images.

I have a JIRA ticket to install in as it's needed on PXB-1818

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: agree @altmannmarcelo it would be better to include xtrabackup.

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: thank you @satya-bodapati for cloud suggestion. would be adding xbcloud test-case as part of PXB-2723.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -11,37 +11,26 @@ vlog "Preparing server"
...
 mysql -e "GRANT ALL ON *.* TO pxb@localhost "
+mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
+mysql -e "INSERT INTO t1 values(1)" test
 mkdir -p $topdir/backup
 
 #pass option like password -p to be run with 
 function grep_in_datadir_ps() {
-	additional_options=$1
+  additional_options=$1
+  mysql -e "SELECT SLEEP(15) from t1 for update" test &
```
altmannmarcelo: I would make this a higher sleep, like 6000 and capture it's PID and kill it at the end of the backup. Jenkins can be unpredictable on this sense and 15 seconds might no be enough.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -11,37 +11,26 @@ vlog "Preparing server"
...
-  while [ ! -r "$pid_file" ]
-  do
-      sleep 1
-      i=$((i+1))
-      echo "Waited $i seconds for $pid_file to be created"
-  done
-  xb_pid=`cat $pid_file`
-
+  #sleep sufficient time to let backup run
+  sleep 5
```
altmannmarcelo: Same here, sleep won't be reliable. I would rather either grep the output file and check for a string that indicates the process has started, or keep running a kill -0 on pxb PID.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -11,37 +11,26 @@ vlog "Preparing server"
...
   vlog "check for $password_string string in ps "
-  if ps -ef | grep '$password_string' | grep -v grep
+  if ps -ef | grep "xtrabackup.*$password_string" | grep -v grep
```
altmannmarcelo: This will not isolate the ps to only capture xtrabackup been execute on this test. You should run instead

```
ps -ef | grep "${job_pid}"  | grep "${password_string}" | grep -v grep
```

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -11,37 +11,26 @@ vlog "Preparing server"
...
   then
      die "found $password_string string in ps output"
   fi
 
-  # Resume xtrabackup
-  vlog "Resuming xtrabackup"
-  kill -SIGCONT $xb_pid
-
   run_cmd wait $job_pid
```
altmannmarcelo, line 54: here we should make sure we kill the SELECT SLEEP(X)

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: added

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: fixed.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -15,22 +18,31 @@ mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
...
   2>&1 | tee $topdir/pxb.log &
   job_pid=$!
-  #sleep sufficient time to let backup run
-  sleep 5
-  vlog "check for $password_string string in ps "
-  if ps -ef | grep "xtrabackup.*$password_string" | grep -v grep
+
+  wait_for_process_to_start $job_pid
+  #sleep is to ensure password is removed by xtrabackup
+  sleep 3
```
satya-bodapati: Does this mean there is window that xtrabackup shows password?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2088,7 +2088,10 @@ char *make_argv(char *buf, size_t len, int argc, char **argv)
 		if (strncmp(*argv, "--password", strlen("--password")) == 0) {
 			arg = "--password=...";
 		}
-		if (strncmp(*argv, "--encrypt-key",
+		if (strncmp(*argv, "-p", strlen("-p")) == 0) {
+			arg = "-p...";
+		}
+                if (strncmp(*argv, "--encrypt-key",
```
satya-bodapati: please indent this properly

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: yes there is window when xtrabackup shows password.  PXB removes the argument passed to script.

satya-bodapati: afair, this is done as soon as soon the argument processing is done. Shouldn't it be less than fraction of second? does this test fail without the sleep?

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,27 @@ function backup_and_restore() {
...
 
+### kill all running queries
+function kill_all_queries()
+{
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where user='root' and time > 1 into outfile '$MYSQLD_TMPDIR/killall.sql';
+  source $MYSQLD_TMPDIR/killall.sql;
+EOF
+  rm -f $MYSQLD_TMPDIR/killall.sql
```
satya-bodapati: isn't it better if we just kill the pid of the MySQL client process? I think such function already exists?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: killing pid is not help full. the process still exists can be seen with show processlist.

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: yes this is done after argument processing. yes test fails randomly without sleep.

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: well, we have to use wait $pid like it is done in other tests.
It looks like this code is duplicated from kill_long_selects.sh. Please move to common place

It is not clear to me if "bg_kill_ok" is used, why do we do "kill_all_queries". Please check with marce

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
satya-bodapati: can you please check (by code or test) if the same behaviour applies for "mysql" client tool as well?

rahulmalik87: yes. same thing happen with mysql client as well.

satya-bodapati: ok

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: what is the wrong indentation?

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: these is code from 2013. do you have any idea @altmannmarcelo   but I have removed duplicated code from kill_long_selects.sh.  let me know if it is fine?

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,27 @@ function backup_and_restore() {
...
+
+function wait_for_process_to_start() {
+  local pid=$1
+  vlog "wait for $pid to start "
+  while  ! kill -0 $pid >/dev/null 2>&1
+  do
+    vlog "waiting for pid $pid to start"
+    sleep 1
+  done
+}
```
satya-bodapati: I don't think this loop is necessary. Check existing kill_long_selects.sh which use "wait $pid". There is an existing function. Please use that and move to .inc file

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: sorry I didn't understand . can you please tell me exact function?

satya-bodapati: bg_wait_ok

rahulmalik87: IIUC bg_wait_ok is just wait untill the process is finish. but it is the background process and we wait  until it is started.

satya-bodapati: Hmm, if you have pid, then process is already started right? why do u need this hack?

storage/innobase/xtrabackup/test/inc/common.sh
altmannmarcelo: @rahulmalik87 same as on the azure PR. I think that a kill -9 on mysql client pid will do the trick, have you tested it ?

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -0,0 +1,88 @@
...
+password_string=secretpwd
+
+MYSQLD_EXTRA_MY_CNF_OPTS="
+secure-file-priv=$TEST_VAR_ROOT
+"
+vlog "Preparing server"
+start_server
+mysql -e "CREATE USER pxb@localhost IDENTIFIED BY '$password_string'"
+mysql -e "GRANT ALL ON *.* TO pxb@localhost "
+mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
```
satya-bodapati, line 17: explain why we need myisam engine table. What if someone tries to change to innodb.
Also we might need a relook when myisam gets removed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -0,0 +1,88 @@
...
+mysql -e "INSERT INTO t1 values(1)" test
+mkdir -p $topdir/backup
+
+#pass option like password -p to be run with
+
+function grep_in_datadir_ps() {
+
+  additional_options=$1
+  mysql -e "SELECT SLEEP(1000) from t1 for update " test &
```
satya-bodapati, line 29: why is this necessary? how does it help.  Please add comments

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,29 @@ function backup_and_restore() {
...
+  local condition=$1
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where $condition into outfile '$MYSQLD_TMPDIR/killcondition.sql';
+  source $MYSQLD_TMPDIR/killcondition.sql;
+EOF
+ rm -f $MYSQLD_TMPDIR/killcondition.sql
+}
+
+function wait_for_file_to_generated() {
```
satya-bodapati, line 1169: add a comment here too, how this wait for the file to be generated is useful?

satya-bodapati (changes requested):
mostly looks good rahul. Some documentation requests. Now that the testcase is changed, I want to see if this passes on both debug and release builds.

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,33 @@ function backup_and_restore() {
...
+## example kill_condition_queries "time > 0 & INFO like '%SLEEP%'"
+function kill_condition_queries()
+{
+  local condition=$1
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where $condition into outfile '$MYSQLD_TMPDIR/killcondition.sql';
+  source $MYSQLD_TMPDIR/killcondition.sql;
+EOF
+ rm -f $MYSQLD_TMPDIR/killcondition.sql
```
satya-bodapati: EOF and rm -rf are wrongly aligned

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,33 @@ function backup_and_restore() {
   verify_db_state $db
 }
 
+### kill all running queries accept the condition argument
+## example kill_condition_queries "time > 0 & INFO like '%SLEEP%'"
+function kill_condition_queries()
```
satya-bodapati: rename to kill_query_pattern() , you can tell the pattern is passed as argument or WHERE condition

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1150,5 +1150,33 @@ function backup_and_restore() {
...
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where $condition into outfile '$MYSQLD_TMPDIR/killcondition.sql';
+  source $MYSQLD_TMPDIR/killcondition.sql;
+EOF
+ rm -f $MYSQLD_TMPDIR/killcondition.sql
+}
+
+### wait for file to be generated.
+### start some process and wait untill that proceses generate the file
```
satya-bodapati: this can be used to check if  backup has started or not

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -0,0 +1,93 @@
...
+mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
+mysql -e "INSERT INTO t1 values(1)" test
+mkdir -p $topdir/backup
+
+#pass option like password -p to be run with
+
+function grep_in_datadir_ps() {
+
+  additional_options=$1
+  ### This SLEEP with update on MyISAM will let backup wait untill this query is killed. PXB waits to acquire FLUSH TABLES on MyISAM tables
```
satya-bodapati: The running update on myisam table will hold access to the table. PXB will do FLUSH TABLES on MyISAM tables which requires access to the table and waits for the running update on MyISAM to finish.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: EOF should come without any spaces. rm -rf is aligned to 2 spaces it it inside function.

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/inc/common.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: fixed

rahulmalik87:
./run.sh -t t/check_for_passwords.sh
Detecting server version...
Running against MySQL 5.7.37-debug-log (InnoDB 5.7.37)
Autodetected number of cores: 24
Using 16 parallel workers

check_for_passwords                      w1	[passed]    27
Spent 27 of 34 seconds executing testcases

satya-bodapati (approved):

satya-bodapati:
@rahulmalik87 please squash

altmannmarcelo (approved):
LGTM. Thanks for addressing all the comments.

storage/innobase/xtrabackup/test/inc/common.sh
satya-bodapati: @rahulmalik87 please test. kill -9 should do imho

satya-bodapati (approved):

PR #1279, opened 2022-03-15
Fixed PXB-2715 validate decompression path

altmannmarcelo:
Problem:

Fix:
Quote those path's in order to make it more secure.


altmannmarcelo:
FYI: This PR is the same as https://github.com/percona/percona-xtrabackup/pull/1263 - I've adjusted the Branch name to reflect PXB JIRA issue rather than Help Desk Ticket.

satya-bodapati (approved):

PR #1280, opened 2022-03-15
8.0 Fixed PXB-2715 validate decompression path

altmannmarcelo:
Problem:

Fix:
escape all special characters within the file name and quote the file
before invoking qpress command.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -2219,10 +2219,16 @@ bool
 decrypt_decompress_file(const char *filepath, uint thread_n)
 {
 	std::stringstream cmd, message;
-	char *dest_filepath = strdup(filepath);
+	char buf[FN_LEN];
 	bool needs_action = false;
 
-	cmd << "cat " << filepath;
+        if (escape_string_for_mysql(&my_charset_utf8mb4_general_ci, buf, 0,
+                                    filepath, strlen(filepath)) == (size_t)-1) {
+          msg_ts("[%02u] Error escaping file: %s\n", thread_n, filepath);
+          return false;
+        }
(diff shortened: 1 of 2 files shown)


storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2570,10 +2570,16 @@ bool copy_back(int argc, char **argv) {
 
 bool decrypt_decompress_file(const char *filepath, uint thread_n) {
   std::stringstream cmd, message;
-  char *dest_filepath = strdup(filepath);
+  char buf[FN_LEN];
   bool needs_action = false;
 
-  cmd << "cat " << filepath;
+  if (escape_string_for_mysql(&my_charset_utf8mb4_general_ci, buf, 0, filepath,
+                              strlen(filepath)) == (size_t)-1) {
```
satya-bodapati, line 2577: I think the commit message doesn't talk about this change. Why do we use it? it is escaping the new line characters?
Can you please update commit message? and a comment here too

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: Hi @satya-bodapati .
The fix is about escaping the filepath with single quotes via SQUOTE, however, if the file name has quotes in it we can still break out the command and inject code. escape_string_for_mysql is meant to scape special characters (include quotes) within the file name.

I will adjust the 2.4 PR and re-import it to 8.0

satya-bodapati: Thanks marce, please rerequest review after PR update

satya-bodapati (approved):
Please merge after circle CI result

PR #1284, opened 2022-03-28
PS-2723 xbcloud Azure password leaked in process status

rahulmalik87:
https://jira.percona.com/browse/PXB-2723

Hide xbcloud password from the ps output. 
Also backported pxb-2112 test case into 2.4

### DIFF
--- a/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
+++ b/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
@@ -12,7 +12,8 @@ secure-file-priv=$TEST_VAR_ROOT
 "
 vlog "Preparing server"
 start_server
-mysql -e "CREATE USER pxb@localhost IDENTIFIED BY '$password_string'" mysql -e "GRANT ALL ON *.* TO pxb@localhost"
+mysql -e "CREATE USER pxb@localhost IDENTIFIED BY '$password_string'"
+mysql -e "GRANT ALL ON *.* TO pxb@localhost"
 ### create table in MyISAM to let backup wait until there is pending update
 mysql -e "CREATE TABLE t1(i int) engine=MYISAM" test
 mysql -e "INSERT INTO t1 values(1)" test
@@ -28,19 +29,24 @@ function grep_in_datadir_ps() {
   ### to the table and waits for the running update on MyISAM to finish.
   mysql -e "SELECT SLEEP(1000) from t1 for update " test &
 
-  $XB_BIN $XB_ARGS --backup -u pxb $additional_options\
+  while ! mysql -e 'SHOW PROCESSLIST' | grep 'SELECT SLEEP' ; do
+    sleep 1;
+  done
+
+  $XB_BIN $XB_ARGS --backup -u pxb $additional_options \
   --transition-key=$password_string --target-dir=$topdir/backup \
-  2>&1 | tee $topdir/pxb.log &
+       2> >( tee $topdir/pxb_2722.log)&
   job_pid=$!
+  wait_for_file_to_generated $topdir/pxb_2722.log
 
   echo "backup pid is $job_pid"
 
-  file=$topdir/backup/xtrabackup_logfile
-  if [[ $additional_options == *"encrypt"* ]]; then
-	  file+=".xbcrypt"
-  fi
-  echo "looking for $file"
-  wait_for_file_to_generated $file
+  #sleep is to ensure password string is removed from ps
+  while ! grep -q 'Connecting to MySQL server host' $topdir/pxb_2722.log
+  do
+	  sleep 1;
+  done
+
   vlog "check for $password_string string in ps with $job_pid "
   if ps -ef | grep $job_pid | grep "xtrabackup.*$password_string" | grep -v grep
   then
@@ -49,15 +55,15 @@ function grep_in_datadir_ps() {
 
   ###kill running mysql query on MyISAM table to let backup resume
   vlog "killing mysql query"
-  kill_query_pattern "time > 0 & INFO like '%SLEEP%'"
+  kill_query_pattern "time > 0 && INFO like '%SLEEP%'"
 
   run_cmd wait $job_pid
 
   vlog "check for $password_string string in directory $topdir"
-  if grep -rq $password_string $topdir
+  if grep -rq $password_string $topdir/backup
   then
-     grep -r $password_string $topdir
-     die "found $password_string string in $topdir "
+     grep -r $password_string $topdir/backup
+     die "found $password_string string in $topdir/backup "
   fi
   rm -rf $topdir/backup
 }
--- a/storage/innobase/xtrabackup/test/t/xbcloud.sh
+++ b/storage/innobase/xtrabackup/test/t/xbcloud.sh
@@ -15,6 +15,9 @@
 
 . inc/common.sh
 
+MYSQLD_EXTRA_MY_CNF_OPTS="
+secure-file-priv=$TEST_VAR_ROOT
+"
 # run this test only when backing up MySQL 5.7+
 # to reduce the load on swift server
 require_server_version_higher_than 5.7.0
@@ -32,13 +35,14 @@ echo ${XBCLOUD_CREDENTIALS} | sed 's/ *--/\'$'\n/g' >> $topdir/xbcloud.cnf
 load_dbase_schema sakila
 load_dbase_data sakila
 
-mysql -e "CREATE database emptydatabas"
+mysql -e "ALTER TABLE payment COMPRESSION='lz4'" sakila
+mysql -e "CREATE database emptydatabase"
 
 now=$(date +%s)
 pwdpart=($(pwd | md5sum))
-
 full_backup_name=${now}-${pwdpart}-full_backup
 inc_backup_name=${now}-${pwdpart}-inc_backup
+storage_class_folder=${now}-${pwdpart}-storage_class
 
 full_backup_dir=$topdir/${full_backup_name}
 inc_backup_dir=$topdir/${inc_backup_name}
@@ -113,23 +117,23 @@ if ! grep -q failed $topdir/pxb-1832.log ; then
 fi
 
 #PXB-2164 xbcloud doesn't return the error if the backup doesn't exist in s3 bucket
-xbcloud --defaults-file=$topdir/xbcloud.cnf get somedummyjunkbackup 2>$topdir/pxb-2164.log
+run_cmd_expect_failure xbcloud --defaults-file=$topdir/xbcloud.cnf get somedummyjunkbackup 2>$topdir/pxb-2164.log
 
 if ! grep -q failed $topdir/pxb-2164.log ; then
     die 'xbcloud did not exit with error on get'
 fi
 
 #PXB-2198 xbcloud doesn't return the error on delete if the backup doesn't exist in s3 bucket
-xbcloud --defaults-file=$topdir/xbcloud.cnf delete somedummyjunkbackup 2>$topdir/pxb-2198.log
+run_cmd_expect_failure xbcloud --defaults-file=$topdir/xbcloud.cnf delete somedummyjunkbackup 2>$topdir/pxb-2198.log
 
 if ! grep -q "error: backup" $topdir/pxb-2198.log ; then
     die 'xbcloud did not exit with error on delete'
 fi
 
 #PXB-2202 Xbcloud does not display an error when xtrabackup fails to create a backup
-xtrabackup --backup --stream=xbstream --extra-lsndir=$full_backup_dir \
+run_cmd_expect_failure xtrabackup --backup --stream=xbstream --extra-lsndir=$full_backup_dir \
 	   --target-dir=/some/unknown/dir | \
-    run_cmd xbcloud --defaults-file=$topdir/xbcloud.cnf put \
+    run_cmd_expect_failure xbcloud --defaults-file=$topdir/xbcloud.cnf put \
 	    --parallel=4 \
 	    somedummyjunkbackup 2>$topdir/pxb-2202.log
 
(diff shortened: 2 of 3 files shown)


storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -32,13 +56,14 @@ echo ${XBCLOUD_CREDENTIALS} | sed 's/ *--/\'$'\n/g' >> $topdir/xbcloud.cnf
 load_dbase_schema sakila
 load_dbase_data sakila
 
-mysql -e "CREATE database emptydatabas"
+mysql -e "ALTER TABLE payment COMPRESSION='lz4'" sakila
```
satya-bodapati, line 38: why do we need this ALTER now?

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +190,105 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
     die 'xbcloud exit with error on upload'
 fi
 
+# PXB-2112 - Support for storage class
```
satya-bodapati, line 172: 2112??

satya-bodapati (changes requested):
Can you also paste test result from jenkins?

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: test case for PXB-2112 was not ported to 2.4 but it is present in 8.0. so just ported it back.

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: This changes exists in 8.0. so just back ported it. the test is successful

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -15,6 +15,30 @@
...
+
+function wait_for_process_to_start() {
+  local pid=$1
+  vlog "wait for $pid to start "
+  while  ! kill -0 $pid >/dev/null 2>&1
+  do
+    vlog "waiting for pid $pid to start"
+    sleep 1
+  done
+}
```
altmannmarcelo: rename this to something more clear like, `wait_for_pid` and move it to common.sh

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -15,6 +15,30 @@
...
+### kill all running queries
+function kill_sleep_queries()
+{
+  run_cmd $MYSQL $MYSQL_ARGS --force --batch  test <<EOF
+  select concat('KILL ',id,';') from information_schema.processlist
+  where user='root' and info like '%SLEEP%' and  time > 1 into outfile '$MYSQLD_TMPDIR/killall.sql';
+  source $MYSQLD_TMPDIR/killall.sql;
+EOF
+  rm -f $MYSQLD_TMPDIR/killall.sql
+}
```
altmannmarcelo: I would say we should make this more generic to allow change parameters such as user, info and time or just check if a `kill -9` on the MySQL client PID will stop the query.

altmannmarcelo (changes requested):
Do we have jenkins or local run results?

rahulmalik87:
> Do we have jenkins or local run results?

{qa-sm-twina02 x7 o7 xtrabackup-test}$./run.sh -t t/xbcloud.sh
Detecting server version...
Running against MySQL 5.7.37-debug-log (InnoDB 5.7.37)
Autodetected number of cores: 24
Using 16 parallel workers

xbcloud                                  w1	[passed]    171
Spent 171 of 179 seconds executing testcases

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: using existing function from common.sh

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +169,111 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
...
+
+xtrabackup --backup --stream=xbstream --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud $XBCLOUD_CREDENTIALS put \
+	    --parallel=4 \
+	    ${full_backup_name} &
+job_pid=$!
+
+#sleep is to ensure password is removed by xbcloud
+sleep 10
```
satya-bodapati: still required?

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -113,23 +117,23 @@ if ! grep -q failed $topdir/pxb-1832.log ; then
 fi
 
 #PXB-2164 xbcloud doesn't return the error if the backup doesn't exist in s3 bucket
-xbcloud --defaults-file=$topdir/xbcloud.cnf get somedummyjunkbackup 2>$topdir/pxb-2164.log
+run_cmd_expect_failure xbcloud --defaults-file=$topdir/xbcloud.cnf get somedummyjunkbackup 2>$topdir/pxb-2164.log
```
altmannmarcelo, line 120: has this changed on this PR?
The comment say's xbcloud does not return error, and we are executing `run_cmd_expect_failure`

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +169,111 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
...
+
+
+#check if password is stored in backup directory or in ps output
+
+## grep for all key in secret_key
+params=($(echo $XBCLOUD_CREDENTIALS | tr ";" "\n"))
+secret_keys=();
+for param in "${params[@]}"
+do
+ if [[ $param =~ "key=" ]]; then
```
altmannmarcelo, line 226: Will this also catch `azure-storage-account` ?

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +169,111 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
...
+mysql -e "INSERT into t1 values(100)" test
+
+### The running update on myisam table will hold access to the table.
+### PXB will do FLUSH TABLES on MyISAM tables which requires access
+### to the table and waits for the running update on MyISAM to finish.
+mysql -e "SELECT SLEEP(1000) from t1 for update " test &
+
+xtrabackup --backup --stream=xbstream --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    run_cmd xbcloud $XBCLOUD_CREDENTIALS put \
```
altmannmarcelo: do we need run_cmd here ?

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +169,111 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
...
+	    --parallel=4 \
+	    ${full_backup_name} &
+job_pid=$!
```
altmannmarcelo, line 252: we are using pipe here, what are we running in background ? pxb + xbcloud or only xbcloud?
You 100% sure that $! give the pid of xbcloud? like for return code we have PIPESTATUS[0] and PIPESTATUS[1] - do we have something similar to catch the PID?

storage/innobase/xtrabackup/test/t/xbcloud.sh
altmannmarcelo: Is this still necessary? can we perhaps grep the log for an indication that the backup has started/passed the argument handling part?

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: The comment is about jira ticket so may be misleading,  It should fail and return error. The problem is seen only with macos when xbcloud fails the testcase stops. so added run_cmd_expect

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: no. it will just grep keys. let me fixed that.

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: both are in background. I tested the test case by commenting AZURE_SECRET_KEY for hide option and test failed. so no worries. 

for example $! is 8068 and below process contains 8068

 ps -ef |grep 8068
  501  8068  7343   0 10:47AM ttys001    0:00.00 /bin/bash ./run.sh -t t/xbcloud.sh
  501  8084  8068   0 10:47AM ttys001    0:00.63 xbcloud --storage=s3 --s3-access-key=
  501  8534  8512   0 10:48AM ttys004    0:00.00 grep 8068

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: Which logs to grep? with backup format being xbstream it won't logged anything on disk.

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: just to ensure it doesn't fail and go unnoticed.

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: It is not generating any file. output into xbstream. so ensure password is removed by ps

storage/innobase/xtrabackup/test/t/xbcloud.sh
altmannmarcelo: Can you please make sure it works on other platforms too?
From `xbcloud` to `run_cmd_expect_failure` we are changing from expecting success vs error. I remember centos7 been more sensitive for those things.

satya-bodapati: I think previously the expected error wasn't captured/verified

storage/innobase/xtrabackup/test/t/xbcloud.sh
altmannmarcelo: hmmm, here job_pid is actually MTR (run.sh) and not xbcloud .
Also, xtrabackup is not running in background. Can you test with `(xtrabackup | xbcloud) &` and send here the output of `ps -ef | grep pid`

storage/innobase/xtrabackup/test/t/xbcloud.sh
altmannmarcelo: log messages are sent to stderr, redirects ( | or >) will redirect only STDOUT. We are still writing stderr to `$OUTFILE`

storage/innobase/xtrabackup/test/t/xbcloud.sh
satya-bodapati: Rahul, which pid will be captured in job_pid? Later we do "run_cmd wait $job_pid" so what do we wait for to finish?

rahulmalik87: some new learning for me. If I don't use run_cmd and uses only xbcloud job_pid captures the xbcloud pid.  else It will store pid of main script. so removed the run_cmd.
fixed

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: with PS. Update on MYISAM blocks LTFB.  not sure what to check if xbcloud has already parsed the value or not. 

error: http request failed: Couldn't resolve host name
error: http request failed: Couldn't resolve host name
220406 14:47:49 xbcloud: Successfully connected.
2022-04-06T14:47:49.315246+05:30 0 [Note] [MY-011825] [Xtrabackup] recognized client arguments: --target-dir=/Users/rahulmalik/MySQL/data/xp8 --core-file --user=root --socket=/tmp/socket_3208.sock --keyring-file-data=/Users/rahulmalik/MySQL/data/p8/key.key --xtrabackup-plugin-dir=/Users/rahulmalik/MySQL/src/x8/bld/plugin_output_directory --backup=1 --stream=xbstream
/Users/rahulmalik/MySQL/src/x8/bld/runtime_output_directory/xtrabackup version 8.0.28-20 based on MySQL server 8.0.28 macos12.2 (x86_64) (revision id: b980f66bc03)
220406 14:47:49  version_check Connecting to MySQL server with DSN 'dbi:mysql:;mysql_read_default_group=xtrabackup;mysql_socket=/tmp/socket_3208.sock' as 'root'  (using password: NO).
Failed to connect to MySQL server as DBD::mysql module is not installed at - line 1548.
2022-04-06T14:47:49.442151+05:30 0 [Note] [MY-011825] [Xtrabackup] Connecting to MySQL server host: localhost, user: root, password: not set, port: not set, socket: /tmp/socket_3208.sock
2022-04-06T14:47:49.454317+05:30 0 [Note] [MY-011825] [Xtrabackup] Using server version 8.0.28-19-debug
2022-04-06T14:47:49.488822+05:30 0 [Note] [MY-011825] [Xtrabackup] Executing LOCK TABLES FOR BACKUP ...

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -28,7 +28,8 @@ function grep_in_datadir_ps() {
   ### to the table and waits for the running update on MyISAM to finish.
   mysql -e "SELECT SLEEP(1000) from t1 for update " test &
 
-  $XB_BIN $XB_ARGS --backup -u pxb $additional_options\
+  ### lock-ddl=false is required since LTFB will wait for update on MyISAM table to finish
+  $XB_BIN $XB_ARGS --backup -u pxb $additional_options --lock-ddl=false \
```
rahulmalik87: @satya-bodapati @altmannmarcelo , I had to use one more hack to make this test release build. for PS LTFB is acquired before PXB generates xtrabackup_log file. 
Let me know if you fine with it else I will make this test debug only.

satya-bodapati: isn't the intention for backup to wait until the query is killed?

rahulmalik87: yes the intention is to let backup wait.  Backup will still wait with --lock-ddl=false. It just it won't acquire LTFB at early stage.

satya-bodapati: does it (LTFB) happens twice? from where are these request for locks initiated?

storage/innobase/xtrabackup/test/t/xbcloud.sh
altmannmarcelo: > 220406 14:47:49 xbcloud: Successfully connected.

This is an indication that we have parsed everything and managed to connect to backend storage and are ready to upload. At this point no sensitive data should be exposed at `ps -ef`

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: yes for for PS it happens twice. at the end instead of flush tables for MyISAM it do LTFB.

storage/innobase/xtrabackup/test/t/xbcloud.sh
rahulmalik87: let me try the backgroup process as well missed that comment will update you.

storage/innobase/xtrabackup/test/t/xbcloud.sh
```diff
@@ -165,8 +169,124 @@ if  grep -q "Upload failed" $topdir/pxb-2279.log ; then
...
+
+### The running update on myisam table will hold access to the table.
+### PXB will do FLUSH TABLES on MyISAM tables which requires access
+### to the table and waits for the running update on MyISAM to finish.
+mysql -e "SELECT SLEEP(1000) from t1 for update " test &
+
+xtrabackup --backup --stream=xbstream --extra-lsndir=$full_backup_dir \
+	   --target-dir=$full_backup_dir | \
+    xbcloud $XBCLOUD_CREDENTIALS put --parallel=4 \
+     ${full_backup_name}  2> >( tee $topdir/pxb_2723.log)&
```
satya-bodapati, line 251: can we keep stdout redirection as well?

satya-bodapati (approved):
LGTM! Please test on single platform jenkins (because of new changes to testcase) and then merge

PR #1291, opened 2022-04-07
PXB-2761 Fix compilation error due to loop-variable creates a copy from its type

tcoyvwac:
https://jira.percona.com/browse/PXB-2761

* Fixes compilation error [in GCC 11+ because of new error flag: `-Wrange-loop-construct`, is now enabled in `-Wall`. ](https://gcc.gnu.org/gcc-11/changes.html#cxx)
* Compiler encouragingly recommends using a [`reference type to prevent copying`](https://gcc.gnu.org/onlinedocs/gcc-11.1.0/gcc/C_002b_002b-Dialect-Options.html#index-Wrange-loop-construct).

### DIFF
--- a/storage/innobase/xtrabackup/src/keyring_plugins.cc
+++ b/storage/innobase/xtrabackup/src/keyring_plugins.cc
@@ -758,7 +758,7 @@ bool xb_tablespace_keys_dump(ds_ctxt_t *ds_ctxt, const char *transition_key,
     }
   }
 
-  for (const auto entry : encryption_info) {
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
Hi @tcoyvwac thanks for your PR. it LGTM.

Would you mind sharing your full name so we can credit this contribution to you at our release notes?

altmannmarcelo (approved):

altmannmarcelo:
Issue tracker - https://jira.percona.com/browse/PXB-2761

PR #1296, opened 2022-04-13
PXB-8.0-2740 PXB displays the encryption key when the backup is decrypted…

rahulmalik87:
… and decompressed

https://jira.percona.com/browse/PXB-2740

Problem:
PXB displays the encryption key when the backup is decrypted and decompressed

Analysis:
PXB uses system function to call xbcrypt with encryption password to decrypt
the password. system function creates sub process. password is only
removed by sub process . The main process shows the encrypted password
in ps

Fix:
set the enviorment variable XBCRYPT_ENCRYPTION_PASSWORD in pxb and use
that variable inside xbcrypt

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -2189,9 +2189,7 @@ decrypt_decompress_file(const char *filepath, uint thread_n)
  	if (ends_with(filepath, ".xbcrypt") && opt_decrypt) {
  		cmd << " | xbcrypt --decrypt --encrypt-algo="
  		    << xtrabackup_encrypt_algo_names[opt_decrypt_algo];
- 		if (xtrabackup_encrypt_key) {
- 			cmd << " --encrypt-key=" << xtrabackup_encrypt_key;
- 		} else {
+ 		if (xtrabackup_encrypt_key == NULL) {
  			cmd << " --encrypt-key-file="
  			    << xtrabackup_encrypt_key_file;
  		}
@@ -2303,6 +2301,10 @@ decrypt_decompress()
 	it = datadir_iter_new(".", false);
 
 	ut_a(xtrabackup_parallel >= 0);
+
+	if (xtrabackup_encrypt_key) {
+		setenv("XBCRYPT_ENCRYPTION_KEY", xtrabackup_encrypt_key, 1);
+	}
 
 	ret = run_data_threads(it, decrypt_decompress_thread_func,
 		xtrabackup_parallel ? xtrabackup_parallel : 1,
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -67,6 +67,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "utils.h"
 #include "xb0xb.h"
 
+#include <cstdlib>
 #include "backup_copy.h"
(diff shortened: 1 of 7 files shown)


storage/innobase/xtrabackup/src/xbcrypt_common.cc
```diff
@@ -305,3 +305,10 @@ gcry_error_t xb_crypt_encrypt(gcry_cipher_hd_t cipher_handle, const uchar *from,
 
   return 0;
 }
+
+void get_env_value(char *&var, const char *env) {
+  char *val;
+  if (var == nullptr && (val = getenv(env)) != nullptr) {
+    var = my_strdup(PSI_NOT_INSTRUMENTED, val, MYF(MY_FAE));
```
satya-bodapati, line 312: where do we free this?

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2584,9 +2585,7 @@ bool decrypt_decompress_file(const char *filepath, uint thread_n) {
   if (ends_with(filepath, ".xbcrypt") && opt_decrypt) {
     cmd << " | xbcrypt --decrypt --encrypt-algo="
         << xtrabackup_encrypt_algo_names[opt_decrypt_algo];
-    if (xtrabackup_encrypt_key) {
-      cmd << " --encrypt-key=" << xtrabackup_encrypt_key;
-    } else {
+    if (xtrabackup_encrypt_key == nullptr) {
```
satya-bodapati, line 2588: can we do setenv() here?
if (xtrabckup_encrypt_key) {
setenv();
} else {
--file
}

satya-bodapati, line 2588: also, I think it is better to unset it after usage.
get old value (getnv()), save it. and restore it after system() call.

satya-bodapati (changes requested):
also better to do GCA and make PR for 2.4

storage/innobase/xtrabackup/src/backup_copy.cc
satya-bodapati: on second thought, I think, the ENV change done by this xtrabackup process will not be visible to other xtrabackup processes. So pxb exit will auto unset it.

Please verify

rahulmalik87: @satya-bodapati , if we do setenv inside decrypt_decompress_file we have to do it for each file. better to do it outside of this function so global environment will be called once .

rahulmalik87: correct, pxb exists with auto unset.

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2655,6 +2654,10 @@ static void decrypt_decompress_thread_func(datadir_thread_ctxt_t *ctxt) {
       continue;
     }
 
+    if (xtrabackup_encrypt_key) {
+      setenv("XBCRYPT_ENCRYPTION_KEY", xtrabackup_encrypt_key, 1);
+    }
+
```
satya-bodapati: here too, we seem to set after every file, this should be set before the while loop isn't it?

rahulmalik87:
@satya-bodapati  test results.



check_for_passwords                      w1	[passed]    60
Spent 60 of 86 seconds executing testcases

satya-bodapati (approved):

PR #1308, opened 2022-04-26
PXB-2729 PXB compilation fails if GTest packages are installed

rahulmalik87:
https://jira.percona.com/browse/PXB-2729

Fixed the compilation error and xbcloud-t.cc test case

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/CMakeLists.txt
+++ b/storage/innobase/xtrabackup/src/xbcloud/CMakeLists.txt
@@ -71,7 +71,7 @@ TARGET_LINK_LIBRARIES(xbcloud
   crc
   )
 
-IF (WITH_DEBUG)
+IF (WITH_UNIT_TESTS AND CMAKE_VERSION VERSION_GREATER 3.22.1)
 
   ENABLE_TESTING()
 
@@ -87,6 +87,7 @@ IF (WITH_DEBUG)
       ../xbstream_read.c
       http.cc
       s3.cc
+      azure.cc
       swift.cc)
 
     TARGET_LINK_LIBRARIES(xbcloud-t
@@ -94,7 +95,8 @@ IF (WITH_DEBUG)
       ${LIBEV_LIBRARIES}
       ${CURL_LIBRARIES}
       ${GTEST_BOTH_LIBRARIES}
-      gmock
+      GTest::gtest
+      GTest::gmock
       mysys
       mysys_ssl
       crc
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
@@ -1,6 +1,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "azure.h"
 #include "http.h"
 #include "s3.h"
 #include "swift.h"
@@ -38,7 +39,8 @@ MATCHER_P(PayloadEq, payload, "") {
 
 TEST(s3_client, basicDNSv4) {
   Mock_http_client http_client;
-  S3_client c(&http_client, "us-east-1", "my-access-key-id", "my-secret-key");
+  S3_client c(&http_client, "us-east-1", "my-access-key-id", "my-secret-key", 1,
+              1);
   EXPECT_CALL(
       http_client,
       make_request(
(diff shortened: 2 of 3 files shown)


storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
```diff
@@ -226,14 +229,14 @@ TEST(s3v4_signer, storageClass) {
...
-               "891034a3bd13729689a54d363380ad1849b26bf2b9e461d4c2bdeeca32e0c1c"
-               "e");
+  ASSERT_STREQ(
+      req.headers().at("Authorization").c_str(),
+      "AWS4-HMAC-SHA256 "
+      "Credential=access_key/20190422/example-region/s3/aws4_request, "
+      "SignedHeaders=content-length;content-type;host;x-amz-content-"
+      "sha256;x-amz-date;x-amz-security-token;x-amz-storage-class, "
+      "Signature="
+      "fe6c888f22fb23a7a3fe6f663013b5df3cc761c3777ed4368b325114c885320f");
```
satya-bodapati, line 241: how did you get new signature?

storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
```diff
@@ -77,7 +78,8 @@ TEST(s3_client, basicDNSv4) {
 
 TEST(s3_client, basicPATHv4) {
```
satya-bodapati, line 80: Should we add similar test for azure APIs?

storage/innobase/xtrabackup/src/xbcloud/CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ TARGET_LINK_LIBRARIES(xbcloud
   crc
   )
 
-IF (WITH_DEBUG)
+IF (WITH_DEBUG AND CMAKE_VERSION VERSION_GREATER 3.22.1)
```
satya-bodapati: any limitation with cmake < 3.22.1?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbcloud/CMakeLists.txt
rahulmalik87: yes. GTest::gmock was introduce after 3.22.1 . 

https://cmake.org/cmake/help/v3.23/module/FindGTest.html.

storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
rahulmalik87: the test was failing and copied the new signature.

altmannmarcelo (changes requested):
@rahulmalik87 I think we should not enforce the tests on WITH_DEBUG builds,

We should instead change this flag with WITH_UNIT_TEST and allow it to be turned on/off at will.

Maybe default to OFF on RelWithDebInfo and default to ON on Debug, but with the option to flip it.

rahulmalik87:
> @rahulmalik87 I think we should not enforce the tests on WITH_DEBUG builds,
> 
> We should instead change this flag with WITH_UNIT_TEST and allow it to be turned on/off at will.
> 
> Maybe default to OFF on RelWithDebInfo and default to ON on Debug, but with the option to flip it.

Do you mean adding new CMAKE flag or using existing CMAKE that mysql uses . There is compilation error in using existing CMAKE with_unit_test flag for PXB

altmannmarcelo:
> > @rahulmalik87 I think we should not enforce the tests on WITH_DEBUG builds,
> > We should instead change this flag with WITH_UNIT_TEST and allow it to be turned on/off at will.
> > Maybe default to OFF on RelWithDebInfo and default to ON on Debug, but with the option to flip it.
> 
> Do you mean adding new CMAKE flag or using existing CMAKE that mysql uses . There is compilation error in using existing CMAKE with_unit_test flag for PXB

@rahulmalik87 I mean using the existing one. On this issue, we found out that there is no way to compile a Debug version but disable Gunit tests if you have Gunit library installed. We should give the ability to disable its compilation.

rahulmalik87:
> > > @rahulmalik87 I think we should not enforce the tests on WITH_DEBUG builds,
> > > We should instead change this flag with WITH_UNIT_TEST and allow it to be turned on/off at will.
> > > Maybe default to OFF on RelWithDebInfo and default to ON on Debug, but with the option to flip it.
> > 
> > 
> > Do you mean adding new CMAKE flag or using existing CMAKE that mysql uses . There is compilation error in using existing CMAKE with_unit_test flag for PXB
> 
> @rahulmalik87 I mean using the existing one. On this issue, we found out that there is no way to compile a Debug version but disable Gunit tests if you have Gunit library installed. We should give the ability to disable its compilation.

It only compiles with WITH_UNIT_TESTS

storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
rahulmalik87: added

satya-bodapati (approved):
plz squash before merge

altmannmarcelo (approved):
LGTM

PR #1338, opened 2022-05-26
PXB 8.0.29 merge

altmannmarcelo:
https://jira.percona.com/browse/PXB-2755


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7092,6 +7246,22 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
 
     srv_shutdown_state = SRV_SHUTDOWN_NONE;
 
+    /* PXB-2792 - Ensure we don't have ib_logfile's */
```
rahulmalik87: It is better to create seprate commit  for this changes as we have jira ticket also

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -2,7 +2,11 @@ set -eu
 
 function xtrabackup()
 {
+  if [ -z ${WITH_RR+x} ]; then
```
rahulmalik87: Is this new option we are introducing? we should mention about it in run.sh --help

storage/innobase/xtrabackup/test/inc/keyring_kms.sh
```diff
@@ -51,16 +51,16 @@ function cleanup_keyring() {
 
 function ping_kms()
 {
-  if [ -z ${KMS_REGION+x} ]; then
+  if [ -z ${KMS_REGION} ]; then
```
rahulmalik87: is it to report failure when KMS_REGION is set to null? Also better to add quotes.

storage/innobase/dict/dict0dd.cc
```diff
@@ -6362,13 +6349,8 @@ const char *dd_process_dd_tables_rec_and_mtr_commit(
 
   /* Get the se_private_id field. */
   field = (const byte *)rec_get_nth_field(
-<<<<<<< HEAD
-      rec, offsets, dd_tables->field_number("se_private_id") + DD_FIELD_OFFSET,
-=======
       nullptr, rec, offsets,
-      dd_object_table.field_number("FIELD_SE_PRIVATE_ID") + DD_FIELD_OFFSET,
->>>>>>> mysql-8.0.29
```
satya-bodapati: Do you know why we changed from "FIELD_SE_PRIVATE_ID" to se_private_id?

storage/innobase/include/fsp0fsp.ic
```diff
@@ -366,11 +366,6 @@ inline bool fsp_flags_is_valid(uint32_t flags) {
     return (false);
   }
 
-<<<<<<< HEAD
-#if FSP_FLAGS_POS_UNUSED != 15
-#error You have added a new FSP_FLAG without adding a validation check.
-#endif
```
satya-bodapati: This seems like a valid check. Is it removed by upstream? Did upstream add a new FSP flag?

include/boost_1_77_0/patches/boost/container/throw_exception.hpp
```diff
@@ -0,0 +1,289 @@
...
+#endif
+
+#if defined(BOOST_HAS_PRAGMA_ONCE)
+#pragma once
+#endif
+
+#include <boost/container/detail/config_begin.hpp>
+#include <boost/container/detail/workaround.hpp>
+#include <boost/core/ignore_unused.hpp>
+
```
satya-bodapati: why is this required by us (PXB) and not upstream?

satya-bodapati: this as in, the entire file

sql/join_optimizer/join_optimizer.cc
```diff
@@ -3449,8 +3452,11 @@ void CostingReceiver::ProposeNestedLoopJoin(
       const CachedPropertiesForPredicate &properties =
           edge->expr->properties_for_equijoin_conditions[join_cond_idx];
 
+      MY_COMPILER_DIAGNOSTIC_PUSH()
+      MY_COMPILER_GCC_DIAGNOSTIC_IGNORE("-Wunused-variable")
       const auto [already_applied_as_sargable, subsumed] =
           AlreadyAppliedAsSargable(condition, left_path, right_path);
+      MY_COMPILER_DIAGNOSTIC_POP()
```
satya-bodapati: this is upstream code and it didn't catch this error? :thinking:

storage/innobase/dict/dict0dd.cc
```diff
@@ -442,10 +444,10 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
                                         const dd::String_type *schema_name,
                                         bool is_implicit, space_id_t space_id) {
   using namespace dict_name;
-  mem_heap_t *heap = mem_heap_create(1000);
+  mem_heap_t *heap = mem_heap_create(1000, UT_LOCATION_HERE);
 
   char table_name[MAX_SPACE_NAME_LEN + 1];
-  char tmp_schema[MAX_DATABASE_NAME_LEN + 1];
+  char tmp_schema[+1];
```
satya-bodapati: is this typo?

storage/innobase/include/fsp0fsp.ic
```diff
@@ -321,6 +321,7 @@ inline bool fsp_flags_is_valid(uint32_t flags) {
   bool has_data_dir = FSP_FLAGS_HAS_DATA_DIR(flags);
   bool is_shared = FSP_FLAGS_GET_SHARED(flags);
   bool is_temp = FSP_FLAGS_GET_TEMPORARY(flags);
+  bool is_encryption = FSP_FLAGS_GET_ENCRYPTION(flags);
```
satya-bodapati: before this change, how did this function compile without the variable `is_encryption`?

storage/innobase/include/log0recv.h
```diff
@@ -742,9 +742,8 @@ void recv_reset_buffer();
 bool recv_sys_resize_buf();
 
 /** Parse log records from a buffer and optionally store them to a
-hash table to wait merging to file pages.
-@param[in]  checkpoint_lsn  the LSN of the latest checkpoint */
-void recv_parse_log_recs(lsn_t checkpoint_lsn);
+hash table to wait merging to file pages. */
+void recv_parse_log_recs();
```
satya-bodapati: are we now accessing checkpoint_lsn from global recv_sys data structure?

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -477,7 +477,7 @@ bool backup_file_print(const char *filename, const char *message, int len) {
   MY_STAT stat;
 
   memset(&stat, 0, sizeof(stat));
-  stat.st_mtime = my_time(0);
+  stat.st_mtime = time(nullptr);
```
satya-bodapati: my_time() vs time(nullptr) ?

satya-bodapati (changes requested):

storage/innobase/dict/dict0dd.cc
```diff
@@ -447,7 +447,7 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
   mem_heap_t *heap = mem_heap_create(1000, UT_LOCATION_HERE);
 
   char table_name[MAX_SPACE_NAME_LEN + 1];
-  char tmp_schema[+1];
+  char tmp_schema[MAX_DATABASE_NAME_LEN + 1];
```
satya-bodapati: oh, you fixed here? I dropped a comment before. Plz ignore then ;)

storage/innobase/xtrabackup/test/inc/keyring_kms.sh
altmannmarcelo: Yes, this is to skip the test if we have not SET the variables.

storage/innobase/dict/dict0dd.cc
altmannmarcelo: @satya-bodapati We/upstream did not change.
PXB gather the information as `dd_tables->field_number("se_private_id")`
Upstream create a `dd_object_table` - https://github.com/mysql/mysql-server/blob/mysql-8.0.29/storage/innobase/dict/dict0dd.cc#L5332 and uses that object to fetch the field number via `dd_object_table.field_number("FIELD_SE_PRIVATE_ID")` . From what I can see, we first added the optimization of fetching the field number as name rather than fixed number. Later upstream did the same in a different way but we never conciliated to their way. 
However, here what changed was the first parameter `nullptr` we pass to `rec_get_nth_field` as this is new work added as part of INSTANT ADD/DROP column commit.

storage/innobase/include/fsp0fsp.ic
altmannmarcelo: This comes from https://github.com/altmannmarcelo/percona-xtrabackup/commit/e001f2a533c4b8edd26e8e1723f00ed4a71d80c7 They removed `#defines` from code.
This particular part change to a `static_assert` - https://github.com/mysql/mysql-server/blob/mysql-8.0.29/storage/innobase/include/fsp0fsp.ic#L376

include/boost_1_77_0/patches/boost/container/throw_exception.hpp
altmannmarcelo: Seems like we do not need this after all. It might have been fixed by other later change. Will remove it

storage/innobase/include/fsp0fsp.ic
altmannmarcelo: This has required as https://github.com/mysql/mysql-server/blob/mysql-8.0.29/storage/innobase/include/fsp0fsp.ic#L376 brought over:
```
  /* Only single-table and general tablespaces and not temp tablespaces
  use the encryption clause. */
  if (is_encryption && (is_temp)) {
    return (false);
  }
```
To be on the safe-side, since we do support encryption of temp table, I've kept that if and the is_encryption flag check under `ifndef XTRABACKUP`

storage/innobase/include/log0recv.h
altmannmarcelo: This comes from https://github.com/mysql/mysql-server/commit/85f9abf99075052964d6434737a7645c311b2990#diff-26f5d4b9822d8a85edd9ea715cde19e90347775be561b8919601cfcb3df8cb49 . We were referencing checkpoint_lsn as a parameter but not using it.

From what I can see, we only used it on 5.7 and just to print a debug message - https://github.com/mysql/mysql-server/blob/5.7/storage/innobase/log/log0recv.cc#L3247

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: yes, this came from https://github.com/mysql/mysql-server/commit/7230b2e10e89e8df920f83fc11f12b97937b3f26
> Change all calls my_time(0) to ::time(nullptr).

sql/join_optimizer/join_optimizer.cc
altmannmarcelo: yes, this error is thrown if compiling with GCC-7

storage/innobase/xtrabackup/test/inc/common.sh
altmannmarcelo: Done

include/boost_1_77_0/patches/boost/container/throw_exception.hpp
satya-bodapati: Thanks

storage/innobase/include/fsp0fsp.ic
satya-bodapati: ok, if there is static_assert, then we are all good

storage/innobase/dict/dict0dd.cc
satya-bodapati: Thanks for clarification.

satya-bodapati:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/198/CMAKE_BUILD_TYPE=Debug,DOCKER_OS=ubuntu%3Afocal,XTRABACKUP_TARGET=innodb80/
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/198/CMAKE_BUILD_TYPE=Debug,DOCKER_OS=centos%3A7,XTRABACKUP_TARGET=innodb80/
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/198/CMAKE_BUILD_TYPE=Debug,DOCKER_OS=asan,XTRABACKUP_TARGET=innodb80/

satya-bodapati (approved):
@altmannmarcelo 
LGTM! I request to you test import export of old version datadir with newer 8.0.29 binaries.
I assume there is a new CFG version and PXB 8.0.29 should create an appropriate CFG version.

rahulmalik87 (approved):
LGTM

PR #1340, opened 2022-05-30
PXB-2759 regression on check_for_passwords.sh

rahulmalik87:
https://jira.percona.com/browse/PXB-2759

added sleep 3 befor checking output of ps

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -2639,6 +2639,8 @@ bool decrypt_decompress() {
   ret = run_data_threads(".", decrypt_decompress_thread_func,
                          xtrabackup_parallel, "decrypt and decompress");
 
+  debug_sync_point("decrypt_decompress_func");
+
   if (ds_data != NULL) {
     ds_destroy(ds_data);
   }
--- a/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
+++ b/storage/innobase/xtrabackup/test/t/check_for_passwords.sh
@@ -41,15 +41,17 @@ function grep_in_datadir_ps() {
   echo "backup pid is $job_pid"
 
   #sleep is to ensure password string is removed from ps
+  sleep 3
   while ! grep -q 'Connecting to MySQL server host' $topdir/pxb_2722.log
   do
 	  sleep 1;
   done
 
-  vlog "check for $password_string string in ps with $job_pid "
-  if ps -ef | grep $job_pid | grep "xtrabackup.*$password_string" | grep -v grep
-  then
-     die "found $password_string string in ps output"
+  vlog "check for $password_string in xtrabackup ps with $job_pid "
+  ps_output=$(ps -ef | grep $job_pid | grep xtrabackup | grep -v grep)
+  echo "ps output xtrabckup is ${ps_output}"
+  if [[ ${ps_output} == *${password_string}* ]];  then
+     die "found $password_string string in $ps_output"
   fi
 
   ###kill running mysql query on MyISAM table to let backup resume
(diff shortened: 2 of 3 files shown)


rahulmalik87:
jenkin results https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/182

rahulmalik87:
jenkin results. https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/201/

satya-bodapati (approved):
Please squash before the merge

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2639,6 +2639,8 @@ bool decrypt_decompress() {
   ret = run_data_threads(".", decrypt_decompress_thread_func,
                          xtrabackup_parallel, "decrypt and decompress");
 
+  debug_sync_point("decrypt_decompress_func");
```
altmannmarcelo, line 2642: Do we really need this one? checking for `decrypt` isn't enough?

storage/innobase/xtrabackup/test/t/check_for_passwords_debug.sh
```diff
@@ -0,0 +1,56 @@
...
+function grep_in_datadir_ps_decrypt() {
+
+  additional_options=$1
+
+  xtrabackup --backup $additional_options --target-dir=$topdir/backup
+
+  $XB_BIN $XB_ARGS --decrypt=AES256 $additional_options \
+  --target-dir=$topdir/backup \
+           --debug-sync="decrypt_decompress_func" \
+       2> >( tee $topdir/decompress.log)&
```
altmannmarcelo: decrypt.log

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
```diff
@@ -70,32 +72,15 @@ function grep_in_datadir_ps() {
...
-function grep_in_datadir_ps_decrypt() {
+function grep_in_datadir_decrypt() {
 
   additional_options=$1
 
   xtrabackup --backup $additional_options --target-dir=$topdir/backup
 
   $XB_BIN $XB_ARGS --decrypt=AES256 $additional_options \
   --target-dir=$topdir/backup \
-       2> >( tee $topdir/pxb_2740.log)&
```
altmannmarcelo, line 81: I would leave this part with extra information to validate why this is failing.

altmannmarcelo (changes requested):
@rahulmalik87 although the suggested change might work, I would investigate why this is failing without the debug sync.
Based on your changes, the issue comes from `decrypting` part. I would enhance debugging there to validate why it fails, Having the `decrypting` string on pxb output log should be enough to be passed the argument handling part, which should be stripping sensitive information.

rahulmalik87:
> @rahulmalik87 although the suggested change might work, I would investigate why this is failing without the debug sync.
> Based on your changes, the issue comes from `decrypting` part. I would enhance debugging there to validate why it fails, Having the `decrypting` string on pxb output log should be enough to be passed the argument handling part, which should be stripping sensitive information.

It fails because we don't have any mechanism to ensure decrypting is not finished at the time we are running ps command.
  
Having the `decrypting` string is enough but how we would ensure that decrypting is still running. That's why I thought to run only debug test case to 100% process would be running by the time we run ps -ef

storage/innobase/xtrabackup/src/backup_copy.cc
rahulmalik87: yeah, It is ensure the decrypting is not complete.

storage/innobase/xtrabackup/test/t/check_for_passwords_debug.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/check_for_passwords.sh
rahulmalik87: I can't find a way to ensure the process would be blocked untill we run ps

altmannmarcelo (approved):
LGTM

PR #1341, opened 2022-06-03
PXB-2702 Backup fails when undo truncation log is present

rahulmalik87:
https://jira.percona.com/browse/PXB-2702

Problem:
Backup fails if there is left over undo truncation(_truc.log) file
present in server data/undo directory.

Analysis:
Due to some bugs in MySQL truncation process does not complete,
leaving MySQL instance alive, but with undo truncation logs present.
PXB during backup tries to fix that by invoking srv_undo_tablespace_fixup_num
on server data/undo directory.

Fix:
PXB should not create undo file on server at any circumstances.
Avoid creating any undo file on server during backup.
PXB backup fails if undo truncation happens during backup

Thank you Dmitry Smal for your contribution for initial patch.

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -776,6 +776,23 @@ static dberr_t srv_undo_tablespace_fixup_num(space_id_t space_num) {
   if (!undo::is_active_truncate_log_present(space_num)) {
     return (DB_SUCCESS);
   }
+#ifdef XTRABACKUP
+  if (srv_backup_mode) {
+    if (opt_lock_ddl) {
+      xb::warn() << "Found _trunc.log in server dirctory. ";
+      return (DB_SUCCESS);
+    } else {
+      xb::error()
+          << "Found _trunc.log file in server directory. Undo truncation might "
+             "be concurrently running, Try backup after undo truncation is "
+             "complete. These *_trunc.log could be left over from previous "
+             "undo truncation, remove them safely to continue backup. You can "
+             "also run backup with --lock-ddl=true that blocks concurrent undo "
+             "truncation";
+      return (DB_ERROR);
+    }
+  }
+#endif
 
   ib::info(ER_IB_MSG_1077, ulong{space_num});
 
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -25,6 +25,7 @@ extern bool innodb_log_checksums_specified;
 extern bool innodb_checksum_algorithm_specified;
 
 extern bool opt_lock_ddl_per_table;
+extern bool opt_lock_ddl;
 extern bool redo_catchup_completed;
 extern bool opt_page_tracking;
 extern char *xtrabackup_incremental;
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -1051,6 +1051,10 @@ bool is_active_truncate_log_present(space_id_t space_num) {
   truncate action was complete. */
 
   if (os_file_exists(log_file_name)) {
+#ifdef XTRABACKUP
+    if (srv_backup_mode) return (true);
+#endif
+
     bool ret;
     pfs_os_file_t handle = os_file_create_simple_no_error_handling(
         innodb_log_file_key, log_file_name, OS_FILE_OPEN, OS_FILE_READ_WRITE,


storage/innobase/srv/srv0start.cc
```diff
@@ -1123,11 +1123,15 @@ static dberr_t srv_undo_tablespaces_open(bool backup_mode) {
   for (space_id_t num = 1; num <= FSP_MAX_UNDO_TABLESPACES; ++num) {
     /* Check if this undo tablespace was in the process of being truncated.
     If so, recreate it and add it to the construction list. */
-    dberr_t err = srv_undo_tablespace_fixup_num(num);
-    if (err != DB_SUCCESS) {
-      undo::spaces->x_unlock();
-      return (err);
+#ifdef XTRABACKUP
+    if (!backup_mode) {
+      dberr_t err = srv_undo_tablespace_fixup_num(num);
```
satya-bodapati: This undo tablespace fixup is only for case with lock-ddl=OFF, right?
and it will be best effort only. ie prepare can fail with lock-ddl=OFF & undo truncations

rahulmalik87: this is required for lock-ddl true as well. Because of existing bug in MySQL, there are left over _trunc.log files.( I could not reproduce this case  but as pointed by user It is possible) Also it wrong  if PXB tries to create any file on server directory 

Backup fails for concurrent undo truncation happens. [PXB-2255](https://jira.percona.com/browse/PXB-2255)

satya-bodapati (approved):

altmannmarcelo (changes requested):
From what I remember, we agreed that we will only allow the --backup to continue in case of the presence of `_truc.log` file and --lock-ddl=ON, as this guarantees that there is no truncation in progress and the log file is a left over. In case of --lock-ddl=OFF we do not know for sure if the truncation is not happening while we are copying the undo files.

During backup, we copy undo files, what will happen with the backup if we copy a file that is halfway through truncation?

rahulmalik87:
> From what I remember, we agreed that we will only allow the --backup to continue in case of the presence of `_truc.log` file and --lock-ddl=ON, as this guarantees that there is no truncation in progress and the log file is a left over. In case of --lock-ddl=OFF we do not know for sure if the truncation is not happening while we are copying the undo files.
> 
> During backup, we copy undo files, what will happen with the backup if we copy a file that is halfway through truncation?

Hello @altmannmarcelo, I have created [PXB-2263](https://jira.percona.com/browse/PXB-2263) to allow the server undo truncation without lock-ddl.  This patch only solves issue with lock-ddl.

storage/innobase/srv/srv0start.cc
```diff
@@ -776,6 +776,22 @@ static dberr_t srv_undo_tablespace_fixup_num(space_id_t space_num) {
...
+      xb::warn() << "Found _trunc.log in server dirctory. ";
+      return (DB_SUCCESS);
+    } else {
+      xb::error()
+          << "Found _trunc.log file in server directory. Undo truncation might "
+             "be concurrently running, Try backup after undo truncation is "
+             "complete. These _trunc.log could be left over of previous undo "
+             "truncation remove them safely to continue backup. You can also "
+             "run backup with --lock-ddl=true which blocks concurrent undo "
+             "truncation";
```
satya-bodapati: arent we supposed to return with error from here?

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed...
Missed that during development. lock test was successful

> 2022-07-12T13:01:08.348170+05:30 0 [ERROR] [MY-011825] [Xtrabackup] Found _trunc.log file in server directory. Undo truncation might be concurrently running, Try backup after undo truncation is complete. These _trunc.log could be left over of previous undo truncation remove them safely to continue backup. You can also run backup with --lock-ddl=true which blocks concurrent undo truncation
> 2022-07-12T13:01:08.348192+05:30 0 [Note] [MY-012902] [InnoDB] Undo tablespace number 1 was being truncated when mysqld quit.
> 2022-07-12T13:01:08.348201+05:30 0 [ERROR] [MY-012903] [InnoDB] Cannot recover a truncated undo tablespace in read-only mode
> 2022-07-12T13:01:08.348230+05:30 0 [ERROR] [MY-011825] [Xtrabackup] xb_load_tablespaces() failed with error code 57

and after  returning with DB_ERROR

> > 2022-07-12T13:02:18.704277+05:30 0 [ERROR] [MY-011825] [Xtrabackup] Found _trunc.log file in server directory. Undo truncation might be concurrently running, Try backup after undo truncation is complete. These _trunc.log could be left over of previous undo truncation remove them safely to continue backup. You can also run backup with --lock-ddl=true which blocks concurrent undo truncation
> 2022-07-12T13:02:18.704341+05:30 0 [ERROR] [MY-011825] [Xtrabackup] xb_load_tablespaces() failed with error code 11

storage/innobase/srv/srv0start.cc
```diff
@@ -776,6 +776,23 @@ static dberr_t srv_undo_tablespace_fixup_num(space_id_t space_num) {
...
+  if (srv_backup_mode) {
+    if (opt_lock_ddl) {
+      xb::warn() << "Found _trunc.log in server dirctory. ";
+      return (DB_SUCCESS);
+    } else {
+      xb::error()
+          << "Found _trunc.log file in server directory. Undo truncation might "
+             "be concurrently running, Try backup after undo truncation is "
+             "complete. These _trunc.log could be left over of previous undo "
+             "truncation remove them safely to continue backup. You can also "
```
satya-bodapati: These *_trunc.log files could be left over from previous undo  truncation, remove them.."

storage/innobase/srv/srv0start.cc
```diff
@@ -776,6 +776,23 @@ static dberr_t srv_undo_tablespace_fixup_num(space_id_t space_num) {
...
+             "run backup with --lock-ddl=true which blocks concurrent undo "
```
satya-bodapati: that blocks

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

satya-bodapati (approved):

altmannmarcelo (approved):
LGTM

PR #436, opened 2017-10-06
PXB-197: 5.6 MTR PXB failure in galera_sst_xtrabackup-v2-options on all nodes

VasilyNemkov:
Ported from 2.3

Fixed various bugs in ds_decrypt.c, all related to caller of decrypt_write
providing data in smaller portions than expected.
Added test for PXB-197 issue;
Updated xbcrypt.sh test.

Original PR: https://github.com/percona/percona-xtrabackup/pull/425
Jira issue: https://jira.percona.com/browse/PXB-197
Jenkins param-build: http://jenkins.percona.com/job/percona-xtrabackup-2.4-param-medium/28/

PR #1342, opened 2022-06-03
Codership: Allow building PXB from pull request #436

abychko:
this patch is to pack all attached submodules into resulting source tarball with `make dist'


altmannmarcelo:
Hi @abychko ,

First of all, thanks for your contribution. However, this change might affect our building scripts, which all work without a problem. Furthermore, we don't have the buffer to validate that this change will not break any of our internal building farms. So, I will just close this PR.

PR #1344, opened 2022-06-15
Implement PXB-2669 - ZSTD compression

altmannmarcelo:
https://jira.percona.com/browse/PXB-2669

Compress -

ZSTD compression works by basically just consuming the read buffer and
passing it to ZSTD_compressStream2. We utilize ZSTD build-in Thread
Pooling. We now enable support for multi thread to static ZSTD.
Adjusted required files to compress and decompress via xtrabackup.
Modified copy_file to handle empty files as ZSTD will refuse to run if
we provide an empty file. It requires that the file at least have a ZSTD
like header with empty content.

Decompress -

Compression using xtrabackup --compress=zstd & xtrabackup --decompress
works fine as we do everything in one go (read the entire file and pass
it to zstd client).

Xbstream is a bit different. There is no alignment between the buffers
we deal with in compression and stream(Read buffer, compression and
xbstream chunk).

Let's take an example of hypothetical File 1. We will have 3 layers of
data:

1. Raw file
We read raw file in read_buffer_size chunks (Default 10Mb):

```
+------------------------------------------------------------+
| File 1                                                     |
+------------------------------------------------------------+
                                ^
                                |
                              10 Mb
```
2. Compressed data
Read buffer is passed into ds_compress_zstd and gets compressed into
one or multiple Frames:
```
+------------------+--------------- +
|  File 1 Frame 1  | File 1 Frame 2 |
+------------------+----------------+
```
Each frame has a header(H) and 1 or more blocks(Bn) with variant length:

```
|  File 1 Frame 1  | File 1 Frame 2 |
+------------------+--------------- +
|H| B1   |B2| B3   |H| B1 |    B2   |
+------------------+----------------+
```
3. XBStream data
xbstream will then write data into multiple chunks of read_buffer_size
```
+------------------$---+-------------+
| XBS chunk 1      $   | XBS chunk 2 |
+------------------$---+-------------+
                   $
             End of File1 Frame 1
```
At xbstream, we receive data as chunks (represented by item #3).
XBStream chunk 1 will have the complete data of File 1 Frame 1 (F1F1)
and part of the data of File 1 Frame 2 (F1F2), while the remaining data
of that frame will only be available at XBStream chunk 2. Thus we need
to first parse xbstream data reading it with the ZSTD compression format
https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md
in order to validate when we should send the data of F1F1 to ZSTD
decompress functions. Also we need to take into consideration that part
of the buffer from chunk1 needs to be saved and append at the beginning
on XBS chunk 2 in order to see the full F1F2 data (reassembly item -
2. Compressed data).

Thus we need to create a ring buffer to parse each xbstream chunk.
Ring buffer work as follow:
 - When we receive a new xb chunk we save it as a new buffer in the ring
buffer at ds_istream.h.
 - Before checking if we have a complete frame we save current position
of ring buffer. We try to validate if our continuous buffer has a
complete frame. We then validate the return of
ZSTD_findFrameCompressedSize. If we have a complete frame we return
ZSTD_OK and the correct frame size (this will be used to decompress the
frame at the main function). In case of an error, we validate if the
error is due to wrong src size meaning this is a partial buffer, in
which case we return ZSTD_INCOMPLETE or ZSTD_ERROR on all other cases.
We always restore the ring buffer position back to the saved position,
as the main funtion is the one responsible to advance the buffer by
the correct frame length bytes.

Decompression:
We utilize ZSTD buffer structs. There is the In(read) and out(write)
buffer. Those buffers receive a pointer to the real buffer where data
will be read/write to and maintain a track of current position we have
processed up to.
On each chunk of data decompressed, we write it out to disk and if
checksum flag is set on frame header, we update the checksum computed
so far.
We keep consuming the read buffer until we have decompressed all the
data. As a final step to consider the frame complete, we validate the
checksum if necessary.

Testing
Extended the necessary test to also cover ZSTD compression.
Adjusted page_compression.sh as the last step we were testing sparseness
on backup files, while the intention is to test sparseness on datadir
files to ensure --copy-back did not loose it.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/207/

storage/innobase/xtrabackup/src/ds_decompress_zstd.cc
```diff
@@ -0,0 +1,341 @@
...
+        frame_size = stream.length_from_to(start_frame_cur, start_frame_pos,
+                                           end_frame_cur, end_frame_pos);
+        stream.restore_pos(start_frame_cur, start_frame_pos);
+        /* we have a complete frame */
+        return ZSTD_OK;
+      }
+    } while (!block_info.last_block);
+  }
+  return ZSTD_ERROR;
+}
```
inikep: You shouldn't read ZSTD headers by yourself. `ZSTD_findFrameCompressedSize` should allow to check if we have a full frame:
```
size_t ZSTD_findFrameCompressedSize(const void* src, size_t srcSize);
 `src` should point to the start of a ZSTD frame or skippable frame.
 `srcSize` must be >= first frame size
 @return : the compressed size of the first frame starting at `src`,
           suitable to pass as `srcSize` to `ZSTD_decompress` or similar,
        or an error code if input is invalid 
```

altmannmarcelo: @inikep I'm using `ZSTD_getFrameHeader` to parse the frame header. I just have a custom function `parse_frame_header` so I can return `ZSTD_stream::err_t` to `ZSTD_stream::next_frame` which has the logic to concatenate the buffers in case of partial frame has been received.

rahulmalik87:
@marce, Seeing some compilation error and warning on macos. Please fix. 

th/libprotobuf.3.19.4.dylib /Users/rahulmalik/MySQL/src/x8/bld/runtime_output_directory/xtrabackup
Undefined symbols for architecture x86_64:
  "_ZSTD_createThreadPool", referenced from:
      compress_init(char const*) in ds_compress_zstd.cc.o
  "_ZSTD_freeThreadPool", referenced from:
      compress_deinit(ds_ctxt*) in ds_compress_zstd.cc.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.

[12/264] Building CXX object storage/in...es/xbstream.dir/ds_decompress_zstd.cc.o
/Users/rahulmalik/MySQL/src/x8/storage/innobase/xtrabackup/src/ds_decompress_zstd.cc:185:44: warning: variable 'start_frame_cur' may be uninitialized when used here [-Wconditional-uninitialized]
        frame_size = stream.length_from_to(start_frame_cur, start_frame_pos,
                                           ^~~~~~~~~~~~~~~
/Users/rahulmalik/MySQL/src/x8/storage/innobase/xtrabackup/src/ds_decompress_zstd.cc:143:25: note: initialize the variable 'start_frame_cur' to silence this warning
  size_t start_frame_cur, end_frame_cur, start_frame_pos, end_frame_pos = 0;
                        ^
                         = 0
/Users/rahulmalik/MySQL/src/x8/storage/innobase/xtrabackup/src/ds_decompress_zstd.cc:185:61: warning: variable 'start_frame_pos' may be uninitialized when used here [-Wconditional-uninitialized]
        frame_size = stream.length_from_to(start_frame_cur, start_frame_pos,
                                                            ^~~~~~~~~~~~~~~
/Users/rahulmalik/MySQL/src/x8/storage/innobase/xtrabackup/src/ds_decompress_zstd.cc:143:57: note: initialize the variable 'start_frame_pos' to silence this warning
  size_t start_frame_cur, end_frame_cur, start_frame_pos, end_frame_pos = 0;
                                                        ^

storage/innobase/xtrabackup/src/ds_decompress_zstd.cc
inikep: @altmannmarcelo I think you can reimplement `ZSTD_stream::next_frame` using just `ZSTD_findFrameCompressedSize` and get rid of ZSTD header parsing because you don't need to go to such a low level. You should assume that working with ZSTD headers is prohibited because ZSTD API functions should be enough for our needs.

altmannmarcelo: re-wrote `ZSTD_stream::next_frame` using `ZSTD_findFrameCompressedSize` - it's way simpler now!

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
```diff
@@ -0,0 +1,198 @@
...
+                             comp_size, 0};
+    size_t const remaining =
+        ZSTD_compressStream2(comp_file->cctx, &output, &input, mode);
+    comp_file->raw_bytes += input.pos;
+    comp_file->comp_bytes += output.pos;
+    finished = last_chunk ? (remaining == 0) : false;
+
+    if (last_chunk && finished) {
+      break;
+    }
```
inikep: Isn't it the same as:
```
if (last_chunk && remaining == 0) {
```

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
```diff
@@ -0,0 +1,198 @@
...
+      comp_file->comp_bytes = 0;
+    }
+    ZSTD_EndDirective const mode = last_chunk ? ZSTD_e_end : ZSTD_e_continue;
+    ZSTD_inBuffer input = {((const char *)buf + comp_file->raw_bytes),
+                           read_chunk_size, 0};
+    bool finished = false;
+    ZSTD_outBuffer output = {comp_file->comp_buf + comp_file->comp_bytes,
+                             comp_size, 0};
+    size_t const remaining =
+        ZSTD_compressStream2(comp_file->cctx, &output, &input, mode);
```
inikep, line 151: Please check the result with `ZSTD_isError()`.

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -1103,6 +1103,13 @@ function require_lz4()
     fi
 }
 
+function require_zstd()
```
rahulmalik87, line 1106: we are building PXB with zstd bundle. so I think we do not need zstd to be installed?

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
```diff
@@ -0,0 +1,198 @@
...
+    return nullptr;
+  }
+
+  ds_compress_file_t *comp_file = new ds_compress_file_t;
+  comp_file->dest_file = dest_file;
+  comp_file->comp_buf = nullptr;
+  comp_file->comp_buf_size = 0;
+  comp_file->raw_bytes = 0;
+  comp_file->comp_bytes = 0;
+  ZSTD_CCtx *cctx = ZSTD_createCCtx();
```
rahulmalik87, line 94: can we reuse the context or  use one different context per thread for parallel execution.

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
```diff
@@ -0,0 +1,198 @@
...
+static int compress_close(ds_file_t *file);
+static void compress_deinit(ds_ctxt_t *ctxt);
+
+datasink_t datasink_compress_zstd = {&compress_init,  &compress_open,
+                                     &compress_write, nullptr,
+                                     &compress_close, &compress_deinit};
+
+static ds_ctxt_t *compress_init(const char *root) {
+  ds_compress_ctxt_t *compress_ctxt = new ds_compress_ctxt_t;
+  compress_ctxt->thread_pool =
```
rahulmalik87, line 61: if we are using compression threads, there would be a dedicated thread for compression. I do not fully understand what would be the performance benefit of using a thread pool.

rahulmalik87 (changes requested):

storage/innobase/xtrabackup/test/inc/common.sh
altmannmarcelo: Good point, we should check either zstd binary or symbols on xtrabackup

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
altmannmarcelo: We already use parallel execution. ZSTD requires all thread to use the same context.

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
altmannmarcelo: We use ZSTD thread pool. we create a thread pool that will have xtrabackup_compress_threads thread working to deal with the queue of request.

storage/innobase/xtrabackup/test/inc/common.sh
altmannmarcelo: @rahulmalik87 actually we do call zstd for decompression:

```
  cmd << " | zstd -d ";
```

storage/innobase/xtrabackup/src/ds_compress_zstd.cc
rahulmalik87: What is mean is this piece of code we call for each file we are compressing Can't we reuse compression context? 
` ZSTD_CCtx *cctx = ZSTD_createCCtx();
  ZSTD_CCtx_refThreadPool(cctx, comp_ctxt->thread_pool);
  ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel,
                         xtrabackup_compress_zstd_level);
  ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, xtrabackup_compress_threads);
  ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1);`

rahulmalik87 (approved):
LGTM

inikep (approved):
LGTM

PR #1352, opened 2022-07-07
PXB-2809 wsrep_sync_wait<>0 causes Lock wait timeout exceeded; try re…

rahulmalik87:
…starting transaction

https://jira.percona.com/browse/PXB-2809

Problem:
wsrep_sync_wait<>0 causes Lock wait timeout exceeded; try restarting
transaction

Fix:
SET SESSION wsrep_sync_wait=0 right after connection to PXC

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -189,6 +189,12 @@ xb_mysql_connect()
 	xb_mysql_query(connection, "SET NAMES utf8",
 		       false, true);
 
+	if (xb_mysql_numrows(connection, "SELECT @@wsrep_sync_wait", false) >
+				0) {
+		xb_mysql_query(connection, "SET SESSION wsrep_sync_wait=0", false,
+				true);
+	}
+
 	return(connection);
 }
 


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -188,8 +188,13 @@ xb_mysql_connect()
...
 	xb_mysql_query(connection, "SET NAMES utf8",
 		       false, true);
+        if (xb_mysql_numrows(connection, "SELECT @@wsrep_sync_wait", false) >
+            0) {
+          xb_mysql_query(connection, "SET SESSION wsrep_sync_wait=0", false,
+                         true);
+        }
 
-	return(connection);
+        return(connection);
```
satya-bodapati: please use tabs in 2.4 for indentation.

satya-bodapati (changes requested):

altmannmarcelo (commented):
@rahulmalik87 did we check if GR also has a similar configuration?

Also please ensure we can run this in any pxc_strict_mode (should not be an issue, but worth checking)

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -189,6 +189,12 @@ xb_mysql_connect()
...
 		       false, true);
 
+	if (xb_mysql_numrows(connection, "SELECT @@wsrep_sync_wait", false) >
+				0) {
+		xb_mysql_query(connection, "SET SESSION wsrep_sync_wait=0", false,
+				true);
+	}
+
 	return(connection);
 }
```
satya-bodapati, line 199: @rahulmalik87 can you please check against upstream? would this log an error message in logs?
upstream doesn't have wsrep_sync_wait

satya-bodapati (changes requested):

fraenki:
Hi @rahulmalik87, as mentioned by @altmannmarcelo in https://github.com/percona/percona-xtrabackup/pull/1370 the extra call to `wsrep_causal_reads` must also be removed, because it is no longer supported in recent versions (and leads to an unrecoverable error).

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: Hello @satya-bodapati, no SELECT @@nonexist_Variables doesn't throw any error message in logs.

rahulmalik87:
> Hi @rahulmalik87, as mentioned by @altmannmarcelo in #1370 the extra call to `wsrep_causal_reads` must also be removed, because it is no longer supported in recent versions (and leads to an unrecoverable error).

Thnak you @fraenki  for the feedback. I have removed `wsrep_casual_reads` from 8.0

rahulmalik87:
> @rahulmalik87 did we check if GR also has a similar configuration?
> 
> Also please ensure we can run this in any pxc_strict_mode (should not be an issue, but worth checking)

Thanks @altmannmarcelo for suggestion. I have added group-replication-consistency=EVENTUAL , which could help us to avoid lock wait.

satya-bodapati:
> > @rahulmalik87 did we check if GR also has a similar configuration?
> > Also please ensure we can run this in any pxc_strict_mode (should not be an issue, but worth checking)
> 
> Thanks @altmannmarcelo for suggestion. I have added group-replication-consistency=EVENTUAL , which could help us to avoid lock wait.

@rahulmalik87 Are you yet to update PR ? the changes about wserp_casual_reads and consistency=EVENTUAL are not yet visible.

rahulmalik87:
> > > @rahulmalik87 did we check if GR also has a similar configuration?
> > > Also please ensure we can run this in any pxc_strict_mode (should not be an issue, but worth checking)
> > 
> > 
> > Thanks @altmannmarcelo for suggestion. I have added group-replication-consistency=EVENTUAL , which could help us to avoid lock wait.
> 
> @rahulmalik87 Are you yet to update PR ? the changes about wserp_casual_reads and consistency=EVENTUAL are not yet visible.

That is only applicable for 8.0. https://github.com/percona/percona-xtrabackup/pull/1374

satya-bodapati (approved):

PR #1361, opened 2022-07-29
Fixed PXB-2810 (Panic when AIO initialization failed)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2810

Problem:
xb_fil_io_init does not return an error when os_aio_init fail to
initialize causing the backup to crash.

Fix:
Adjust xb_fil_io_init and function depending on it to properly handle
initialization errors.

Thanks to Yan Huang for the patch.

Closes #1346


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/215/

PR #1362, opened 2022-07-29
Fixed PXB-2517 - Check if files-from is successfully closed before rsync

altmannmarcelo:
https://jira.percona.com/browse/PXB-2517

Issue
When hard disks ard under heavy workload, `fclose() ` of
`xtrabackup_rsyncfiles_pass` may fail.
And the content of any unwritten output buffer will be lost.
So `rsync` will fail due to incomplete `--files-from` or synchronize
incomplete file list.

Solution
Call fflush before flcose and check if files-from is successfully
closed before rsync.

Thanks to Garen Chan for the patch.

Closes #1100


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-multijob/192/

PR #1363, opened 2022-07-29
Fixed PXB-2840 (Create bucket does not work with regional endpoint)

altmannmarcelo:
https://jira.percona.com/browse/PXB-2840

Problem:
When specifying a non existing bucket, xbcloud will attempt to create
it. If we are using a non-default region (--s3-region=...) xbcould will
be using a regional endpoint (either by computing it or via
-s3-endpoint= parameter). Non default endpoint requires location
constraint to be set at bucket creation:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html#API_CreateBucket_RequestBody

Fix:
Adjust S3_client::create_bucket to properly set LocationConstraint on
API request.

Thanks to Jason Zareski for the patch.

Closes #882


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-2.4-multijob/45/

PR #1371, opened 2022-09-13
PERCONA_SCHEMA.xtrabackup_history CHARACTER SET

fmbiete:
xtrabackup_history is using "utf8".

With "utf8" being an alias and Oracle modifying back and forward this alias to utf8mb3, utf8mb4, I think it makes sense to use the full charset name. 

We also should use "utf8mb4" as a future-proof option and convert existing tables.


altmannmarcelo:
Hi @fmbiete .

Thanks for your contribution. We will incorporate it as part of https://jira.percona.com/browse/PXB-3103.

PR #1373, opened 2022-09-21
Implemented PXB-2866 Make PXB 8.0.30 be able to backup 8.0.29 (and pr…

altmannmarcelo:
…ior) server

https://jira.percona.com/browse/PXB-2866

Adjusted redo log logic to populate start and end lsn of old redo log layout.
Changed redo_log.cc to use log_sys->m_files_ctx instead of creating a new Log_files_context each time. log_sys->m_files_ctx will already have the correct Log_files_ruleset (CURRENT or PRE_8_0_30) that can be used to identify if we are dealing with ib_logfile or #ib_redo layout.

Fixed --prepare warning about redo been created by unknown creator.

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1546,7 +1546,8 @@ static dberr_t log_sys_handle_creator(log_t &log) {
                      "read-only mode!";
       return DB_ERROR;
     }
-  } else if (!str_starts_with(creator_name, "MySQL ")) {
+  } else if (!str_starts_with(creator_name, "MySQL ") &&
+             !str_starts_with(creator_name, "xtrabkup ")) {
     ib::warn(ER_IB_MSG_LOG_FILES_CREATED_BY_UNKNOWN_CREATOR,
              creator_name.c_str());
   }
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1546,8 +1546,7 @@ static dberr_t log_sys_handle_creator(log_t &log) {
                      "read-only mode!";
       return DB_ERROR;
     }
-  } else if (!str_starts_with(creator_name, "MySQL ") &&
-             !str_starts_with(creator_name, "xtrabkup ")) {
+  } else if (!str_starts_with(creator_name, "MySQL ")) {
     ib::warn(ER_IB_MSG_LOG_FILES_CREATED_BY_UNKNOWN_CREATOR,
              creator_name.c_str());
   }
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3844,17 +3844,56 @@ bool meb_scan_log_recs(
 #endif
         auto data_len = block_header.m_data_len;
 
-    if (scanned_lsn + data_len > recv_sys->scanned_lsn &&
-        recv_sys->scanned_epoch_no > 0 &&
-        !log_block_epoch_no_is_valid(block_header.m_epoch_no,
-                                     recv_sys->scanned_epoch_no)) {
-      /* Garbage from a log buffer flush which was made
-      before the most recent database recovery */
+#ifdef XTRABACKUP
+    /**
+     * 8.0.30 records LOG_BLOCK_EPOCH_NO at 8th by of log block header
+     * while pre-8.0.30 we recorded LOG_BLOCK_CHECKPOINT_NO
+     */
+    if (scanned_lsn + data_len > recv_sys->scanned_lsn) {
+      if (xtrabackup_original_log_format >= Log_format::VERSION_8_0_30 &&
+          recv_sys->scanned_epoch_no > 0 &&
+          !log_block_epoch_no_is_valid(block_header.m_epoch_no,
+                                       recv_sys->scanned_epoch_no)) {
+        /* Garbage from a log buffer flush which was made
+        before the most recent database recovery */
 
-      finished = true;
+        finished = true;
 
-      break;
+        break;
+      } else {
+        /** block_header.m_epoch_no = log_block_get_checkpoint_no
+         * recv_sys->scanned_epoch_no = recv_sys->scanned_checkpoint_no
+         *  log_block_get_checkpoint_no(log_block) <
+            recv_sys->scanned_checkpoint_no &&
+        (recv_sys->scanned_checkpoint_no -
+             log_block_get_checkpoint_no(log_block) >
+         0x80000000UL)
+         */
+        if (block_header.m_epoch_no < recv_sys->scanned_epoch_no &&
+            (recv_sys->scanned_epoch_no - block_header.m_epoch_no >
+             0x80000000UL)) {
+          /* Garbage from a log buffer flush which was made
+          before the most recent database recovery */
+
+          finished = true;
+
+          break;
+        }
+      }
     }
+#else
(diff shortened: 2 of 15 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -1138,7 +1138,9 @@ static
       return;
     }
 
-    if (!file.contains(checkpoint_in_file.m_checkpoint_lsn)) {
+    if (IF_XB(log.m_files_ctx.m_files_ruleset >
+              Log_files_ruleset::PRE_8_0_30) &&
+        !file.contains(checkpoint_in_file.m_checkpoint_lsn)) {
```
satya-bodapati, line 1143: why isn't applicable for pre 8030? checkpoint_lsn would be in the file detected above, isn't it?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -65,11 +65,8 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
   /* Redo log file header is never encrypted. */
   Encryption_metadata unused_encryption_metadata;
 
-  Log_files_context log_files_ctx =
-      Log_files_context{srv_log_group_home_dir, Log_files_ruleset::CURRENT};
-
   auto file_handle =
-      Log_file::open(log_files_ctx, checkpoint.m_checkpoint_file_id,
+      Log_file::open(log_sys->m_files_ctx, checkpoint.m_checkpoint_file_id,
```
satya-bodapati: where do we populate log_sy->m_files_ctx?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -132,12 +125,10 @@ static bool reopen_log_files() {
   log.m_log_flags = log_flags;
   log.m_log_uuid = log_uuid;
   log.m_files = std::move(files);
-  auto file = log.m_files.front();
-  log_encryption_read(log, file);
+  auto file = log.m_files.find(desired_lsn);
```
satya-bodapati, line 152: is the encryption information only in one of the ib_logfile* files?

side question: do we write encryption info in every ib_redo* file?

storage/innobase/log/log0recv.cc
altmannmarcelo: on pre-8.0.30 we always have checkpoint marker on ib_logfile0, even if the LSN it points to is located on ib_logfile1. This causes the backup to fail in case we are currently writing on ib_logfile1.

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: at the end'ish of log_sys_init in log0log.cc

storage/innobase/log/log0log.cc
```diff
@@ -1546,7 +1546,8 @@ static dberr_t log_sys_handle_creator(log_t &log) {
                      "read-only mode!";
       return DB_ERROR;
     }
-  } else if (!str_starts_with(creator_name, "MySQL ")) {
+  } else if (!str_starts_with(creator_name, "MySQL ") &&
+             !str_starts_with(creator_name, "xtrabkup ")) {
```
rahulmalik87: we have added LOG_HEADER_CREATOR_PXB, so It is already handled.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -132,12 +125,10 @@ static bool reopen_log_files() {
...
+  if (file == log.m_files.end()) return false;
+  log_encryption_read(log, *file);
```
rahulmalik87: Better to check db_err of log_encryption_read and return false, because it can fail if the file is removed by the server.  
Also we could only try to read encryption if redo_file is encrypted.

rahulmalik87 (changes requested):
How do we ensure wrap-around of redo files?

storage/innobase/log/log0recv.cc
satya-bodapati: can you please add this as a comment? Maybe for versions < 8_0_30 we can check that the front file will always have checkpoint header?

satya-bodapati: Another doubt I have is how will we find wrap around logic? because we always have two files loaded in loop. 
Unlike 8.0.30, the loaded files start_lsn will keep changing. In 8.0.30, the files start_lsn is fixed and new files appear.

altmannmarcelo: @satya-bodapati logic explained as part of commit message (adjusted)
wrap around logic has been fixed at 8d8feb3

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: Adjusted to get it on the file contained the required LSN.

> side question: do we write encryption info in every ib_redo* file?

just the ones that have a checkpoint.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -162,8 +186,19 @@ lsn_t Redo_Log_Reader::read_log_seg(log_t &log, byte *buf, lsn_t start_lsn,
 
   do {
     os_offset_t source_offset;
-
-    source_offset = file->offset(start_lsn);
+    if (log.m_files.ctx().m_files_ruleset == Log_files_ruleset::CURRENT) {
+      source_offset = file->offset(start_lsn);
```
satya-bodapati, line 193: @altmannmarcelo for pre 8030 files, cant we populate a file object with m_start_lsn ? Then same logic should work ?

satya-bodapati: ie ib_logfile0 & ib_logfile1 map to Log_file objects with m_start_lsn. Is this possible or not?

altmannmarcelo: @satya-bodapati as per our slack conversation and research, the problem comes when we remove redo logs and let the server re-create them. start_lsn is calculated different in this case on pre 8.0.30 and we cannot use `file->offset(lsn)` function to get the correct file offset.

storage/innobase/xtrabackup/test/t/pxb-2710-predict_memory.sh
```diff
@@ -6,8 +6,13 @@ require_debug_pxb_version
...
 innodb-flush-log-at-trx-commit=0
 sync-binlog=0
-innodb-redo-log-capacity=512M
 "
+if is_server_version_higher_than 8.0.29
+then
+  MYSQLD_EXTRA_MY_CNF_OPTS="${MYSQLD_EXTRA_MY_CNF_OPTS:-""}
+  innodb-redo-log-capacity=512M
+  "
+fi
```
satya-bodapati, line 15: else 
innodb-log-file-size=512M?

satya-bodapati (approved):

storage/innobase/xtrabackup/test/t/pxb-2710-predict_memory.sh
altmannmarcelo: it's not needed for pre 8.0.30 as we do not remove the files (wrap around) unless we actually write more than 48M*2

altmannmarcelo:
closed in favor of https://github.com/percona/percona-xtrabackup/pull/1377

PR #1374, opened 2022-09-30
Pxb 8.0 2809 wsrep_sync_wait<>0 causes Lock wait timeout exceeded; try restarting transaction

rahulmalik87:
The wsrep_causal_reads parameter was already deprecated in
    MySQL-wsrep 5.5.42-25.12.
    so use wsrep_sync_wait=0 to avoid lock wait timeout.

    Also for 8.0 GR,PXB now sets group-replication-consistency=EVENTUAL
    to avoid possible time out

    Thanks to Frank Well for providing initial patch

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -189,6 +189,12 @@ xb_mysql_connect()
 	xb_mysql_query(connection, "SET NAMES utf8",
 		       false, true);
 
+	if (xb_mysql_numrows(connection, "SELECT @@wsrep_sync_wait", false) >
+				0) {
+		xb_mysql_query(connection, "SET SESSION wsrep_sync_wait=0", false,
+				true);
+	}
+
 	return(connection);
 }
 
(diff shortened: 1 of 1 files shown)


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -171,6 +171,17 @@ MYSQL *xb_mysql_connect() {
 
   xb_mysql_query(connection, "SET SESSION wait_timeout=2147483", false, true);
 
+  if (xb_mysql_numrows(connection, "SELECT @@wsrep_sync_wait", false) > 0) {
```
altmannmarcelo, line 174: @rahulmalik87 can we use have_galera_enabled here instead of doing a new query?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -171,6 +171,17 @@ MYSQL *xb_mysql_connect() {
...
+    xb_mysql_query(connection, "SET SESSION wsrep_sync_wait=0", false, true);
+  }
+
+  if (xb_mysql_numrows(connection, "SELECT @@group-replication-consistency",
```
altmannmarcelo, line 178: Same here, have a var for GR enabled that we check when we do the initial SHOW GLOBAL VARIABLES and use this new var here.

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: @marcalff there are one query we execute before we set have_galera_enabled.
`    SELECT COUNT(*) FROM information_schema.tables WHERE engine = 'MyISAM' OR engine = 'RocksDB'`
So not sure if this can cause time out. So I choose better to test and set an approach.

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: I could not find var for GR. Also same reason previous commit.

altmannmarcelo (approved):
LGTM

PR #1375, opened 2022-10-05
Fixed PXB-2858 Prepare of full and incremental backup uses the same a…

altmannmarcelo:
…mount of buffer pool memory with memory estimation

https://jira.percona.com/browse/PXB-2858

Problem:
When preparing incremental backups, we call read_metadata twice. First time we read metadata file from incremental dir, but second time we read it from target-dir (full base backup). The second call was replacing redo_memory & redo_frames variable.

Fix:
Adjusted code to use a variable that gets populated conditionally in case we are doing a prepare on full or inc directory.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -2124,6 +2124,8 @@ static bool innodb_init_param(void) {
                     xtrabackup_use_free_memory_pct != 0;
 
   if (estimate_memory) {
+    ut_ad(real_redo_frames != UINT64_MAX);
+
     ulint free_memory_total = xtrabackup::utils::host_free_memory();
     ulint free_memory_usable =
         (free_memory_total * xtrabackup_use_free_memory_pct) / 100;
@@ -2151,9 +2153,9 @@ static bool innodb_init_param(void) {
       }
       xb::info() << "Setting buffer pool to: " << xtrabackup_use_memory;
     }
-    if ((long)redo_frames > innobase_open_files) {
-      xb::info() << "Setting innobase_open_files to: " << redo_frames;
-      innobase_open_files = (long)redo_frames;
+    if ((long)real_redo_frames > innobase_open_files) {
+      xb::info() << "Setting innobase_open_files to: " << real_redo_frames;
+      innobase_open_files = (long)real_redo_frames;
     }
   }
 
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -264,7 +264,7 @@ ulonglong xtrabackup_encrypt_chunk_size = 0;
 size_t redo_memory = 0;
 ulint redo_frames = 0;
 size_t real_redo_memory = 0;
-ulint real_redo_frames = 0;
+ulint real_redo_frames = UINT64_MAX;
 
 ulint xtrabackup_rebuild_threads = 1;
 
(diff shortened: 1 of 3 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2106,19 +2111,27 @@ static bool innodb_init_param(void) {
...
         (free_memory_total * xtrabackup_use_free_memory_pct) / 100;
-    ulint mem =
-        buf_pool_size_align(redo_memory + (redo_frames * UNIV_PAGE_SIZE));
+    ulint mem = buf_pool_size_align(real_redo_memory +
+                                    (real_redo_frames * UNIV_PAGE_SIZE));
     if (mem > (ulint)xtrabackup_use_memory) {
-      xb::info() << "Got estimation from backup: " << redo_memory
-                 << " bytes for parsing and " << redo_frames
+      xb::info() << "Got estimation from backup: " << real_redo_memory
+                 << " bytes for parsing and " << real_redo_frames
```
satya-bodapati, line 2136: little below we have this:
```
if ((long)redo_frames > innobase_open_files) {
      xb::info() << "Setting innobase_open_files to: " << redo_frames;
      innobase_open_files = (long)redo_frames;
}
```
shoudln't these be changed to real_redo_frames? (I think it is better to unset redo_frames/redo_memory after it is read to real_redo_frames/memory, may be use some max VALUES and add some assertions.  Idea is to prevent unintended usage)

satya-bodapati: and add ut_ad(real_redo/frames != UINT64_MAX)

satya-bodapati: another questions I have is why are we using redo_frames as files :thinking:

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: @satya-bodapati regarding the free frames - From the document design:
```
Adjusting the free frames also means xtrabackup will need to keep more files open simultaneously (open files limit), thus we also need to take into consideration if redo_frames will exceed innobase_open_files.
```
Pages are not individual files. But the more pages, higher is the chance to exceed default open_files.

satya-bodapati: ok ,some guesstimate :D
at the time of initialization, we will not upfront how many files we need to open. 

But i think it is OK to have fewer open files limit because innodb file LRU kicks in..
anyways, thanks for checking.. we can ignore the frames as files part of the question.

altmannmarcelo: > and add ut_ad(real_redo/frames != UINT64_MAX)

This is to make sure we are not exceeding innodb_buffer_pool_size max value?
If so, what about `ut_ad(real_redo_memory + (real_redo_frames * UNIV_PAGE_SIZE) < UINT64_MAX)`

satya-bodapati: nah, it was to ensure we are not reading uninitialized values of real_redo_frames.

altmannmarcelo: Ack. Will add

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2124,6 +2124,8 @@ static bool innodb_init_param(void) {
                     xtrabackup_use_free_memory_pct != 0;
 
   if (estimate_memory) {
+    ut_ad(real_redo_frames != UINT64_MAX);
```
satya-bodapati, line 2127: please fix the initialization (global value) to  UINT64_MAX

satya-bodapati (changes requested):

satya-bodapati (approved):

PR #1376, opened 2022-10-07
PXB-2831 8.0.30 merge

rahulmalik87:


sql/CMakeLists.txt
```diff
@@ -1098,136 +1098,7 @@ ENDIF()
...
-IF(APPLE AND WITH_LTO)
-  SET_TARGET_PROPERTIES(mysqld PROPERTIES INTERPROCEDURAL_OPTIMIZATION TRUE)
-ENDIF()
-
-IF(APPLE)
-  # Add CoreServices framework since some dloadable plugins may need it
-  FIND_LIBRARY(CORESERVICES NAMES CoreServices)
-  IF(CORESERVICES)
-    TARGET_LINK_LIBRARIES(mysqld ${CORESERVICES})
-  ENDIF()
```
satya-bodapati: what about other binaries? how do we skip them from creating? for example ibd2sdi, innochecksum and other binaries..?

satya-bodapati (changes requested):

storage/innobase/include/log0files_io.h
```diff
@@ -644,6 +644,15 @@ inline bool log_data_block_header_deserialize(const byte *buf,
...
 }
+#ifdef XTRABACKUP
+/** This version is used to recreate the log files in the same format that
+PXB read. i.e. if PXB is processing redo from 8.0.19 or below (v3), then
+it has to recreate log files with v3 header. If PXB is processing redo from
+8.0.20 above (v4), it has to recreate log files with v4 header.
+We cannot reuse log_sys->format for this purpose because it is dellocated and
+re-initialized during PXB prepare */
+extern uint32_t log_detected_format;
+#endif /*XTRABACKUP*/
```
satya-bodapati: better to place this in log0log.h?

storage/innobase/log/log0chkp.cc
```diff
@@ -318,203 +318,10 @@ static void log_update_available_for_checkpoint_lsn(log_t &log) {
...
-    LOG_HEADER_SET_FLAG(header_flags, LOG_HEADER_FLAG_NO_LOGGING);
-  }
-  if (crash_unsafe) {
-    LOG_HEADER_SET_FLAG(header_flags, LOG_HEADER_FLAG_CRASH_UNSAFE);
-  }
-  mach_write_to_4(buf + LOG_HEADER_FLAGS, header_flags);
-
-  log_block_set_checksum(buf, log_block_calc_checksum_crc32(buf));
-}
-
```
satya-bodapati: we no longer need log_detected_format and this code to set appropriate header?

satya-bodapati (changes requested):

cmake/mysql_version.cmake
```diff
@@ -78,51 +78,51 @@ MACRO(GET_MYSQL_VERSION)
...
-     DEFINED CLUSTER_PATCH_VERSION)
-    # Set MySQL Cluster version to the specific version defined in MYSQL_VERSION
-    SET(MYSQL_CLUSTER_VERSION "${CLUSTER_MAJOR_VERSION}")
-    SET(MYSQL_CLUSTER_VERSION
-      "${MYSQL_CLUSTER_VERSION}.${CLUSTER_MINOR_VERSION}")
-    SET(MYSQL_CLUSTER_VERSION
-      "${MYSQL_CLUSTER_VERSION}.${CLUSTER_PATCH_VERSION}")
-    IF(DEFINED CLUSTER_EXTRA_VERSION)
+  
+  IF(WITH_NDBCLUSTER)
```
satya-bodapati, line 82: Does this mean PS/upstream tries to compile NDB cluster by default?

storage/innobase/log/log0log.cc
```diff
@@ -559,7 +559,7 @@ that the proper size of the log buffer should be a power of two.
 @param[out]     log             redo log */
 static void log_calc_buf_size(log_t &log);
 
-uint32_t log_detected_format = UINT32_MAX;
+Log_format log_detected_format = Log_format::LEGACY;
```
satya-bodapati: Do we use this log_detected_format at all? And the redo log created during prepare will have the 8.0.30 tag which are applied and later removed.

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
```diff
@@ -3764,7 +3764,7 @@ static bool recv_scan_log_recs(log_t &log,
 #else  /* !UNIV_HOTBACKUP */
 bool meb_scan_log_recs(
 #endif /* !UNIV_HOTBACKUP */
-                               size_t max_memory, const byte *buf, size_t len,
+                               size_t *max_memory, const byte *buf, size_t len,
```
satya-bodapati: is this max_memory change due to predic_memory PR by marce?

satya-bodapati: you should also fix Doxygen comment, and change the in [in] to [in,out] parameter?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2053,7 +2045,6 @@ bool write_backup_config_file() {
     << innodb_checksum_algorithm_names[srv_checksum_algorithm] << "\n"
     << "innodb_log_checksums=" << srv_log_checksums << "\n"
     << "innodb_data_file_path=" << innobase_data_file_path << "\n"
-    << "innodb_log_files_in_group=" << srv_n_log_files << "\n"
```
satya-bodapati: for pre-8.0.30, shouldn't the conifg file contain innodb_log_files_in_group.

Similarly, shouldn't we save the new innodb_redo_log_capacity in the config for 8.0.30 server restore?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2053,7 +2045,6 @@ bool write_backup_config_file() {
...
     << "innodb_log_file_size=" << innobase_log_file_size << "\n"
```
satya-bodapati: we might have to consider when both the old (log)parameter and the new redo_log_capacity parameters are set.

storage/innobase/xtrabackup/src/backup_mysql.cc
satya-bodapati: infact, we might want to retain the config for 8.0.30 too if the user is still using the deprecated parameter

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -5019,17 +5019,14 @@ static bool xtrabackup_init_temp_log(void) {
 
   uint64_t file_size;
 
-  lsn_t max_no;
   lsn_t max_lsn = 0;
-  lsn_t checkpoint_no;
```
satya-bodapati: I believe we still need this to determine the right checkpoint_header to use. Or not?
max checkpoint_lsn is sufficient I guess. (from 8.0.30 checkoint_no is definitely not used)

@altmannmarcelo may be you can double-check if checkpoint_no is required for pre 8030 or not?

satya-bodapati (changes requested):

storage/innobase/include/log0recv.h
```diff
@@ -742,11 +742,31 @@ block.
...
+  /** File containing checkpoint header and checkpoint lsn. */
+  Log_file_id m_checkpoint_file_id{0};
+
+  /** Checkpoint header number. */
+  Log_checkpoint_header_no m_checkpoint_header_no{};
+
+  /** Checkpoint LSN. */
+  lsn_t m_checkpoint_lsn{0};
+};
+/** Find the latest checkpoint (check all existing redo log files).
```
satya-bodapati: I wonder why this code should go into #ifdef XTRABACKUP define. This is from upstream, isn't it?

storage/innobase/include/log0recv.h
```diff
@@ -742,11 +742,31 @@ block.
...
+@param[in,out]  log             redo log
+@param[out]     checkpoint      the latest checkpoint found (if any)
+@return true iff any checkpoint has been found */
+bool recv_find_max_checkpoint(log_t &log, Log_checkpoint_location &checkpoint);
+#endif /* XTRABACUP */
```
satya-bodapati: /* XTRABACKUP */

storage/innobase/log/log0recv.cc
```diff
@@ -790,8 +790,12 @@ with regards to the epoch_no it has stored in its header, during the recovery.
 @param[in]  log_block_epoch_no  epoch_no of the log data block to check
 @param[in]  last_epoch_no       epoch_no of the last data block scanned
 @return true iff the provided log block has valid epoch_no */
-static bool log_block_epoch_no_is_valid(uint32_t log_block_epoch_no,
-                                        uint32_t last_epoch_no) {
+#ifndef XTRABACKUP
+static
+#endif /* XTRABACK */
```
satya-bodapati: /* XTRABACUP */

storage/innobase/log/log0recv.cc
```diff
@@ -1051,6 +1055,7 @@ static dberr_t recv_log_recover_pre_8_0_30(log_t &log) {
   return log_start(log, checkpoint_lsn, checkpoint_lsn, nullptr);
 }
 
+#ifndef XTRABACKUP
```
satya-bodapati: ah, you are moving Log_checkpoint_location to new place.
The dilemma is whether to add #ifdef XTRABACKUP around the new place.

I think, we shouldn't add the #define. (see above)

storage/innobase/log/log0recv.cc
```diff
@@ -1107,8 +1113,11 @@ struct Log_checkpoint_location {
 @param[in,out]  log             redo log
 @param[out]     checkpoint      the latest checkpoint found (if any)
 @return true iff any checkpoint has been found */
-static bool recv_find_max_checkpoint(log_t &log,
-                                     Log_checkpoint_location &checkpoint) {
+#ifndef XTRABACKUP
+static
+#endif /*XTRABACKUP */
```
satya-bodapati: may be its time to create IF_N_XB() similar to IF_XB(). Then we can use readable code
IF_N_XB(static)
bool recv_find_max_checkpoint()..

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -84,28 +86,28 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
...
     log_files_header_read(*log_sys, max_cp_field);
 
-    if (checkpoint_no_start !=
-        mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_NO)) {
+    if (checkpoint_lsn_start !=
+        mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_LSN)) {
       checkpoint_lsn_start =
           mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_LSN);
-      checkpoint_no_start =
-          mach_read_from_8(log_sys->checkpoint_buf + LOG_CHECKPOINT_NO);
```
satya-bodapati: I wonder if we use this checkpoint_no stuff for pre-8.0.30. @altmannmarcelo can you please confirm if you are using this or not?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -84,28 +86,28 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
...
       continue;
     }
 
     break;
   }
 
+  */
```
satya-bodapati: Remove this entire block, it is commented out

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -300,7 +301,7 @@ ssize_t Redo_Log_Reader::scan_log_recs(byte *buf, bool is_last, lsn_t start_lsn,
     }
   }
 
-  *read_upto_lsn = scanned_lsn;
+  *read_upto_lsn = scanned_epoch_no;
```
satya-bodapati: is this a mistake? we are assigne block epoch_no to LSN?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -474,7 +475,7 @@ bool Redo_Log_Writer::write_buffer(byte *buf, size_t len) {
 
   if (srv_redo_log_encrypt) {
     IORequest req_type(IORequestLogWrite);
-    fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_first_id);
+    fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_id);
```
satya-bodapati: we are not supposed to use fil_system for redo log. This will give nullptr .
How does this code work?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -519,7 +520,7 @@ ssize_t Archived_Redo_Log_Reader::read_logfile(bool *finished) {
 
   if (srv_redo_log_encrypt) {
     IORequest req_type(IORequestLogRead);
-    fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_first_id);
+    fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_id);
```
satya-bodapati: same as above, gone in 8.0.30

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -915,9 +917,9 @@ static dberr_t open_or_create_log_file(bool *log_file_created, ulint i,
     /* Create in memory the file space object
     which is for this log group */
 
-    *log_space = fil_space_create(name, dict_sys_t::s_log_space_first_id,
+    *log_space = fil_space_create(name, dict_sys_t::s_log_space_id,
```
satya-bodapati: Hmm, looks like you answered my question above, so we will create fil_sapce_t object for redo log files.
You shouldn't do this because now redo file are multiple files and each one of them is like a different tablespace.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -935,14 +937,17 @@ bool Redo_Log_Data_Manager::init() {
...
+  if (!log_sys_init(false, flushed_lsn, flushed_lsn)) {
     return (false);
   }
 
   recv_sys_create();
   recv_sys_init();
 
+  /*XB30 srv_n_log_files */
+  ulong srv_n_log_files = 2;
   ut_a(srv_n_log_files > 0);
```
satya-bodapati: server guarantees at least two active redo files?
Also, assertion seems irrelevant.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -982,7 +987,8 @@ bool Redo_Log_Data_Manager::init() {
 
   ut_a(log_space != nullptr);
 
-  log_read_encryption();
+  /*XB30 add log_encryption_read */
```
satya-bodapati: remove it. (May be you fixed it in later commits :D not sure)

satya-bodapati (changes requested):

storage/innobase/include/log0log.h
```diff
@@ -380,4 +380,20 @@ void log_crash_safe_validate(log_t &log);
 
 #endif /* !UNIV_HOTBACKUP */
 
+#ifdef XTRABACKUP
```
satya-bodapati: if this code is moved around, this shouldn't be under ifdef XTRABACKUP tag.

storage/innobase/log/log0log.cc
```diff
@@ -1579,8 +1581,12 @@ static dberr_t log_sys_check_format(const log_t &log) {
...
 }
 
-static dberr_t log_sys_check_directory(const Log_files_context &ctx,
-                                       std::string &path, bool &found_files) {
+#ifndef XTRABACKUP
+static
+#endif /* XTRABACKUP */
+    dberr_t
+    log_sys_check_directory(const Log_files_context &ctx, std::string &path,
+                            bool &found_files) {
```
satya-bodapati: formatting looks messed up?

storage/innobase/log/log0log.cc
```diff
@@ -1850,6 +1856,7 @@ dberr_t log_sys_init(bool expect_no_files, lsn_t flushed_lsn,
     auto result = mtr_t::s_logging.disable(nullptr);
     /* Currently never fails. */
     ut_a(result == 0);
+    /* XB30 assert if redo log is disabled */
```
satya-bodapati: this is a leftover TODO? Can you complete it?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -52,7 +52,7 @@ bool Redo_Log_Reader::find_start_checkpoint_lsn() {
   /* Look for the latest checkpoint */
   Log_checkpoint_location checkpoint;
 
-  if (recv_find_max_checkpoint(*log_sys, checkpoint)) {
+  if (!recv_find_max_checkpoint(*log_sys, checkpoint)) {
```
satya-bodapati: really? the definitions changed to opposite? can you please double check?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -301,7 +344,7 @@ ssize_t Redo_Log_Reader::scan_log_recs(byte *buf, bool is_last, lsn_t start_lsn,
     }
   }
 
-  *read_upto_lsn = scanned_epoch_no;
+  *read_upto_lsn = scanned_lsn;
```
satya-bodapati: Looks like you have fixed an issue that I pointed in earlier commit

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -869,84 +912,27 @@ void Archived_Redo_Log_Monitor::thread_func() {
...
     return (false);
   }
 
+  ut_a(log_sys != nullptr);
+
   recv_sys_create();
   recv_sys_init();
 
   /*XB30 srv_n_log_files */
+  /*
```
satya-bodapati: Remove commented out code

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -502,6 +504,7 @@ extern TYPELIB innodb_flush_method_typelib;
 #include "sslopt-vars.h"
 
 bool redo_catchup_completed = false;
+Log_format original_log_format;
```
satya-bodapati: @altmannmarcelo I think we have another duplicate called log_detected_format. Can we reuse this?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -5121,6 +5138,12 @@ static bool xtrabackup_init_temp_log(void) {
   }
 
   log_format = (Log_format)mach_read_from_4(log_buf + LOG_HEADER_FORMAT);
+  original_log_format = log_format;
+  if (log_format == Log_format::VERSION_8_0_28) {
+    // We interpret v5 as v6 by writting proper start lsn header later in this
```
satya-bodapati: s/writting/writing

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6182,6 +6216,31 @@ static bool xtrabackup_close_temp_log(bool clear_flag) {
...
+          [](const char *path, const char *file_name) {
+            if (strcmp(file_name, ".") == 0 || strcmp(file_name, "..") == 0) {
+              return;
+            }
+            const auto to_remove =
+                std::string{path} + OS_PATH_SEPARATOR + file_name;
+            unlink(to_remove.c_str());
+          },
+          true)) {
+    ib::error(ER_IB_CLONE_STATUS_FILE)
```
satya-bodapati: Better to not use CLONE_STAUTS_FILE tag?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -491,13 +494,22 @@ bool Redo_Log_Writer::close_logfile() {
   return (true);
 }
 
+/* set encryption info of redo log on io request */
+static void set_encryption_info(IORequest &req_type) {
+  auto encryption_metadata = log_sys->m_encryption_metadata;
```
satya-bodapati: The famous 'auto' bug. I remember :)

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -138,6 +139,14 @@ static bool reopen_log_files() {
   log.m_log_flags = log_flags;
   log.m_log_uuid = log_uuid;
   log.m_files = std::move(files);
+  auto file = log.m_files.find(log_get_checkpoint_lsn(log));
```
satya-bodapati: can we also do such a check when we first open the file for checkpoint_lsn? (this code is for re-open scenario)

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -139,15 +143,8 @@ static bool reopen_log_files() {
...
-  if (file == log.m_files.end()) {
-    xb::error() << "Cannot find file with LSN " << log_get_checkpoint_lsn(log)
-                << ". Required file has been deleted by server.";
-    xb::error() << "Please consider increasing your innodb_redo_log_capacity "
-                   "or using redo log archiving.";
-    return (false);
-  }
-  log_encryption_read(log);
+  auto file = log.m_files.front();
+  log_encryption_read(log, file);
```
satya-bodapati: Better to check return value from encryption read? is it guaranteed that we always read encryption info?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1165,10 +1154,10 @@ bool Redo_Log_Data_Manager::stop_at(lsn_t lsn, lsn_t checkpoint_lsn) {
...
   archived_log_monitor.stop();
 
-  /* check if redo log are disabled */
-  if (!reader.validate_redo_log_file()) {
-    return (false);
-  }
+  /* to ensure redo logs are not disabled during the backup, reopen the log
+  files to read HEADER */
+  reopen_log_files();
+  log_crash_safe_validate(*log_sys);
```
satya-bodapati: if validation fails, we should return error?

storage/innobase/xtrabackup/test/t/xb_stats.sh
```diff
@@ -17,15 +17,25 @@ xtrabackup --datadir=$mysql_datadir --prepare --apply-log-only --target-dir=$top
...
+xtrabackup --copy-back --datadir=$mysql_datadir --target-dir=$topdir/backup
+
+start_server
+innodb_wait_for_flush_all
+mysql -e "set global innodb_fast_shutdown=0"
+shutdown_server
+
+vlog " checking Stats on data directory "
+
+xtrabackup --stats --datadir=$mysql_datadir
```
satya-bodapati: This was not the original intention of the test. It looks like stats should be allowed on prepared datadir. I think you should ad run_cmd_expect failure with xtrabckup cribbing about missing redo log files on the prepared dir.

we will add it as limitation

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/pxb-2710-predict_memory.sh
```diff
@@ -6,6 +6,7 @@ require_debug_pxb_version
 MYSQLD_EXTRA_MY_CNF_OPTS="
 innodb-flush-log-at-trx-commit=0
 sync-binlog=0
+innodb-redo-log-capacity=512M
```
satya-bodapati: if you run the same test against 8.0.29 server, this might fail

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_incremental_compressed.inc
```diff
@@ -164,7 +164,7 @@ incremental_sample`
 
   rm -rf $mysql_datadir
   xtrabackup --datadir=$mysql_datadir --copy-back \
```
satya-bodapati: hey we were trying to do copy-backf from wrong dir, how did this even work before? :)

storage/innobase/log/log0encryption.cc
```diff
@@ -121,12 +125,21 @@ dberr_t log_encryption_read(log_t &log, const Log_file &file) {
...
         return DB_ERROR;
       }
+      log_files_update_encryption(log, e_m);
+      // TODO: is this required? Since the key & iv are from dumped_file, we
+      // cannot write the same block as is. We have to recreate the header lock
+      // with the new key&iv Currently this saved buffer is used only on
+      // checkpoints and when checkpoint moves to new redo file. Will this
+      // happen in PXB?
+      // std::memcpy(log.m_encryption_buf, log_block_buf,
+      // OS_FILE_LOG_BLOCK_SIZE);
```
satya-bodapati: with use_dumped_tablesapce_keys, we write the key as unencrypted, and --generate-master-key should take care of reencrypting the tablespace key.

we can remove this TODO

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/ib_part_include.sh
```diff
@@ -34,15 +34,18 @@ ib_part_restore $TEST_VAR_ROOT/remote $mysql_datadir
...
-xtrabackup --stats --tables='test.test$' --datadir=$topdir/backup
 
-COUNT=`xtrabackup --stats --tables='test.test$' --datadir=$topdir/backup \
+ib_part_assert_checksum $checksum_a
+
+mysql -e "set global innodb_fast_shutdown=0"
+shutdown_server
+
+xtrabackup --stats --tables='test.test$' --datadir=$mysql_datadir
+COUNT=`xtrabackup --stats --tables='test.test$' --datadir=$mysql_datadir \
```
satya-bodapati: you are changing from backup dir to server MySQL datadir

So PXB 8.0.30 is not able to read stats from backup dir. This can be added as restriction and the test can fail.

Or we can decide to fix later.

Please do not change to server datadir , keep the original testcaes as is (ie --datadir=topdir/backup)

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.h
```diff
@@ -353,6 +354,12 @@ class Redo_Log_Data_Manager {
   /** error flag. */
   std::atomic<bool> error;
 
+  /** redo log consumer */
+  Redo_Log_Consumer redo_log_consumer;
+
+  /** MySQL connection to register redo log consumer */
+  MYSQL *redo_log_consumer_cnx = nullptr;
```
satya-bodapati: can't we use existing mysql connection?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1096,6 +1113,20 @@ void Redo_Log_Data_Manager::copy_func() {
...
+    if (xtrabackup_register_redo_log_consumer) {
+      if (archived_log_monitor.is_ready() &&
+          archived_log_state == ARCHIVED_LOG_POSITIONED) {
+        redo_log_consumer.deinit(redo_log_consumer_cnx);
+        mysql_close(redo_log_consumer_cnx);
+        xtrabackup_register_redo_log_consumer = false;
+      } else {
+        if (consumer_lsn != reader.get_scanned_lsn()) {
+          consumer_lsn = reader.get_scanned_lsn();
+          redo_log_consumer.advance(redo_log_consumer_cnx, consumer_lsn);
```
satya-bodapati: we are already advancing above in the copy_func(). Can you please explain why do we need this and possibly add as comment?

how do we deal with archive log that is not yet positioned to desired LSN but has started scanning and doing skip_for_block(LSN)?

storage/innobase/xtrabackup/src/redo_log_consumer.cc
```diff
@@ -0,0 +1,47 @@
+/******************************************************
+Copyright (c) 2019,2022 Percona LLC and/or its affiliates.
```
satya-bodapati: only 2022 is sufficient

storage/innobase/xtrabackup/src/redo_log_consumer.h
```diff
@@ -0,0 +1,51 @@
+/******************************************************
+Copyright (c) 2019,2022 Percona LLC and/or its affiliates.
```
satya-bodapati: 2022 is sufficient (and all other places when new file is created)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1574,6 +1577,15 @@ Disable with --skip-innodb-checksums.",
      &opt_rocksdb_wal_dir, &opt_rocksdb_wal_dir, 0, GET_STR_ALLOC, REQUIRED_ARG,
      0, 0, 0, 0, 0, 0},
 
+    {"register_redo_log_consumer", OPT_REGISTER_REDO_LOG_CONSUMER,
+     "Register a redo log consumer in the start of the backup. If this option "
+     "is enabled, it will block the server from purging redo log if PXB redo "
+     "follow thread is still copying it, causing stall in writes on the "
```
satya-bodapati: still copying it and will stall DMLs on the server

satya-bodapati (changes requested):

storage/innobase/log/log0files_io.cc
```diff
@@ -1167,6 +1167,10 @@ static dberr_t log_check_file(const Log_files_context &ctx, Log_file_id file_id,
     ib::error(ER_IB_MSG_LOG_FILE_IS_EMPTY, file_path.c_str());
     return DB_ERROR;
   }
+
+  /* PXB logs redo log doesn't care about server redo log size
```
satya-bodapati: PXB redo logs do not care about server redo log size restrictions and PXB has a single redo file and it can be > 4G size

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/suites/keyring/pxb-1942.sh
```diff
@@ -1,13 +1,16 @@
...
+. inc/common.sh
+. inc/keyring_file.sh
 
 require_server_version_higher_than 8.0.16
 
 mkdir $TEST_VAR_ROOT/b
 
 start_server --innodb-redo-log-archive-dirs=":$TEST_VAR_ROOT/b"
 
+mysql -e "set global innodb_redo_log_encrypt=on"
```
satya-bodapati: Please add a testcase that tests without redo log encryption

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6170,10 +6170,14 @@ static bool xtrabackup_replace_log_file(const char *redo_path,
...
-    return false;
+    xb::error() << "Failed to find log files in redo directory ";
+    goto error;
   }
 
-  ut_ad(listed_files.size() == 1);
+  if (listed_files.size() != 1) {
+    xb::error() << "Found more than one file in redo directory ";
+    goto error;
+  }
```
satya-bodapati: we can remove this function altogether.. there shouldn't be any rename back.

satya-bodapati (changes requested):

satya-bodapati (changes requested):
Do we need to support gcc 7.1??  Doesn't cmake enforce a GCC version check?

storage/innobase/log/log0files_io.cc
```diff
@@ -1032,6 +1037,9 @@ dberr_t log_create_unused_file(const Log_files_context &ctx,
 
   const auto file_path = log_file_path_for_unused_file(ctx, file_id);
 
+  ib::info() << "Creating redo log file at " << file_path << " with file_id "
+             << file_id << " with size " << size_in_bytes << " bytes";
+
```
satya-bodapati: I wonder if these messages are too detailed. (a question to self) :)

storage/innobase/log/log0files_capacity.cc
```diff
@@ -184,7 +184,13 @@ void Log_files_capacity::initialize(const Log_files_dict &files,
   update_exposed(
       hard_logical_capacity_for_physical(m_current_physical_capacity));
 #else
-  m_target_physical_capacity = m_current_physical_capacity = LOG_CAPACITY_MAX;
+  if (recv_recovery_is_on()) {
+    m_target_physical_capacity = m_current_physical_capacity =
+        log_files_size_of_existing_files(files);
+  } else {
+    m_target_physical_capacity = m_current_physical_capacity =
+        srv_redo_log_capacity;
```
satya-bodapati: a comment would be nice here

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2169,6 +2169,8 @@ static bool innodb_init_param(void) {
 
   srv_force_recovery = (ulint)innobase_force_recovery;
 
+  dblwr::g_mode = dblwr::Mode::OFF;
```
satya-bodapati: I would rather set it in modules that do not need to be ON.
like stats, prepare, backup for now.

Reason is, we might want to allow dblwr to be ON for prepares.

satya-bodapati: This is common to all functionalities, so avoid it here

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6240,17 +6222,25 @@ redo directory
...
+        << "Unable to validate the redo log files at the end of prepare";
+    return (true);
+  }
+
+  if (xb_generated_redo && !clear_flag) {
+    xb::info() << "prepare generated new redo log files. Skipping rename. ";
+    return (false);
   }
 
   if (xtrabackup_logfile_is_renamed) {
```
satya-bodapati: Hmm, so it looks like we keep xtrbackup_logfile in incremental backup dirs and also in case of full backups that didn't generated new redo files.

When the last prepare is done and new redo files are generated, we should avoid rename back. This should be fixed.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -869,84 +912,27 @@ void Archived_Redo_Log_Monitor::thread_func() {
...
-
-  ut_a(fil_node_create(name, srv_log_file_size / univ_page_size.physical(),
-                       *log_space, false, false) != NULL);
-
-  return (DB_SUCCESS);
-}
-
 bool Redo_Log_Data_Manager::init() {
+  /*XB30 srv_redo_log_capacity */
+  srv_redo_log_capacity = srv_redo_log_capacity_used = 1024 * 1024 * 1024;
```
satya-bodapati: can you please what happens when you remove this? Because this was fixed for prepare.

In this case, it is backup and we just need to read the redo logs.

satya-bodapati (changes requested):

sql/CMakeLists.txt
rahulmalik87: Ibd2sdi is present in runtime_output_directory but during installation we exclude it.  The behavior is the same in 8.0.29 and 8.0.30. I am trying to find out how we do that. Will update

storage/innobase/log/log0log.cc
rahulmalik87: this is formatted by clang.

storage/innobase/log/log0log.cc
rahulmalik87: Fixed in later commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yup. :)

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yes we are doing that in the latest commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: Fixed by Marcelo in his 8.0.29 Branch

storage/innobase/xtrabackup/test/t/xb_stats.sh
rahulmalik87: We have that in xb_stats.sh
`if ! grep -q "Cannot create redo log files" $OUTFILE
then
    die "Cannot find the expected error message from xtrabackup --stats"
fi`

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: I figured it out. We do not use checkpoint_no to determine which checkpoint_lsn to consider.
We just take the highest.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: ignore, I think we are still renaming on incremental backup apply.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Fixed by later commit

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: fixed by later commit

storage/innobase/xtrabackup/test/t/pxb-2710-predict_memory.sh
rahulmalik87: Fixed by Marcelo in 8.0.29 branch

cmake/mysql_version.cmake
satya-bodapati: ignore

storage/innobase/xtrabackup/src/backup_mysql.cc
satya-bodapati: @rahulmalik87 @altmannmarcelo when do we use this .cnf file?

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: it didn't make any sense in 8.0.29 too. Better to remove the definition in  log_sys struct @altmannmarcelo

storage/innobase/include/log0recv.h
rahulmalik87: fixed

storage/innobase/log/log0recv.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: not present in the latest commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: not present in latest commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: not present in latest commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: fixed in latest commit.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yes. 
`/** Find the latest checkpoint (check all existing redo log files).
  @param[in,out]  log             redo log
  @param[out]     checkpoint      the latest checkpoint found (if any)
  @return true iff any checkpoint has been found */`

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: not present in latest code.

storage/innobase/xtrabackup/src/xtrabackup.cc

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: not present in the latest commit.

storage/innobase/xtrabackup/src/redo_log.h
rahulmalik87: @altmannmarcelo ^

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: @altmannmarcelo

storage/innobase/log/log0chkp.cc
rahulmalik87: not present in the latest commit.

storage/innobase/log/log0recv.cc
rahulmalik87: @altmannmarcelo

storage/innobase/xtrabackup/src/redo_log_consumer.cc
satya-bodapati: @altmannmarcelo ^

storage/innobase/xtrabackup/src/redo_log_consumer.h
satya-bodapati: @altmannmarcelo

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: @altmannmarcelo ^

sql/CMakeLists.txt
satya-bodapati: @rahulmalik87 may be there is SKIP_INSTALL. Please grep for it.

rahulmalik87: thanks @satya-bodapati . so that's how we skip install
`MYSQL_ADD_EXECUTABLE(ibd2sdi
  ibd2sdi.cc
  COMPONENT Server
  DEPENDENCIES GenError
  LINK_LIBRARIES mysys innodb_zipdecompress ${LZ4_LIBRARY}
  SKIP_INSTALL
  )`

satya-bodapati: @rahulmalik87 please verify after install, only xtrabackup,xbcloud,xbstream binaries are present

rahulmalik87: bin contains
`libprotobuf-lite.3.19.4.dylib	xbcloud				xbcrypt				xtrabackup
libprotobuf.3.19.4.dylib	xbcloud_osenv			xbstream`
which is same as PXB 8.0.29

storage/innobase/include/log0recv.h
rahulmalik87: removed the ifdef. Moved the declaration to log0recv.h

storage/innobase/log/log0recv.cc
rahulmalik87: removed the ifdef

storage/innobase/include/log0log.h
rahulmalik87: fixed.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: code not present in the latest commit.

storage/innobase/log/log0files_io.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/test/suites/keyring/pxb-1942.sh
rahulmalik87: added.

storage/innobase/log/log0files_capacity.cc
rahulmalik87: @satya-bodapati  :) Please addressed the review comment by @satya-bodapati

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: Should I replace
srv_redo_log_capacity = srv_redo_log_capacity_used = 0; Is it fine?

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: I am also disabling it inside storage/innobase/log/log0log.cc. 
`
} else if (str_starts_with(creator_name, LOG_HEADER_CREATOR_PXB)) {
      dblwr::g_mode = dblwr::Mode::OFF;
      dblwr::set();

      if (srv_read_only_mode) {
        ib::error() << "Cannot restore from xtrabackup, InnoDB running in "
                       "read-only mode!";
        return DB_ERROR;
      }
      `
      
      so the prepare is disabled here as well.

storage/innobase/log/log0log.cc
rahulmalik87: we use log_detected_format for mach_write_compressed and few other methods in mach0data.ic

storage/innobase/include/log0files_io.h
rahulmalik87: moved

storage/innobase/log/log0recv.cc
rahulmalik87: Should I create a bug to fix it? so we can fix other occurrence as well.

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: no orignal_log_format and log_detected_format is different. 
`
 log_format = (Log_format)mach_read_from_4(log_buf + LOG_HEADER_FORMAT);
  original_log_format = log_format;
  if (log_format == Log_format::VERSION_8_0_28 ||
      log_format == Log_format::VERSION_8_0_19) {
    // We interpret v4 & v5 as v6 by writing proper start lsn header later in
    // this function
    log_format = Log_format::VERSION_8_0_30;
  }
  log_detected_format = log_format;
  `

storage/innobase/log/log0encryption.cc
rahulmalik87: @satya-bodapati I am not sure If I understand fully. Can you please add/remove the appropriate comment?

storage/innobase/xtrabackup/test/t/ib_part_include.sh
rahulmalik87: updated the test case.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: @rahulmalik87 why do it here? if we do at the top of prepare, isn't it sufficient?

rahulmalik87: yeah, It got duplicated. I will remove it from log0log.cc

storage/innobase/log/log0encryption.cc
satya-bodapati: @rahulmalik87 right, this one is on me :)

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: Please check if it works fine. During backup, we initialize redo subsystem only for LSN parsing/etc. We don't generate redo at all from backup

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: @satya-bodapati we use them at --prepare. We needed this config before to create the correct number of files on the second time we start innodb during --prepare. We don't create the files anymore (we delegate it to the server that will restore the backup) so its safe to remove this.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: basic test is fine. But will analyze jenkin as well.

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: @satya-bodapati same here, this is no longer necessary.

storage/innobase/xtrabackup/test/t/xb_incremental_compressed.inc
rahulmalik87: fixed in latest commit

storage/innobase/log/log0log.cc
satya-bodapati: Hmm, we use it for mach_read_compressed(). Good point @rahulmalik87

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: We use `original_log_format` to store what was the original format.
at open tmp logfile we adjust it's header for it to be interpreted as v6 otherwise crash recovery will complain that we are trying to run a recovery (not clean shutdown) on and older version and later when we close the log file, we re-write `original_log_format` back to its header. 

We also use the original format at log_sys_init and at --prepare for deciding if we will interpret the 8th byte of each block as LOG_BLOCK_EPOCH_NO or LOG_BLOCK_CHECKPOINT_NO

storage/innobase/log/log0log.cc
satya-bodapati: and  mach_write_compressed() incase we write redo log during trx rollback etc.

storage/innobase/xtrabackup/src/redo_log.h
altmannmarcelo: @satya-bodapati since this will happen very often (on each redo scan) and on a dedicated thread, in order to be thread safe, we would require to have a mutex on the original connection so it could be shared between multiple threads. I thought it would be safe to just open a new dedicated connection for it.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: @altmannmarcelo This is wrong and its a bug. log_detected_format should be redo log version. Not the version modified for 8030 prepare

I suggest to remove one of them.

storage/innobase/xtrabackup/src/redo_log.h
satya-bodapati: @altmannmarcelo ok to keep it as is. Thanks for the explanation.

storage/innobase/xtrabackup/test/t/ib_part_include.sh
satya-bodapati: :+1: marking it as resolved

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Thanks. Mark it as resolved once you have addressed it

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: Ok, marking it as resolved

storage/innobase/log/log0recv.cc
satya-bodapati: keep it as is for now. No need to creat IF_N_XB(). we will do it next time. Droping this

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: @satya-bodapati we cannot run prepare if log_detected_format != v6 (this var comes from server code).
If we do not store the original format, we loose this information later for places where we need to switch between LOG_BLOCK_EPOCH_NO versus LOG_BLOCK_CHECKPOINT_NO.

I don't see a problem on having both.

rahulmalik87: mach_write_compressed handles all versions. so we can store the actual version in log_detected_format and use that instead of orignal_log_format.

altmannmarcelo: @satya-bodapati  @rahulmalik87  As per our slack conversation:

log_detected_format & orignal_log_format are duplicates. We can combine them into a single var that will hold the original log version.

log_version has to be set to V6.

log_detected_format is not been set at --backup, we should set it as part of `log_file_header_deserialize` .

Will do it once we merge .29 PR here.

satya-bodapati: Agree @altmannmarcelo

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: @satya-bodapati on copy_func we advance consumer twice:

* once we start the thread (where we basically set the lower boundary to checkpoint LSN as we did not start to copy anything yet.
* inside the while loop that continues to copy redo log blocks to advance the lower boundary to the last copied lsn from copy_once

> how do we deal with archive log that is not yet positioned to desired LSN but has started scanning and doing skip_for_block(LSN)?

While we have not fully switched to archive ( archived_log_monitor.is_ready() is false & archived_log_state != ARCHIVED_LOG_POSITIONED) we still hold the lower boundary to the last copied lsn from reader.get_scanned_lsn(). Once we fully switched to archive log, we deinit redo log consumer as its safe to let the server purge redo log now as archive component will itself be registered as a redo log consumer.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: @satya-bodapati  Done

storage/innobase/xtrabackup/src/redo_log_consumer.h
altmannmarcelo: @satya-bodapati Done

storage/innobase/xtrabackup/src/redo_log_consumer.cc
altmannmarcelo: @satya-bodapati Done

storage/innobase/log/log0recv.cc
altmannmarcelo: @satya-bodapati @rahulmalik87 I don't know why this change shows up here. We have it as a pointer already on current 8.0 - https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/log/log0recv.cc#L3777

Regarding doxygen, we don't modify the value of max_memory, it still only input parameter

storage/innobase/log/log0recv.cc
altmannmarcelo: @rahulmalik87 I see that you left it at its original place in storage/innobase/include/log0recv.h. Is that the intended outcome?

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: @satya-bodapati  removed

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: Thanks, I was a bit confused with ARCHIVED_LOG_MATCHED. POSITIONED ensured that we have found the desired LSN in the archive redo file.

storage/innobase/log/log0recv.cc
satya-bodapati: ok to ignore then

storage/innobase/log/log0files_io.cc
satya-bodapati: Keeping them as is. We can remove later

storage/innobase/log/log0files_capacity.cc
satya-bodapati: Looks like I already added at a later commit

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: Looks l like rewrote this part

storage/innobase/log/log0recv.cc
rahulmalik87: no invalid ifdef present in storage/innobase/include/log0recv.h.

storage/innobase/log/log0recv.cc
```diff
@@ -3857,7 +3857,7 @@ bool meb_scan_log_recs(
      * check redo_log.cc read_log_seg_pre8030 & scan_log_recs_pre8030.
      */
     if (scanned_lsn + data_len > recv_sys->scanned_lsn) {
-      if (xtrabackup_original_log_format >= Log_format::VERSION_8_0_30 &&
+      if (log_detected_format >= Log_format::VERSION_8_0_30 &&
```
satya-bodapati: @altmannmarcelo just a minor suggestion. To maintain clarity that the variable is introduced by pxb, we should use xb_ prefix.
xb_log_format or xb_log_detected_format?

Then seeing it any place we know that it is our variable. Even if it is in server code.

We can enforce this rule for new vars that go around and sit in innodb/server code.

As of now, this suggestion is optional.

satya-bodapati: btw, I was the one who introduced this var. Didn't give much thought back then

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -5116,15 +5115,15 @@ static bool xtrabackup_init_temp_log(void) {
...
-  log_format = (Log_format)mach_read_from_4(log_buf + LOG_HEADER_FORMAT);
-  xtrabackup_original_log_format = log_format;
+  log_format =
+      static_cast<Log_format>(mach_read_from_4(log_buf + LOG_HEADER_FORMAT));
+  log_detected_format = log_format;
   if (log_format == Log_format::VERSION_8_0_28 ||
       log_format == Log_format::VERSION_8_0_19) {
     // We interpret v4 & v5 as v6 by writing proper start lsn header later in
     // this function
     log_format = Log_format::VERSION_8_0_30;
```
satya-bodapati: How modifying this in-memory variable will fix log_files_validate_format() to not throw error on old redo formats? There, it will get format from the file. isn't it?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -5116,15 +5115,15 @@ static bool xtrabackup_init_temp_log(void) {
...
   }
-  log_detected_format = log_format;
```
satya-bodapati: few lines below we have this line:
```
if (log_format < Log_format::VERSION_8_0_1) {
```
This becomes invalid if we use log_format. Because it is saved as v6 always.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6199,8 +6198,7 @@ static bool xtrabackup_replace_log_file(const char *redo_path,
   memset(log_buf + LOG_HEADER_CREATOR, ' ', 4);
 
   /* restore original log format */
-  mach_write_to_4(log_buf + LOG_HEADER_FORMAT,
```
satya-bodapati: if we are not changing format in header, do we need to restore at all?

satya-bodapati (changes requested):

storage/innobase/include/log0recv.h
satya-bodapati: @rahulmalik87 why there is ifdef XTRABACKUP around this? For server sources, moved around, there is no need to put this ifdef tag?

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: What fixes log_files_validate_format is the if below:
```
if (log_format == Log_format::VERSION_8_0_28 ||
      log_format == Log_format::VERSION_8_0_19) {
    // We interpret v4 & v5 as v6 by writing proper start lsn header later in
    // this function
    log_format = Log_format::VERSION_8_0_30;
    ```

Comment (satya-bodapati) [reply storage/innobase/xtrabackup/src/xtrabackup.cc]: ah, we rewrite header then?

File: storage/innobase/xtrabackup/src/xtrabackup.cc (hunk 57 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/xtrabackup.cc]: @satya-bodapati in fact we are writing the new format on log header at xtrabackup_init_temp_log:

```
  mach_write_to_4(log_buf + LOG_HEADER_FORMAT, to_int(log_format));

  /* write start_lsn header */
  start_lsn = ut_uint64_align_down(max_lsn, OS_FILE_LOG_BLOCK_SIZE);
  mach_write_to_8(log_buf + LOG_HEADER_START_LSN, start_lsn);

  update_log_temp_checkpoint(log_buf, max_lsn);

  success = os_file_write(write_request, src_path, src_file, log_buf, 0,
                          LOG_FILE_HDR_SIZE);
```

satya-bodapati: @altmannmarcelo it is better to avoid this rewriting header version. start_LSN is OK. I would rather suppress check that does empty redo check.

satya-bodapati: I have suppressed checks in 8.0.30 which are not valid for pre 8.0.30 files like
1. max file size to be 4G 
2. min start_lsn to be > LOG_START_LSN (8.0.29 and older wrotes 0xc which is lesser than 2048(LOG_START_LSN)

storage/innobase/log/log0recv.cc
satya-bodapati: Thanks

satya-bodapati (approved):
@altmannmarcelo please paste jenkins link here

altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/238/
https://pxb.cd.percona.com/view/PXB%208.0/job/pxb80-single-platform-run/69/

altmannmarcelo:
Failing jenkins tests have been tested locally on same distro. All passed.

altmannmarcelo (approved):

PR #1378, opened 2022-10-12
PXB 8.0.30 merge  Disallow stats for lower version

rahulmalik87:


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4896,8 +4896,21 @@ static void xtrabackup_stats_func(int argc, char **argv) {
...
+  auto log_files_8_30_pre_ctx =
+      Log_files_context{srv_log_group_home_dir, Log_files_ruleset::PRE_8_0_30};
+  ut::vector<Log_file_id> listed_files_8_30;
+  if (log_list_existing_files(log_files_8_30_pre_ctx, listed_files_8_30) ==
+      DB_SUCCESS) {
+    xb::error() << "This version of Xtrabackup Cannnot not take statsistics  "
+                   "for version below 8.0.30";
+    exit(EXIT_FAILURE);
+  }
+  */
```
satya-bodapati: isn't it better if we check for the number of files?

storage/innobase/log/log0recv.cc
```diff
@@ -4236,7 +4236,7 @@ dberr_t recv_recovery_from_checkpoint_start(log_t &log, lsn_t flush_lsn,
       ut_ad(0);
       recv_sys->found_corrupt_log = true;
       return DB_ERROR;
-}
+  }
```
satya-bodapati, line 4239: remove this

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4896,8 +4896,21 @@ static void xtrabackup_stats_func(int argc, char **argv) {
...
+
   /* Check if the log files have been created, otherwise innodb_init()
   will crash when called with srv_read_only == true */
+  /*
```
satya-bodapati: we should retain this. without redo log files, we will error out like before.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Hmm didn't realize that this code is commented :scream:  shouldn't this be enabled?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: missed to remove the comments. fixed

storage/innobase/log/log0recv.cc
rahulmalik87: If I remove clang-format complains. So let it be?

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: Fixed.

storage/innobase/log/log0recv.cc
satya-bodapati: sure, if it is by clang-format, please keep it

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4896,6 +4896,16 @@ static void xtrabackup_stats_func(int argc, char **argv) {
...
+  auto log_files_8_30_pre_ctx =
+      Log_files_context{srv_log_group_home_dir, Log_files_ruleset::PRE_8_0_30};
+  ut::vector<Log_file_id> listed_files_8_30;
+  if (log_list_existing_files(log_files_8_30_pre_ctx, listed_files_8_30) ==
+      DB_SUCCESS) {
+    xb::error() << "This version of Xtrabackup Cannnot not take statsistics  "
+                   "for version below 8.0.30";
+    exit(EXIT_FAILURE);
+  }
+
```
altmannmarcelo: This if seems like a duplicate from the one below (same information can be extracted there).
I would rather adjust below error message to also inform that it is not compatible with pre 8.0.30.

FYI: For :
```auto log_files_ctx =
      Log_files_context{srv_log_group_home_dir, Log_files_ruleset::CURRENT};
      const dberr_t err = log_list_existing_files(log_files_ctx, listed_files);
```
`err` will be `DB_SUCCESS` even if `listed_files` == 0, so the error condition is not correct.

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: Fixed, Removed the duplicate code. now checking for additional condition listed_files> 0

altmannmarcelo: @rahulmalik87 , cosmetic request.

> you need a clean copy of the database version higher than 8.0.29
change to
you need a clean copy of the database version 8.0.30 or higher

altmannmarcelo (changes requested):
@rahulmalik87 , cosmetic request.

> you need a clean copy of the database version higher than 8.0.29

change to

> you need a clean copy of the database version 8.0.30 or higher

altmannmarcelo (approved):
LGTM

PR #1395, opened 2022-11-15
Fixed PXB-2890 - Xtrabackup 8.0 no longer compresses compressed table…

altmannmarcelo:
…s which is space inefficient

https://jira.percona.com/browse/PXB-2890

Problem:

InnoDB has two levels of compression:

Table compression - where InnoDB compress the data so it can use a smaller page_size for that table.

```
Ex: CREATE TABLE tb1 (ID INT PRIMARY KEY) ROW_FORMAT=COMPRESSED
```

Page compression - Where InnoDB compress the data, and uses punch hole to release empty blocks in the end of the page. InnoDB keeps track of the size of the data withing the page in its header and system page size (UNIV_PAGE_SIZE) - data size = size of the hole.

```
Ex: CREATE TABLE tb1 (ID INT PRIMARY KEY) COMPRESSION="lz4"
```

Page Compression brings some challenges as the sparse mapping will be lost after decompression.
Currently we skip compressing both, table compression and page compression tables.

We can do decompression on xtrabackup and xbstream backups.

Xtrabackup uses compression algorithm cli (zstd / lz4) to decompress files, which makes it impossible for us to do a in-place restore of sparseness.

Currently, xbstream is a standalone tool, which has no visibility or understanding of InnoDB concepts of page / compression / flags ...

Fix:

Created a functionality to restore sparseness after decompress by reading all compressed table page header, checking if it was compressed with Page compression and punching hole in the empty space after page data.
Moved datafile_open / datafile_close / datafile_read / datafile_cur_t / xb_fil_cur_result_t to file_utils so the same (more generic) function can be used on xtrabackup & xbstream.

Some InnoDB layer function are now linked to xbstream ( fsp_header_get_flags / FSP_FLAGS_GET_PAGE_SSIZE / ...)

Created a dedicated suite for --compress tests


storage/innobase/include/mach0data.ic
```diff
@@ -139,7 +139,9 @@ bytes. The most significant byte is at the lowest address.
 @param[in]      b       pointer to 4 bytes to read
 @return 32 bit integer */
 static inline uint32_t mach_read_from_4(const byte *b) {
+#ifndef XTRABACKUP
```
satya-bodapati: this doesn't seem right. Who is passing nullptr? can't we fix the caller? Also if it is nullptr, the below code is disaster.

satya-bodapati (commented):
Can you please fix the JIRA ticket with the old and new behavior after this logic? Maybe add it as a commit message?

Do we not decompress if xbstream uses zstd library/binary?

storage/innobase/include/mach0data.ic
altmannmarcelo: it's not that someone is passing a nullptr. The problem is that we don't have low level call such as ut_a or ut_ad.
We could potentially change it:
```
#ifdef XTRABACKUP
#ifdef UNIV_DEBUG
assert(b);
#endif
#else
ut_a(b);
#endif
```

satya-bodapati: ah, ut_() is the problem :D never guessed it will be suppressed because of that :D

altmannmarcelo:
@satya-bodapati 
> Do we not decompress if xbstream uses zstd library/binary?

xbstream uses the ds_decompress_ALGORITHM.cc functions.
For xbstream we call the restore sparseness once we get an EOF type of chunk

satya-bodapati:
> @satya-bodapati
> 
> > Do we not decompress if xbstream uses zstd library/binary?
> > xbstream uses the ds_decompress_ALGORITHM.cc functions.
> > For xbstream we call the restore sparseness once we get an EOF type of chunk

so we are fully back to 2.4 behaviour? ie restore spareness in all cases (with --decompress)?

storage/innobase/include/mach0data.ic
satya-bodapati: in release build, it can be empty. because ut_ad() is debug only

altmannmarcelo:
> > @satya-bodapati
> > > Do we not decompress if xbstream uses zstd library/binary?
> > > xbstream uses the ds_decompress_ALGORITHM.cc functions.
> > > For xbstream we call the restore sparseness once we get an EOF type of chunk
> 
> so we are fully back to 2.4 behaviour? ie restore spareness in all cases (with --decompress)?

Yes and no :) . 2.4 does not have the concept of sparse map (sparseness is lost with compression).

8.0 for compressed tables now will decompress a single table which will loose sparseness then in case of page header showing the table is compressed, it will read all page headers and punch a whole in the required free space. 

So we are compressing compressed tables as 2.4, but actually also restoring the sparseness on decompression(which 2.4 does not do).

storage/innobase/include/mach0data.ic
altmannmarcelo: right. the preprocessor macro is XTRABACKUP && UNIV_DEBUG

satya-bodapati: I mean, the "#else " part is not necessary and ut_a() will be a problem too right?

altmannmarcelo: correct, we don't include files that ut_a or ut_ad are defined. If we want to include them, there is a bunch of other unnecessary dependencies that will have to be resolved.
I would rather go with the preprocessor macros for this one.

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -100,3 +108,160 @@ int mkdirp(const char *pathname, int Flags, myf MyFlags) {
...
+  size_t page_size = 0;
+
+  if (!file_has_suffix("ibd", src_file_path)) return true;
+
+  if (!datafile_open(src_file_path, &cursor, false, buffer_size)) {
+    strcpy(error, "Cannot open file");
+    return false;
+  }
+
+  while ((res = datafile_read(&cursor)) == XB_FIL_CUR_SUCCESS) {
```
satya-bodapati: is this variable 'res' necessary?
Also it is better to avoid assignments in if and while conditions.

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -100,3 +108,160 @@ int mkdirp(const char *pathname, int Flags, myf MyFlags) {
...
+        if (compressed_len < page_size) {
+#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE
+          int ret =
+              fallocate(cursor.fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE,
+                        offset + compressed_len, page_size - compressed_len);
+          if (ret != 0) {
+            strcpy(error, "fallocate returned ");
+            std::string err = std::to_string(errno);
+            strcat(error, err.c_str());
+            return false;
```
satya-bodapati: we can rewrite it a bit using a lambda or std::function?
Essentially, we iterate the page, and on every page, call the functor with the page. The functor will then do whatever it wants to do with the page.

satya-bodapati (changes requested):

satya-bodapati:
> > > @satya-bodapati
> > > > Do we not decompress if xbstream uses zstd library/binary?
> > > > xbstream uses the ds_decompress_ALGORITHM.cc functions.
> > > > For xbstream we call the restore sparseness once we get an EOF type of chunk
> > 
> > 
> > so we are fully back to 2.4 behaviour? ie restore spareness in all cases (with --decompress)?
> 
> Yes and no :) . 2.4 does not have the concept of sparse map (sparseness is lost with compression).
> 
> 8.0 for compressed tables now will decompress a single table which will loose sparseness then in case of page header showing the table is compressed, it will read all page headers and punch a whole in the required free space.
> 
> So we are compressing compressed tables as 2.4, but actually also restoring the sparseness on decompression(which 2.4 does not do).

Ah, so even if there are page IO compressed tables in 5.7, 2.4 xbstream never had the concept of sparse_map?

altmannmarcelo:
> > > > @satya-bodapati
> > > > > Do we not decompress if xbstream uses zstd library/binary?
> > > > > xbstream uses the ds_decompress_ALGORITHM.cc functions.
> > > > > For xbstream we call the restore sparseness once we get an EOF type of chunk
> > > 
> > > 
> > > so we are fully back to 2.4 behaviour? ie restore spareness in all cases (with --decompress)?
> > 
> > 
> > Yes and no :) . 2.4 does not have the concept of sparse map (sparseness is lost with compression).
> > 8.0 for compressed tables now will decompress a single table which will loose sparseness then in case of page header showing the table is compressed, it will read all page headers and punch a whole in the required free space.
> > So we are compressing compressed tables as 2.4, but actually also restoring the sparseness on decompression(which 2.4 does not do).
> 
> Ah, so even if there are page IO compressed tables in 5.7, 2.4 xbstream never had the concept of sparse_map?

Correct, f1fc1db7492a6e6e2044d2f14652e2a0adc1fc4d was only added to 8.0

satya-bodapati:
> > > > > @satya-bodapati
> > > > > > Do we not decompress if xbstream uses zstd library/binary?
> > > > > > xbstream uses the ds_decompress_ALGORITHM.cc functions.
> > > > > > For xbstream we call the restore sparseness once we get an EOF type of chunk
> > > > 
> > > > 
> > > > so we are fully back to 2.4 behaviour? ie restore spareness in all cases (with --decompress)?
> > > 
> > > 
> > > Yes and no :) . 2.4 does not have the concept of sparse map (sparseness is lost with compression).
> > > 8.0 for compressed tables now will decompress a single table which will loose sparseness then in case of page header showing the table is compressed, it will read all page headers and punch a whole in the required free space.
> > > So we are compressing compressed tables as 2.4, but actually also restoring the sparseness on decompression(which 2.4 does not do).
> > 
> > 
> > Ah, so even if there are page IO compressed tables in 5.7, 2.4 xbstream never had the concept of sparse_map?
> 
> Correct, [f1fc1db](https://github.com/percona/percona-xtrabackup/commit/f1fc1db7492a6e6e2044d2f14652e2a0adc1fc4d) was only added to 8.0

Thanks for clarifying marce.

storage/innobase/include/mach0data.ic
altmannmarcelo: Fixed

storage/innobase/xtrabackup/src/CMakeLists.txt
```diff
@@ -203,6 +204,7 @@ TARGET_LINK_LIBRARIES(xbstream
   mysys
   crc
   ${LZ4_LIBRARY}
+  mysqlclient
```
satya-bodapati: why do we have to link mysqclient to xbstream?

storage/innobase/xtrabackup/src/fil_cur.cc
```diff
@@ -57,20 +57,7 @@ static bool xb_get_zip_size(const char *file_name, pfs_os_file_t file,
...
-    if (!ret) {
-      return (false);
-    }
-    if (Compression::is_compressed_page(buf + page_size.physical())) {
-      is_compressable = false;
-    } else {
-      is_compressable = true;
-    }
-  }
+  is_compressable = FSP_FLAGS_GET_ENCRYPTION(flags) ? false : true;
```
satya-bodapati: A tablespace can be encrypted and PageIOcompressed at same time. will this avoid doing punch holes on such tablespaces?

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -516,6 +516,30 @@ static void *extract_worker_thread_func(void *arg) {
       ctxt->filehash.erase(entry->path);
       file_entry_free(entry);
       pthread_mutex_unlock(ctxt->mutex);
+      /*
+       * no need for mutex here. At this point, we are guarantee that all other
+       * threads have completed its work with this file
+       */
+      if (opt_decompress && ctxt->ds_ctxt->fs_support_punch_hole &&
+          (is_compressed_suffix(chunk.path) ||
+           is_encrypted_and_compressed_suffix(chunk.path))) {
```
satya-bodapati, line 528: will try to do punch hole on non-InnoDB files? they will have the same extension? :thinking:

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/fil_cur.cc
altmannmarcelo: encrypted tables are not compressed during backup. That is what this part of code validates (if this table will be entitle to be compressed or not).

For example, we test it at https://github.com/altmannmarcelo/percona-xtrabackup/blob/PXB-2890-8.0/storage/innobase/xtrabackup/test/inc/page_compression_common.sh#L35-L38 that gets called by https://github.com/altmannmarcelo/percona-xtrabackup/blob/PXB-2890-8.0/storage/innobase/xtrabackup/test/suites/compression/page_compression_zstd.sh

storage/innobase/xtrabackup/src/xbstream.cc
altmannmarcelo: Non InnoDB files will fail later at restore_sparseness function:

https://github.com/altmannmarcelo/percona-xtrabackup/blob/PXB-2890-8.0/storage/innobase/xtrabackup/src/file_utils.cc#L218

storage/innobase/xtrabackup/src/CMakeLists.txt
altmannmarcelo: ut0ut.h requires mysqld_Error.h which only gets generates by mysqlclient target:

```
In file included from /work/pxb/src/8.0/storage/innobase/include/ut0cpu_cache.h:33,
                 from /work/pxb/src/8.0/storage/innobase/include/ut0counter.h:42,
                 from /work/pxb/src/8.0/storage/innobase/include/sync0sync.h:45,
                 from /work/pxb/src/8.0/storage/innobase/include/sync0types.h:39,
                 from /work/pxb/src/8.0/storage/innobase/include/univ.i:546,
                 from /work/pxb/src/8.0/storage/innobase/include/fsp0fsp.h:36,
                 from /work/pxb/src/8.0/storage/innobase/xtrabackup/src/file_utils.cc:11:
/work/pxb/src/8.0/storage/innobase/include/ut0ut.h:72:10: fatal error: mysqld_error.h: No such file or directory
   72 | #include "mysqld_error.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [storage/innobase/xtrabackup/src/CMakeFiles/xbstream.dir/build.make:167: storage/innobase/xtrabackup/src/CMakeFiles/xbstream.dir/file_utils.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....

```

satya-bodapati: Please avoid this at all cost. I think, we can have InnoDB Page parser class. it is OK to duplicate for parsing purpose.

altmannmarcelo: fixed

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2607,6 +2516,14 @@ bool decrypt_decompress_file(const char *filepath, uint thread_n) {
       }
     }
   }
+  if (ds_data->fs_support_punch_hole) {
+    char error[512];
+    if (!restore_sparseness(dest_filepath, opt_read_buffer_size, error)) {
+      xb::warn() << "restore_sparseness failed: " << error;
```
satya-bodapati: include filename in error message and add that file size may increase

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -693,6 +603,7 @@ bool copy_file(ds_ctxt_t *datasink, const char *src_file_path,
     } else {
       if (ds_write(dstfile, cursor.buf, cursor.buf_read)) goto error;
     }
+    xtrabackup_io_throttling();
```
satya-bodapati, line 606: I believe this is because we decoupled datafile_read and throttling?
Can you please carefully check other callers data_file_read() that would need this additional throttling function?

storage/innobase/xtrabackup/src/backup_copy.cc
satya-bodapati: or is the filename displayed by caller error handler?

storage/innobase/xtrabackup/src/fil_cur.cc
satya-bodapati: with so many compressions around, would be nice to document what this variable is for
1. key_block_size=, record level compression
2. compression='zlib' page level IO compression
3. backup compression, --compress

:)

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -100,3 +147,166 @@ int mkdirp(const char *pathname, int Flags, myf MyFlags) {
 
   return (-1);
 }
+
+const char *get_relative_path(const char *path) {
```
satya-bodapati, line 151: xb_get_relative_path() should be moved instead? or is that gone?

storage/innobase/xtrabackup/src/file_utils.h
```diff
@@ -35,4 +77,97 @@ bool is_path_separator(char c);
...
+  Reads file content into cursor->buf.
+
+  @param [in/out] cursor     cursor containing file metadata.
+
+  @return false in case of error, true otherwise
+*/
+xb_fil_cur_result_t datafile_read(datafile_cur_t *cursor);
+
+/**
+  Restore sparseness of a file.
```
satya-bodapati: A little more documentation please. When do we use this? on which files? Is this function used only on datafiles backuped with --compress option?

satya-bodapati: ie, only on --decompress + IBD with a suffix?

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -516,6 +516,28 @@ static void *extract_worker_thread_func(void *arg) {
...
+        char path[FN_REFLEN] = {0};
+        memcpy(path, chunk.path, strlen(chunk.path));
+        unsigned short int qpress_offset = is_qpress_file(path) ? 1 : 0;
+
+        if (is_compressed_suffix(path))
+          path[strlen(path) - 4 + qpress_offset] = 0;
+        if (is_encrypted_and_compressed_suffix(path))
+          path[strlen(path) - 12 + qpress_offset] = 0;
```
satya-bodapati: possible to avoid hardcoded numbers?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: done

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: No, missed it here. I did add it on xbstream. 
Fixed

storage/innobase/xtrabackup/src/backup_copy.cc
altmannmarcelo: We only use it on copy_file and restore_Sparseness

storage/innobase/xtrabackup/src/fil_cur.cc
altmannmarcelo: Adjusted comment on the header file we declare is_compressable.

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: problem with xb_get_relative_path is that it is related to innodb, as it requires fil_space_t as parameter. To have it as part of xbcloud then we need to again add all the inner innodb layer code.

storage/innobase/xtrabackup/src/file_utils.h
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbstream.cc

storage/innobase/xtrabackup/src/fil_cur.cc
satya-bodapati: is it committed?

storage/innobase/xtrabackup/src/file_utils.cc
satya-bodapati: ok, lets not bring more innodb dependencies then.

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -100,3 +147,166 @@ int mkdirp(const char *pathname, int Flags, myf MyFlags) {
...
+    if (cursor.buf_offset == cursor.buf_read) {
+      const uint32_t flags = fsp_header_get_flags(cursor.buf);
+      const ulint ssize = FSP_FLAGS_GET_PAGE_SSIZE(flags);
+      if (ssize == 0) {
+        page_size = XB_UNIV_PAGE_SIZE_ORIG;
+      } else {
+        page_size = ((XB_UNIV_ZIP_SIZE_MIN >> 1) << ssize);
+      }
+    }
+
```
satya-bodapati, line 306: Can you please add assertions or checks here on page_size validity?
ie. page_size should fall be min 1k and max 64k page size

storage/innobase/xtrabackup/src/file_utils.h
```diff
@@ -18,8 +18,50 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
...
+typedef byte page_t;
+
+/** Compressed page */
+constexpr size_t XB_FIL_PAGE_COMPRESSED = 14;
+
+/** Compressed and Encrypted page */
+constexpr size_t XB_FIL_PAGE_COMPRESSED_AND_ENCRYPTED = 16;
+
+/** Size after compression (u16) */
+constexpr uint32_t XB_FIL_PAGE_COMPRESS_SIZE_V1 = 32U;
```
satya-bodapati, line 58: it seems a bit consistent that InnoDB used size_t for some and unit32_t for others? :thinking:

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/file_utils.h
altmannmarcelo: yeah, seems like the use different types, FIL_PAGE_COMPRESSED_AND_ENCRYPTED is declared at storage/innobase/include/fil0fil.h and uses size_t while FIL_PAGE_COMPRESS_SIZE_V1 is declared at storage/innobase/include/fil0types.h and uses uint32_t

satya-bodapati (approved):

satya-bodapati:
LGTM! with this PR we will be better than pxb 2.4 :+1: :clap:

satya-bodapati:
@altmannmarcelo side note:
it looks like we can force an int to a bool (non-zero to true and zero to false). Not sure if this is a good practice but I did find innodb doing it :) Am not asking you change in PR, just an observation

```
  /** Check if the tablespace is encrypted.
  @return true if encrypted, false otherwise. */
  [[nodiscard]] bool is_encrypted() const noexcept {
    return FSP_FLAGS_GET_ENCRYPTION(flags);
  }

```

PR #1396, opened 2022-11-16
PXB-2886 Merge 8.0.31 into PXB 

rahulmalik87:
Merge MySQL 8.0.31 in PXB

 https://jira.percona.com/browse/PXB-2886


storage/innobase/xtrabackup/test/t/bug1717158.sh
```diff
@@ -48,8 +48,8 @@ function run_test()
         --safe-slave-backup-timeout=15 \
         --target-dir=$topdir/backup
 
-    stop_server_with_id $master_id
-    stop_server_with_id $slave_id
+    shutdown_server_with_id $master_id
```
altmannmarcelo, line 51: Do you know which change brought this error ?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2367,8 +2367,10 @@ void check_dump_innodb_buffer_pool(MYSQL *connection) {
  * @param[in]   connection  MySQL connection handler
  * @return true if tables with row versions > 0 */
 bool print_instant_versioned_tables(MYSQL *connection) {
-  /* PS not affected by INSTANT issues. Upstream only affected on 8.0.29+ */
-  if (server_flavor == FLAVOR_PERCONA_SERVER || mysql_server_version < 80029)
+  /* PS not affected by INSTANT issues. Upstream only affected between 8.0.29
+   * and 8.0.31 */
+  if (server_flavor == FLAVOR_PERCONA_SERVER || mysql_server_version < 80029 ||
+      mysql_server_version >= 80031)
```
altmannmarcelo: As agreed, we still see bugs been fixed on INSTANT. We are not considering it stable as part of 8.0.31

storage/innobase/xtrabackup/test/t/all_files.sh
```diff
@@ -25,7 +25,6 @@ diff -u <( ( ( cd $dir1; find . | grep -Ev $ign_list )
 ./xtrabackup_checkpoints
 ./xtrabackup_info
 ./xtrabackup_logfile
-./xtrabackup_master_key_id
```
altmannmarcelo, line 28: good catch!

plugin/keyring_vault/keyring_vault-test/keyring_vault-test.cc
```diff
@@ -324,7 +324,7 @@ int main(int argc, char **argv) {
   BOOST_SCOPE_EXIT(void) { mysql_mutex_destroy(&LOCK_keys_in_keyring); }
   BOOST_SCOPE_EXIT_END
 
-  system_charset_info = &my_charset_utf8_general_ci;
```
altmannmarcelo, line 327: > The default MySQL server character set and collation are utf8mb4 and utf8mb4_0900_ai_ci

shall we use the default `utf8mb4_0900_ai_ci`

sql-common/client.cc
```diff
@@ -3229,7 +3229,7 @@ static MYSQL_METHODS client_methods = {
     cli_fetch_lengths,          /* fetch_lengths */
     cli_flush_use_result,       /* flush_use_result */
     cli_read_change_user_result /* read_change_user_result */
-#ifndef MYSQL_SERVER
```
altmannmarcelo: Add comment or respective `#endif` directive

sql-common/client.cc
```diff
@@ -3346,7 +3346,7 @@ MYSQL_EXTENSION *mysql_extension_init(MYSQL *mysql [[maybe_unused]]) {
       key_memory_MYSQL, sizeof(struct MYSQL_ASYNC), MYF(MY_WME | MY_ZEROFILL)));
   /* set default value */
   ext->mysql_async_context->async_op_status = ASYNC_OP_UNSET;
-#ifdef MYSQL_SERVER
```
altmannmarcelo, line 3292: Add comment or respective #endif directive

sql/server_component/mysql_command_backend.cc
```diff
@@ -43,7 +43,7 @@ MYSQL_METHODS mysql_methods = {
     csi_connect,       csi_read_query_result, csi_advanced_command,
     csi_read_rows,     csi_use_result,        csi_fetch_row,
     csi_fetch_lengths, csi_flush_use_result,  csi_read_change_user_result,
-#if !defined(MYSQL_SERVER) && !defined(MYSQL_COMPONENT)
```
altmannmarcelo: Add comment or respective #endif directive

components/keyrings/common/operations/operations.h
```diff
@@ -631,12 +631,6 @@ class Keyring_operations {
   */
   bool generate(const meta::Metadata &metadata, const data::Type type,
                 size_t length) {
-<<<<<<< HEAD
-    Data_extension generated_data(keyring_common::data::Data{"", type});
```
altmannmarcelo, line 612: Upstream had:
```
Data_extension generated_data("", type);
``` 
We had:
```
Data_extension generated_data(keyring_common::data::Data{"", type});
```

This was introduced as part of KMIP work at https://github.com/percona/percona-xtrabackup/commit/5e1fd79c459129a429b7234cf0c5d2eb2bb165cf#diff-aa7cafebcf8bbdfbc25c73827b287795dfe88d766cda819b33a1f19d4f7cd565L612 .

We should validate why just `Data_extension generated_data("", type);` was not suficient for KMIP and check if the new `data::Data g_data(type);`  won't bring any problems.

CMakeLists.txt
```diff
@@ -2382,21 +2361,11 @@ IF(HAVE_BUILD_ID_SUPPORT AND NOT WITHOUT_SERVER)
 ENDIF()
 
 UNSET(MYSQLD_EXECUTABLE_FOR_INFO_BIN)
-IF(HAVE_BUILD_ID_SUPPORT AND NOT WITHOUT_SERVER)
+IF(HAVE_BUILD_ID_SUPPORT AND NOT WITHOUT_SERVER AND NOT WITH_XTRABACKUP)
   SET(MYSQLD_EXECUTABLE_FOR_INFO_BIN
     "-DMYSQLD_EXECUTABLE=$<TARGET_FILE:mysqld>")
 ENDIF()
 
-# Build flags: This must be postponed to the make phase.
```
altmannmarcelo: better to add `IF(NOT WITH_XTRABACKUP)` here rather than deleting. In case they change anything on those lines in the future we wont get conflicts.

storage/innobase/log/log0recv.cc
```diff
@@ -1881,48 +1793,17 @@ static byte *recv_parse_or_apply_log_rec_body(
...
+#ifdef XTRABACKUP
       /* While scaning redo logs during  backup phase a
       MLOG_INDEX_LOAD type redo log record indicates a DDL
       (create index, alter table...)is performed with
       'algorithm=inplace'. This redo log indicates that
 
-      1. The DDL was started after MEB started backing up, in which
-      case MEB will not be able to take a consistent backup and should
+      1. The DDL was started after PXB started backing up, in which
+      case PXB will not be able to take a consistent backup and should
```
altmannmarcelo, line 1812: A few lines below, we still have mentions to MEB inside a solo `#ifdef XTRABACKUP` directive

altmannmarcelo (changes requested):
@rahulmalik87 way easier to review the merge now with `conflictstyle = diff3` and meaningful commit messages.
Kudos on reworking it.

A few comments to address before its a good to go from my side.

altmannmarcelo:
@rahulmalik87  
Another point, please make sure you checked all items from Not Easy to spot issues.

storage/innobase/xtrabackup/test/t/bug1717158.sh
rahulmalik87: No Marcelo, The error message  _killed in MTS mode_ is very old. It could be we just encounter it now.

storage/innobase/xtrabackup/src/backup_mysql.cc
rahulmalik87: Dropped the patch

sql-common/client.cc
rahulmalik87: fixed

sql-common/client.cc
rahulmalik87: fixed

sql/server_component/mysql_command_backend.cc
rahulmalik87: fixed

storage/innobase/log/log0recv.cc
rahulmalik87: fixed

CMakeLists.txt
rahulmalik87: fixed

components/keyrings/common/operations/operations.h
rahulmalik87: we were explicitly creating the data object and then passing it to generated_data. I do not why we did that. But now upstream is also doing the same. That's why dropped our changes. It look similar to me

`data::Data g_data(type);
  Data_extension generated_data(g_data);`

plugin/keyring_vault/keyring_vault-test/keyring_vault-test.cc
rahulmalik87: I do not understand what would be impact. Do we need to change a similar change in xtrabackup.cc as well?

altmannmarcelo: Same here we would need to investigate. Lets keep the scope to change it mb3

components/keyrings/common/operations/operations.h
altmannmarcelo: @rahulmalik87 Check with Zsolt and Yura about this one.
Do we have a jenkins job for this merge?

rahulmalik87:
> @rahulmalik87
> Another point, please make sure you checked all items from Not Easy to spot issues.

@altmannmarcelo , I checked and none of the relevent change in 8.0.31.

altmannmarcelo:
> > @rahulmalik87
> > Another point, please make sure you checked all items from Not Easy to spot issues.
> 
> @altmannmarcelo , I checked and none of the relevent change in 8.0.31.

@rahulmalik87 I see commits from PS missing on PXB in regards to components (KMIP and KMS)
Example:
- https://github.com/percona/percona-server/commit/a489ca750dca28488bf806af6d5f3dcf056ba25a
- https://github.com/percona/percona-server/commit/bd1b6e51aab3aa85af88c0b930c270af516d6516
Those seems to have been missed on 8.0.30 release.

Also check PS 8.0.31 merge branch if they have anything new there.

Please carefully double check it.

components/keyrings/common/operations/operations.h
dutow: We changed it, because the original code was incorrect - and looks like upstream realized this too.

`Data_extension` is a template parameter with a default type `Data`. The original code only compiled in this case, a custom `Data_extension` resulted in a compilation error. This fix (both ours and upstream) fixes the issue.

rahulmalik87: > @rahulmalik87 Check with Zsolt and Yura about this one.
> Do we have a jenkins job for this merge?

No Marce. The AWS spot instances are getting killed. I think because of Thanksgiving. Started again. I have old [jenkin.](https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/212/) which doesn't contain the latest fix.  So one unknown issue pxb-1679.sh. which I saw in the feature branch also and is based on 8.0.30. 
So not new. But I can analyze pxb-1679.sh after Keep_dir -k option to see what was wrong.

storage/innobase/xtrabackup/test/suites/keyring/keyring_encryption.sh
```diff
@@ -34,6 +34,7 @@ run_cmd backup_and_restore test
 
 #for the remaining test need PS server
 require_xtradb
+require_server_version_lower_than 8.0.30
```
altmannmarcelo: We should not add a skip in the middle of a test. Split it into a separate file and add the skip in the start.

altmannmarcelo (changes requested):

rahulmalik87:
@altmannmarcelo [Jenkin](https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/254/) result for PS 8.0.30 and MS 8.0.31. I will give one more jenkin test for PS 8.0.28 and MS 8.0.29. 
keyring falilure happens when server is slow to write Encryption key in keyring file after ALTER. 
We need to modify the framework to wait until key is written. 
Need to investigate pxb-1679.sh. But I saw this failure on PXB 8.0.30 also. I will try to investigate.
Other failures are due to Dynamic metadatada.

storage/innobase/xtrabackup/test/suites/keyring/keyring_encryption.sh
rahulmalik87: I have removed the skip but added a condition to execute the remaining test for PS version than 8.0.30. Is that look fine?

altmannmarcelo: I would rather have it separate @rahulmalik87 . When reviewing test failures its clear that it was skipped or not. Does not require us to go back and remember all those logics, if it is test X and version lower than Y ...

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -179,12 +183,15 @@ lsn_t Redo_Log_Reader::read_log_seg_pre8030(log_t &log, byte *buf,
...
+    if (file != log.m_files.end()) {
+      xb::error() << "Failed to find redo file id " << file_id;
+      return 0;
+    }
 
     auto file_handle = file->open(Log_file_access_mode::READ_ONLY);
 
     if (!file_handle.is_open()) {
-      // file not found
-      m_error = true;
```
altmannmarcelo: isn't m_error still necessary here?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -228,24 +235,17 @@ lsn_t Redo_Log_Reader::read_log_seg_8030(log_t &log, byte *buf, lsn_t start_lsn,
   // update the in-memory structure log files by scanning
   if (log.m_files.find(start_lsn) == log.m_files.end()) {
     if (!reopen_log_files(start_lsn)) {
-      // file not found
-      m_error = true;
```
altmannmarcelo: isn't m_error still necessary here?

altmannmarcelo: Error message will be printed by reopen_log_files ?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -228,24 +235,17 @@ lsn_t Redo_Log_Reader::read_log_seg_8030(log_t &log, byte *buf, lsn_t start_lsn,
...
       return 0;
     }
   }
 
+  /* file should be present as ensure by reopen_log_files */
   auto file = log.m_files.find(start_lsn);
-  if (file == log.m_files.end()) {
```
altmannmarcelo: Is it guaranteed that we will find the file with this start_lsn?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -325,17 +328,25 @@ ssize_t Redo_Log_Reader::read_logfile(bool is_last, bool *finished) {
     const lsn_t end_lsn =
         read_log_seg(log, log_buf + len, start_lsn, start_lsn + RECV_SCAN_SIZE);
 
+    if (end_lsn == 0) {
+      xb::error() << "read_log_seg() failed.";
+      m_error = true;
+      return (-1);
```
altmannmarcelo: what is the difference between return 0 and -1 ? Where do we deal with return code?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -325,17 +328,25 @@ ssize_t Redo_Log_Reader::read_logfile(bool is_last, bool *finished) {
...
+
     auto size = scan_log_recs(log_buf + len, is_last, start_lsn, &scanned_lsn,
                               log_scanned_lsn, finished);
-    if (end_lsn == start_lsn) {
-      break;
-    }
 
-    if (size < 0 || end_lsn == 0) {
-      xb::error() << "read_logfile() failed.";
+    if (size < 0) {
```
altmannmarcelo: isn't end_lsn = 0 an error?

storage/innobase/xtrabackup/test/t/pxb-1679.sh
```diff
@@ -27,26 +27,17 @@ CREATE TABLE test02 (id int auto_increment primary key, a TEXT, b TEXT);
...
-        if [ $( echo $line | grep -c './test/test02.ibd') -eq 1 ]; then
-            mysql -e "TRUNCATE TABLE test02" test
-        fi
-    done
-
-if ! [ ${PIPESTATUS[0]} -eq 0 ] ; then
-    die "backup failed"
-fi
-
+xtrabackup --backup --target-dir=$topdir/backup
```
altmannmarcelo: We got rid of lock-ddl=FALSE and this is part of reproducible test case of https://jira.percona.com/browse/PXB-1679

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: Will be set by the caller `read_logfile` function.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yes. error handing is handled by `reopen_log_files`

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: It is handled. but I think it better would be to add an error message here as well. fixed it

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: so 0 means. We continue the operation -1 means some error.
same is mention in comment of `read_log`.
 @return read length or -1 if error. */

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yes it is error. Just moved it before calling `scan_log_recs`

storage/innobase/xtrabackup/test/t/pxb-1679.sh
rahulmalik87: yes, concurrent truncate is not a safe operation with lock-ddl=false. I saw [regression](https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/256/CMAKE_BUILD_TYPE=RelWithDebInfo,DOCKER_OS=debian%3Abuster,XTRABACKUP_TARGET=innodb80/testReport/junit/(root)/t_pxb-1679/sh/)

altmannmarcelo: @rahulmalik87 what I mean is that PR #573  added code to handle this situation. Why is this broken now?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -179,12 +180,15 @@ lsn_t Redo_Log_Reader::read_log_seg_pre8030(log_t &log, byte *buf,
     Log_file_id file_id = source_offset / file_size;
 
     auto file = log.m_files.file(file_id);
+    if (file != log.m_files.end()) {
```
altmannmarcelo: shouldn't this be file == log.m_files.end() ?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -225,27 +229,21 @@ lsn_t Redo_Log_Reader::read_log_seg_8030(log_t &log, byte *buf, lsn_t start_lsn,
...
   ut_a(start_lsn < end_lsn);
 
-  // update the in-memory structure log files by scanning
   if (log.m_files.find(start_lsn) == log.m_files.end()) {
-    if (!reopen_log_files(start_lsn)) {
-      // file not found
-      m_error = true;
+    // update the in-memory structure log files by scanning
+    if (!reopen_log_files(start_lsn) ||
+        log.m_files.find(start_lsn) == log.m_files.end()) {
```
altmannmarcelo: this line is unnecessary as we do exactly this in reopen_log_files. Here we are dealing with in-memory structures, which is guarantee that will not change until the next call of reopen_log_files

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: yeah, some how missed. thanks for catching.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: fixed.

storage/innobase/xtrabackup/src/redo_log.cc
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/pxb-1679.sh
rahulmalik87: I have dropped the changes and created [PXB-2948 ](https://jira.percona.com/browse/PXB-2948) for tracking.

storage/innobase/xtrabackup/test/suites/keyring/keyring_encryption.sh
rahulmalik87: fixed

storage/innobase/xtrabackup/test/t/pxb-1679.sh
altmannmarcelo: Ok, can you please start work on investigating it so it doesn't die on backlog?

altmannmarcelo (approved):
LGTM. two final requests (you can complete them without another round of review)

1. squash test fix commits
2. do a final round on jenkins once you are done addressing all reviews.


Thanks a lot for completing this merge :tada:

PR #1398, opened 2022-11-24
8.0 null merge

VarunNagaraju:


rahulmalik87 (approved):
LGTM

altmannmarcelo (changes requested):
There are too many upstream commits on this null merge. Essentially we should see only ours commit.
The thing here is that we don't have 8.0.31 merged into 8.0 branch yet, which will bring those same upstream commits.
We should wait until #1396 is merged before proceeding with this null-merge.

altmannmarcelo:
@VarunNagaraju You can re-create your null-merge and force-push here. It should shrink the commits now that 8.0.31 has been merged.

altmannmarcelo (approved):
LGTM

PR #1400, opened 2022-11-28
Fixed PXB-2928 - Xtrabackup crashes with signal 11 when taking a backup

altmannmarcelo:
https://jira.percona.com/browse/PXB-2928

Problem:
When page tracking is used, we need to build a list of tables that will
require to be fully copied. For this list we use a std::map during the
initial scan of redo logs, which is guaranteed to be single thread and
later use the operator[] to check if a given space id needs to be fully
copied. The access to this map is done in multiple threads but since
all threads only read from it this was assumed to be thread safe.
However, std::map::operator[] will create the accessed key if it does
not exists. This will lead to copy threads populating the map which can
lead to crash.

Fix:
Adjusted the map to use std::unordered_set as we only need the space_id.
This set is populated by the redo first scan phase (by a single thread)
and after the initial parsing is over, we only use it for reads. Hence,
it is not required to protect with mutex.
Lookup is done by find() operator which does not modify the content of
the set and only access for reading.

### DIFF
--- a/storage/innobase/include/log0recv.h
+++ b/storage/innobase/include/log0recv.h
@@ -59,6 +59,11 @@ struct recv_addr_t;
  * pagetracking*/
 extern std::unordered_set<space_id_t> full_scan_tables;
 
+/* count of tablespaces that will require full table scan. This is calculated
+ * after full_scan_tables is fully populated and should never changed. */
+
+extern size_t full_scan_tables_count;
+
 namespace xtrabackup {
 struct recv_sys_t {
   struct mem_block_t {
--- a/storage/innobase/include/log0recv.h
+++ b/storage/innobase/include/log0recv.h
@@ -57,7 +57,7 @@ struct recv_addr_t;
 #ifdef XTRABACKUP
 /** map of tablespace_id, that would be full scan during backup with
  * pagetracking*/
-extern std::map<space_id_t, bool> full_scan_tables;
+extern std::unordered_set<space_id_t> full_scan_tables;
 
 namespace xtrabackup {
 struct recv_sys_t {
--- a/storage/innobase/xtrabackup/src/read_filt.cc
+++ b/storage/innobase/xtrabackup/src/read_filt.cc
@@ -252,7 +252,7 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
   we do full scan full_scan_tables is populated during the first scan of redo */
 
   if (ctxt->space_id == dict_sys_t::s_dict_space_id ||
-      full_scan_tables[ctxt->space_id] == true) {
+      full_scan_tables.find(ctxt->space_id) != full_scan_tables.end()) {
     *read_batch_start = ctxt->offset;
     if (ctxt->offset >= ctxt->data_file_size) {
       *read_batch_len = 0;
--- a/storage/innobase/xtrabackup/src/read_filt.cc
+++ b/storage/innobase/xtrabackup/src/read_filt.cc
@@ -253,6 +253,7 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
 
   if (ctxt->space_id == dict_sys_t::s_dict_space_id ||
       full_scan_tables.find(ctxt->space_id) != full_scan_tables.end()) {
+    ut_ad(full_scan_tables_count == full_scan_tables.size());
     *read_batch_start = ctxt->offset;
     if (ctxt->offset >= ctxt->data_file_size) {
       *read_batch_len = 0;
@@ -260,6 +261,7 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
     }
     *read_batch_len = ctxt->data_file_size - ctxt->offset;
   } else {
+    ut_ad(full_scan_tables_count == full_scan_tables.size());
     /* if no page changed for given space return */
     if (!changed_page_tracking->count(ctxt->space_id)) {
 #ifdef UNIV_DEBUG
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -111,6 +111,8 @@ This set is populated by the redo first scan phase (by a single thread) and
 after the initial parsing is over, we only use it for reads. Hence, it is not
 required to protect with mutex. */
 std::unordered_set<space_id_t> full_scan_tables;
+
+size_t full_scan_tables_count = 0;
 #endif /* XTRABACKUP */
 
 #ifdef UNIV_HOTBACKUP
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -105,8 +105,12 @@ rolling back incomplete transactions. */
 volatile bool recv_recovery_on;
 volatile lsn_t backup_redo_log_flushed_lsn;
 #ifdef XTRABACKUP
-/** map of space_id, that would be full scan during backup with pagetracking */
-std::map<space_id_t, bool> full_scan_tables;
+/* List of tablespaces that should be copied fully. Even if page tracking is
+active, the tablespaces in this list will be fully copied to backup directory.
+This set is populated by the redo first scan phase (by a single thread) and
+after the initial parsing is over, we only use it for reads. Hence, it is not
+required to protect with mutex. */
+std::unordered_set<space_id_t> full_scan_tables;
 #endif /* XTRABACKUP */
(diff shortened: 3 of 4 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/252/

storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -252,7 +252,7 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
   we do full scan full_scan_tables is populated during the first scan of redo */
 
   if (ctxt->space_id == dict_sys_t::s_dict_space_id ||
-      full_scan_tables[ctxt->space_id] == true) {
+      full_scan_tables.find(ctxt->space_id) != full_scan_tables.end()) {
```
percona-ysorokin, line 255: @altmannmarcelo There is also `count()` method in `std::unordered_set`
```suggestion
      full_scan_tables.count(ctxt->space_id) == 1) {
```

satya-bodapati:
> https://jira.percona.com/browse/PXB-2928
> 
> Problem: When page tracking is used, we need to build a list of tables that will require to be fully copied. For this list we use a std::map during the initial scan of redo logs, which is guaranteed to be single thread and later use the operator[] to check if a given space id needs to be fully copied. The access to this map is done in multiple threads but since all threads only read from it this was assumed to be thread safe. However, std::map::operator[] will create the accessed key if it does not exists. This will lead to copy threads populating the map which can lead to crash.
> 
> Fix: Adjusted the map to use std::unordered_set which will be thread safe since the map is const at the time we access it and unordered_set will give us constant access time in average, which can also speed table lookup.

@altmannmarcelo unorded-set is not thread safe too. The safeness comes from the way the hash is used. ie
It is populated during scan phase by single thread. After it is populated, the maps is used only for reading by multiple threads. Hence we don't need a mutex. Can you please modify the github description/commit message

storage/innobase/include/log0recv.h
```diff
@@ -57,7 +57,7 @@ struct recv_addr_t;
 #ifdef XTRABACKUP
 /** map of tablespace_id, that would be full scan during backup with
  * pagetracking*/
```
satya-bodapati, line 59: Can you please fix the comment.

List of tablespaces that should be copied fully. Even if paget racking is active, the tablespaces in this list will be fully copied to backup directory. This set/map is populated by the redo scan phase (by a single thread) and after the parsing is over, we only use it for reads.  Hence, it is not required to protect with mutex.

satya-bodapati (changes requested):

altmannmarcelo:
> @altmannmarcelo unorded-set is not thread safe too. The safeness comes from the way the hash is used. ie
> It is populated during scan phase by single thread. After it is populated, the maps is used only for reading by multiple threads. Hence we don't need a mutex. Can you please modify the github description/commit message

Right, that is what I meant by saying the set is `const`  . I will adjust the commit message.

storage/innobase/include/log0recv.h
altmannmarcelo: Done

storage/innobase/xtrabackup/src/read_filt.cc
altmannmarcelo: Thanks for the suggestion @percona-ysorokin . I think that underneath this will be calling find anyway.

satya-bodapati (commented):
Looks good. Is it possible to create testcase? 
May be we can put assert that the number of elements shouldn't change after redo scan is over and without the fix, the assertion will fail?

altmannmarcelo:
> Looks good. Is it possible to create testcase? May be we can put assert that the number of elements shouldn't change after redo scan is over and without the fix, the assertion will fail?

@satya-bodapati good point. I just don't think a new test case just for this will bring any benefit. I propose we add the assertion, other test cases in the suite will fail due to it (without the fix)

altmannmarcelo:
## Test cases without the fix and with the assertion

```
🐬 marcelo  📂 /work/pxb/ins/8.0/xtrabackup-test/test  ▶ 
 ╰▶ $ ./run.sh -d /work/ps/ins/8.0/ -r /dev/shm/ -s pagetracking
Detecting server version...
Running against Percona Server 8.0.30-22-debug (XtraDB 8.0.30-22)
Using 16 parallel workers
Running suite(s): pagetracking

encryption_after_checkpoint              w1	[failed]    24
Spent 24 of 41 seconds executing testcases

SUMMARY: 11 run, 0 successful, 0 skipped, 1 failed

Failed tests:  encryption_after_checkpoint

 🐬 marcelo  📂 /work/pxb/ins/8.0/xtrabackup-test/test  ▶ 
 ╰▶ $ tail -n 50 results/encryption_after_checkpoint | head -n 26
2022-11-30T14:01:50.755298-03:00 2 [Note] [MY-011825] [Xtrabackup] Copying ./ibdata1 to /dev/shm/var/w1/var1/inc/ibdata1.delta
InnoDB: Assertion failure: read_filt.cc:264:full_scan_tables_count == full_scan_tables.size()
InnoDB: thread 139832149305088InnoDB: We intentionally generate a memory trap.
InnoDB: Submit a detailed bug report to https://jira.percona.com/projects/PXB.
InnoDB: http://dev.mysql.com/doc/refman/8.0/en/forcing-innodb-recovery.html
17:01:50 UTC - mysqld got signal 6 ;
Most likely, you have hit a bug, but this error can also be caused by malfunctioning hardware.
Thread pointer: 0x7f2d1c000b80
terribly wrong...
stack_bottom = 7f2d35965720 thread_stack 0x100000
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(my_print_stacktrace(unsigned char const*, unsigned long)+0x59) [0x55d14f58518b]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(print_fatal_signal(int)+0x27e) [0x55d14dbfad6c]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(handle_fatal_signal+0x6d) [0x55d14dbfaf66]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14420) [0x7f2d40d32420]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb) [0x7f2d3fcf400b]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x12b) [0x7f2d3fcd3859]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(my_abort()+0x11) [0x55d14f57a7e9]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(ut_dbg_assertion_failed(char const*, char const*, unsigned long)+0x151) [0x55d14e41d6e6]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(+0x31bf52c) [0x55d14d36852c]
/work/pxb/ins/8.0/xtrabackup-test/test/../../bin/xtrabackup(xb_fil_cur_read(xb_fil_cur_t*)+0x43) [0x55d14d365f97]
```

## Suite running with assertion and fix

```
 🐬 marcelo  📂 /work/pxb/ins/8.0/xtrabackup-test/test  ▶ 
 ╰▶ $ ./run.sh -d /work/ps/ins/8.0/ -r /dev/shm/ -s pagetracking
Detecting server version...
Running against Percona Server 8.0.30-22-debug (XtraDB 8.0.30-22)
Using 16 parallel workers
Running suite(s): pagetracking

encryption_after_checkpoint              w1	[passed]    37
xb_incremental_compressed_pagetracking_16kb w3	[passed]    64
xb_incremental_disable_pagetracking      w9	[passed]    64
xb_incremental_disable_enable_pagetracking w8	[passed]    65
xb_incremental_pagetracking              w11	[passed]    77
xb_incremental_pagetracking_lsn          w10	[passed]    80
xb_incremental_compressed_pagetracking_8kb w7	[passed]    81
xb_incremental_compressed_pagetracking_1kb w4	[passed]    118
xb_incremental_compressed_pagetracking_2kb w5	[passed]    146
xb_incremental_compressed_pagetracking_4kb w6	[passed]    162
validate_tracking_start_lsn              w2	[passed]    254
Spent 1148 of 272 seconds executing testcases

SUMMARY: 11 run, 11 successful, 0 skipped, 0 failed

```

satya-bodapati (approved):
Thanks for addressing all review comments.
LGTM!

PR #1366, opened 2022-08-19
PXB-2854 - Quicklz decompression memory corruption issue fix

Chaloff:
There is a memory corruption issue inside the quicklz.c source file that ships
with Percona XtraBackup. Specifically the problem happens on copying
user-supplied binary data over heap allocated memory buffers of user-controlled
size. This allows corruption of heap data structures and potential arbitrary
code execution.

The code in question is inside the qlz_decompress function of quicklz.c file:

```
size_t qlz_decompress(const char *source, void *destination, qlz_state_decompress *state)
{
        size_t dsiz = qlz_size_decompressed(source);

        if (state->stream_counter + qlz_size_decompressed(source) - 1 >= QLZ_STREAMING_BUFFER)
        {
                if((*source & 1) == 1)
                {
                        reset_table_decompress(state);
                        dsiz = qlz_decompress_core((const unsigned char *)source, (unsigned char *)destination, dsiz, state, (const unsigned char *)destination);
                }
                else
                {
                        memcpy(destination, source + qlz_size_header(source), dsiz);
                }
                state->stream_counter = 0;
                reset_table_decompress(state);
        }
        else
        {
                unsigned char *dst = state->stream_buffer + state->stream_counter;
                if((*source & 1) == 1)
                {
                        dsiz = qlz_decompress_core((const unsigned char *)source, dst, dsiz, state, (const unsigned char *)state->stream_buffer);
                }
                else
                {
                        memcpy(dst, source + qlz_size_header(source), dsiz);
                        reset_table_decompress(state);
                }
                memcpy(destination, dst, dsiz);
                state->stream_counter += dsiz;
        }
        return dsiz;
}
```

Note the first memcpy invocation: that does copy data from user-provided
compressed file into a heap-allocated buffer for which size is also controlled
by the user via the compressed file header. This allows heap corruption with
user-controlled data. Potentially this means arbitrary code execution for the
processes that utilize the vulnerable function - one example is xbstream with
—decompress flag.

Steps to reproduce:

 - Create a compressed file, e.g. with qpress from some file larger than 65535
   bytes.
 - Edit compressed file so that the four bytes at offset 8 are changed to be
   less than 0x10000, for example set to 0x1000 instead.
 - Edit the file so that the byte at offset 50 is an even value to pass the
   test: if((*source & 1) == 1)
 - Replace the bytes of actual file with some recognizable pattern, e.g. 0x41
   0x42 0x43 0x44
[...]

PR #1404, opened 2022-12-12
Fixed PXB-2854 - Quicklz decompression memory corruption issue

altmannmarcelo:
There is a memory corruption issue inside the quicklz.c source file that ships
with Percona XtraBackup. Specifically the problem happens on copying
user-supplied binary data over heap allocated memory buffers of user-controlled
size. This allows corruption of heap data structures and potential arbitrary
code execution.

The code in question is inside the qlz_decompress function of quicklz.c file:

```
size_t qlz_decompress(const char *source, void *destination, qlz_state_decompress *state)
{
        size_t dsiz = qlz_size_decompressed(source);

        if (state->stream_counter + qlz_size_decompressed(source) - 1 >= QLZ_STREAMING_BUFFER)
        {
                if((*source & 1) == 1)
                {
                        reset_table_decompress(state);
                        dsiz = qlz_decompress_core((const unsigned char *)source, (unsigned char *)destination, dsiz, state, (const unsigned char *)destination);
                }
                else
                {
                        memcpy(destination, source + qlz_size_header(source), dsiz);
                }
                state->stream_counter = 0;
                reset_table_decompress(state);
        }
        else
        {
                unsigned char *dst = state->stream_buffer + state->stream_counter;
                if((*source & 1) == 1)
                {
                        dsiz = qlz_decompress_core((const unsigned char *)source, dst, dsiz, state, (const unsigned char *)state->stream_buffer);
                }
                else
                {
                        memcpy(dst, source + qlz_size_header(source), dsiz);
                        reset_table_decompress(state);
                }
                memcpy(destination, dst, dsiz);
                state->stream_counter += dsiz;
        }
        return dsiz;
}
```

Note the first memcpy invocation: that does copy data from user-provided
compressed file into a heap-allocated buffer for which size is also controlled
by the user via the compressed file header. This allows heap corruption with
user-controlled data. Potentially this means arbitrary code execution for the
processes that utilize the vulnerable function - one example is xbstream with
—decompress flag.

Steps to reproduce:

 - Create a compressed file, e.g. with qpress from some file larger than 65535
   bytes.
 - Edit compressed file so that the four bytes at offset 8 are changed to be
   less than 0x10000, for example set to 0x1000 instead.
 - Edit the file so that the byte at offset 50 is an even value to pass the
   test: if((*source & 1) == 1)
 - Replace the bytes of actual file with some recognizable pattern, e.g. 0x41
   0x42 0x43 0x44
 - Add the file to an xbstream file: xbstream -c Demo.qp > Demo.xbstream
 - Now try to extract with decompression using xbstream under a debugger, e.g.
   gdb and observe the corruption: xbstream —decompress -x < Demo.xbstream

```
head -c 100000 </dev/urandom > payload.bin

qpress payload.bin payload.qp

ls -l payload.qp -rw-r--r-- 1 me me 100107 Feb 17 18:08 payload.qp

printf '\x00\x01\x00' | dd of=payload.qp bs=1 seek=8 count=3 conv=notrunc

printf '\x10' | dd of=payload.qp bs=1 seek=49 count=1 conv=notrunc

python -c 'import sys; sys.stdout.write("A"*100040)' | dd of=payload.qp bs=1
seek=50 count=100040 conv=notrunc

xbstream-80 -c payload.qp > corrupted.xbstream

$ xbstream-80 --decompress -x < corrupted.xbstream Segmentation fault ```
```

Fix by prevent XtraBackup read/write outside array bounds ``` We noticed that
quicklz already has a mechanism to check if the data was corrupted and prevent
reading and writing outside array bounds. Program should be compiled against
QLZ_MEMORY_SAFE flag. So we uncommented the flag and add the error message.

All new code of the whole pull request, including one or several
files that are either new files or modified ones, are contributed under the
BSD-new license. I am contributing on behalf of my employer Amazon Web Services, Inc.

(cherry picked from commit https://github.com/percona/percona-xtrabackup/commit/dac3d82f559efbf06d116c066d30e688fe3fb323)

closes #1366


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/260/

PR #1408, opened 2023-01-16
PXB-2964 : PXB needs too many open files to take backups

satya-bodapati:
Problem:
-------
PXB 8.0 requires very high open files limit (Operating system configuration ulimit -n).

Since, --lock-ddl is ON by default, the parameter --close-files is also ON implicitly. This means files are safe to be closed during backup. Even after closing fies, PXB asks for very high open files operating configuration.

This is because of a bug in PXB. It opens an extra handle to each IBD file which is never closed. Hence, the total number of open handles required in OS is equal to total number of IBD files.

Fix
---
Remove the duplicate open file handle which is never closed (only closed by exit of process)

### DIFF
--- a/storage/innobase/xtrabackup/src/fil_cur.cc
+++ b/storage/innobase/xtrabackup/src/fil_cur.cc
@@ -116,7 +116,6 @@ xb_fil_cur_result_t xb_fil_cur_open(
 {
   page_size_t page_size(0, 0, false);
   ulint page_size_shift;
-  bool success;
 
   /* Initialize these first so xb_fil_cur_close() handles them correctly
   in case of error */
@@ -143,10 +142,7 @@ xb_fil_cur_result_t xb_fil_cur_open(
   srv_close_files has an effect only on IBD tablespaces. */
   if (cursor->is_system || !srv_backup_mode ||
       (srv_close_files && cursor->is_ibd)) {
-    node->handle = os_file_create_simple_no_error_handling(
-        0, node->name, OS_FILE_OPEN, OS_FILE_READ_ONLY, srv_read_only_mode,
-        &success);
-    if (!success) {
+    if (!fil_node_open_file(node)) {
       /* The following call prints an error message */
       os_file_get_last_error(true);
 
@@ -154,8 +150,6 @@ xb_fil_cur_result_t xb_fil_cur_open(
 
       return (XB_FIL_CUR_ERROR);
     }
-
-    fil_node_open_file(node);
   }
 
   ut_ad(node->is_open);
--- a/storage/innobase/xtrabackup/test/t/pxb-1870.sh
+++ b/storage/innobase/xtrabackup/test/t/pxb-1870.sh
@@ -23,5 +23,5 @@ PARTITION BY HASH (uid)
 PARTITIONS 3000;
 EOF
 
-xtrabackup --backup --lock-ddl=false --target-dir=$topdir/backup
+xtrabackup --backup --target-dir=$topdir/backup
 xtrabackup --prepare --target-dir=$topdir/backup


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/230/

storage/innobase/xtrabackup/src/fil_cur.cc
```diff
@@ -143,9 +143,8 @@ xb_fil_cur_result_t xb_fil_cur_open(
   srv_close_files has an effect only on IBD tablespaces. */
   if (cursor->is_system || !srv_backup_mode ||
       (srv_close_files && cursor->is_ibd)) {
-    node->handle = os_file_create_simple_no_error_handling(
-        0, node->name, OS_FILE_OPEN, OS_FILE_READ_ONLY, srv_read_only_mode,
-        &success);
+    success = fil_node_open_file(node);
```
altmannmarcelo: We don't need success variable at all, we can check the return of fil_node_open_file directly on the if statement.

storage/innobase/xtrabackup/test/t/pxb-1870.sh
```diff
@@ -23,5 +23,5 @@ PARTITION BY HASH (uid)
 PARTITIONS 3000;
 EOF
 
-xtrabackup --backup --lock-ddl=false --target-dir=$topdir/backup
+xtrabackup --backup --target-dir=$topdir/backup
```
altmannmarcelo, line 26: Good catch. There is no reason to have --lock-ddl=false here.

altmannmarcelo (changes requested):
@satya-bodapati still waiting on test results to validate it doesn't introduce any regression.
Just a minor adjustment in the code.

Do you have any idea why this extra handled was opened?

storage/innobase/xtrabackup/test/t/pxb-1870.sh
satya-bodapati: The reason, with --lock-ddl=false, the file close optimization is disabled (implicitly) and then we required high open file limit in operating system.

storage/innobase/xtrabackup/src/fil_cur.cc
satya-bodapati: sure, I will optimize this part

satya-bodapati:
> @satya-bodapati still waiting on test results to validate it doesn't introduce any regression. Just a minor adjustment in the code.
> 
> Do you have any idea why this extra handled was opened?

I checked 5.7 code base(pxb 2.4). We did the entire opening of the file manually. In 8.0, it is refactored to use an existing function but forgot to remove the extra open.

storage/innobase/xtrabackup/test/t/pxb-1870.sh
altmannmarcelo: @satya-bodapati in this case, isn't keeping the number of open files high a key part of reproducing the bug PXB-1870 fixes? Perhaps we should keep lock-ddl disabled.

satya-bodapati: @altmannmarcelo PXB-1870 was not about open files, It was SDI buffers and memory allocations.

altmannmarcelo (approved):
LGTM. Do you mind squashing and re-basing on top of 8.0 as some of your tests are failing at https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/230/  .

We have already fixed it on top of 8.0

satya-bodapati:
> LGTM. Do you mind squashing and re-basing on top of 8.0 as some of your tests are failing at https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/230/ .
> 
> We have already fixed it on top of 8.0

Sure marce, will rebase and submit it to Jenkins

PR #1427, opened 2023-02-27
PXB-2901 Add Full page tracking feature in PXB

rahulmalik87:
Add option --page-tracking=incremental/full
default is incremental

Added multiple phase in PXB
1st phase is copy of idb
2nd phase is copy of redo and delta pages

Added Apply of delta to full backup

Added feature to verify skip pages in full backup

now backup need page tracking to be started already

added new condition in prepare

In test framewrok added new option -p to use pagetracking as default


altmannmarcelo:
Closing. Once this is implemented, it will result in a new PR.

PR #1429, opened 2023-03-01
PXB-2865 Failed to copy dymainc metadata

rahulmalik87:
https://jira.percona.com/browse/PXB-2865

We have the logic of skipping dynamic metadata on the apply-log phase and relying on MySQL.ibd to copy those changes. see PXB-2180 The issue comes if there are change that are not flushed to mysql.ibd before the Final incremental backup.
Final Incremental prepare will skip applying of DYNAMIC_METADATA redo and mysql.ibd will not have those changes.

PXB does not allow final prepare on backup prepared already with apply-log-only. Last incremental backup should be prepared without apply-log-only.

### DIFF
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -7104,6 +7104,14 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
   IORequest write_request(IORequest::WRITE);
 
   read_metadata();
+  if (strcmp(metadata_type_str, "log-applied") == 0 &&
+      !xtrabackup_incremental) {
+    xb::error() << "Backup is already prepared. The final incremental should "
+                   "be prepared without --apply-log-only. or if you restoring "
+                   "full backup, It should be only prepare without option "
+                   "--apply-log-only";
+    exit(EXIT_FAILURE);
+  }
 
   if (!strcmp(metadata_type_str, "full-backuped")) {
     xb::info() << "This target seems to be not prepared yet.";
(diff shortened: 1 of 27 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7104,6 +7104,14 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
   IORequest write_request(IORequest::WRITE);
 
   read_metadata();
+  if (strcmp(metadata_type_str, "log-applied") == 0 &&
+      !xtrabackup_incremental) {
+    xb::error() << "Backup is already prepared. The final incremental should "
+                   "be prepared without --apply-log-only. or if you restoring "
+                   "full backup, It should be only prepare without option "
+                   "--apply-log-only";
```
satya-bodapati, line 7112: I don't understand the intention of the second part of the message "or if you restoring.."

storage/innobase/xtrabackup/test/suites/keyring/innodb_encryption_export.sh
```diff
@@ -66,7 +66,7 @@ ${XB_BIN} --prepare --apply-log-only --incremental-dir=$topdir/inc1 \
 	  --keyring-file-data=$keyring_file_plugin \
 	  --xtrabackup-plugin-dir=${plugin_dir} ${keyring_args}
 
-${XB_BIN} --prepare --apply-log-only --incremental-dir=$topdir/inc2 \
+${XB_BIN} --prepare --incremental-dir=$topdir/inc2 \
```
satya-bodapati, line 69: is it possible to combine these two steps? (Just asking. Not asking you to change yet :))

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: PXB-2180 is fixed in 8.0.23. Is it possible to reproduce these assertions using PXB 8.0.23? (with the disable page cleaner patch)?

satya-bodapati (changes requested):

altmannmarcelo (changes requested):
I would rather work on a final solution rather than silent test cases

altmannmarcelo:
Closing. Once this is implemented, it will be done as a new PR.

PR #1432, opened 2023-03-03
PXB-2972 : InnoDB: Assertion failure: log0log.cc:1627:srv_is_being_st…

satya-bodapati:
…arted || srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS

Problem:
--------
When PXB errors out after applying redo, i.e. when we are booting InnoDB to load tables to do rollback, an error at this stage causes PXB to crash instead of a graceful exit.

PXB renames the xtrabackup_logfile to 'ib_redo0' file to apply redo and later on errors (or on success too).
On success (and if PXB hasn't generated any new redo files) and on errors, we rename back the ib_redo0 to xtrabackup_logfile

To check if any new redo files are created, PXB checks the state of redo files.

At this stage, InnoDB is de-initialized and the InnoDB APIs accessing the redo files do not expect the state to be in SRV_SHUTDOWN_NONE.

Fix:
----
Relax the assertion to make PXB exit gracefully

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1627,8 +1627,18 @@ dberr_t log_sys_init(bool expect_no_files, lsn_t flushed_lsn,
   Log_file_handle::s_on_before_read = [](Log_file_id, Log_file_type file_type,
                                          os_offset_t, os_offset_t read_size) {
     ut_a(file_type == Log_file_type::NORMAL);
+
+#ifdef XTRABACKUP
+    /* If there is an error during startup (ie error loading SDI etc), innodb
+    is not fully inited. An exit at this stage, we don't call innodb_end() and
+    the srv_shutdown_state remains SRV_SHUTDOWN_NONE. PXB opens the redo log
+    files again at the end to rename back to xtrabackup_logfile. Hence the
+    srv_shutdown_state can be SRV_SHUTDOWN_NONE */
+#endif /* XTRABACKUP */
     ut_a(srv_is_being_started ||
-         srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS);
+         srv_shutdown_state.load() ==
+             SRV_SHUTDOWN_EXIT_THREADS IF_XB(|| srv_shutdown_state.load() ==
+                                                    SRV_SHUTDOWN_NONE));
 #ifndef UNIV_HOTBACKUP
     srv_stats.data_read.add(read_size);
 #endif /* !UNIV_HOTBACKUP */
--- a/storage/innobase/xtrabackup/src/xtrabackup.cc
+++ b/storage/innobase/xtrabackup/src/xtrabackup.cc
@@ -2434,6 +2434,8 @@ dberr_t dict_load_tables_from_space_id(space_id_t space_id, THD *thd,
 
   fil_space_t *space = fil_space_get(space_id);
 
+  DBUG_EXECUTE_IF("force_sdi_error", return (DB_ERROR););
+
   if (fsp_has_sdi(space_id) != DB_SUCCESS) {
     xb::warn() << "Failed to read SDI from " << space->name;
     return (DB_SUCCESS);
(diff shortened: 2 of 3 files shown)


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/260/

storage/innobase/log/log0log.cc
```diff
@@ -1627,8 +1627,15 @@ dberr_t log_sys_init(bool expect_no_files, lsn_t flushed_lsn,
   Log_file_handle::s_on_before_read = [](Log_file_id, Log_file_type file_type,
                                          os_offset_t, os_offset_t read_size) {
     ut_a(file_type == Log_file_type::NORMAL);
+
```
altmannmarcelo, line 1630: #ifdef XTRABACKUP

storage/innobase/log/log0log.cc
```diff
@@ -1627,8 +1627,15 @@ dberr_t log_sys_init(bool expect_no_files, lsn_t flushed_lsn,
...
+    /* PXB opens the redo log files before or after innodb initalization is
+    over. This is because, on startup, we open PXB redo log and rename to
+    ib_redo. At end (on error or success), we will open redo files to rename it
+    back to xtrabackup_logfile. Hence the srv_shutdown_state can be
+    SRV_SHUTDOWN_NONE */
     ut_a(srv_is_being_started ||
-         srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS);
+         srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS ||
```
altmannmarcelo:
```suggestion
         srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS IF_XB(||...
```

altmannmarcelo (changes requested):
@satya-bodapati my only concern here is that we might be allowing pxb to initialize the log
_sys twice. We should not do it, but if we erroneously do so, the assertion will prevent that.
Is it possible for us to change the srv_shutdown_state in case of failure at dd boot?

satya-bodapati:
> @satya-bodapati my only concern here is that we might be allowing pxb to initialize the log _sys twice. We should not do it, but if we erroneously do so, the assertion will prevent that. Is it possible for us to change the srv_shutdown_state in case of failure at dd boot?
@altmannmarcelo 
this is a good question. but the most interesting part it is it has nothing to do with "log_sys_init()"

In fact, the assertion puzzled me. Why would you call log_sys_int() on redo log IO read? Isn't it surprising?

satya-bodapati:
> > @satya-bodapati my only concern here is that we might be allowing pxb to initialize the log _sys twice. We should not do it, but if we erroneously do so, the assertion will prevent that. Is it possible for us to change the srv_shutdown_state in case of failure at dd boot?
> > @altmannmarcelo
> > this is a good question. but the most interesting part it is it has nothing to do with "log_sys_init()"
> 
> In fact, the assertion puzzled me. Why would you call log_sys_int() on redo log IO read? Isn't it surprising?

Answer is, these are callbacks that registered on IO read and write but the callbacks are defined in log_sys_init()

satya-bodapati:
> > > @satya-bodapati my only concern here is that we might be allowing pxb to initialize the log _sys twice. We should not do it, but if we erroneously do so, the assertion will prevent that. Is it possible for us to change the srv_shutdown_state in case of failure at dd boot?
> > > @altmannmarcelo
> > > this is a good question. but the most interesting part it is it has nothing to do with "log_sys_init()"
> > 
> > 
> > In fact, the assertion puzzled me. Why would you call log_sys_int() on redo log IO read? Isn't it surprising?
> 
> Answer is, these are callbacks that registered on IO read and write but the callbacks are defined in log_sys_init()

I've updated the comment in the code. It is just that, if an error occurs too early in innodb initialization, innodb is not considered inited. We have not created innodb threads yet.

An exit at this stage, wouldn't call innodb_end() and so srv_shutdown_state remains NONE.

altmannmarcelo (approved):
LGTM. Please squash before merging

PR #1433, opened 2023-03-10
Fixed PXB-3013 - smart memory test fail with 0 redo log

altmannmarcelo:
https://jira.percona.com/browse/PXB-3013

Problem:
Smart memory test fails with 0 redo log because it thinks that smart memory has not ran, however, there is no redo log to be applied.

Fix:
Ensure the test generates redo log.

### DIFF
--- a/storage/innobase/xtrabackup/test/t/estimate_memory.sh
+++ b/storage/innobase/xtrabackup/test/t/estimate_memory.sh
@@ -6,7 +6,11 @@ fi
 . inc/common.sh
 require_debug_pxb_version
 
-
+function run_insert() {
+  while true ; do
+    mysql -e "INSERT INTO pxb_2710 VALUES (NULL); SELECT SLEEP(0.5)" test
+  done
+}
 # PXB-2710 - Predict memory at --prepare
 
 MYSQLD_EXTRA_MY_CNF_OPTS="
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/296/

storage/innobase/xtrabackup/test/t/estimate_memory.sh
```diff
@@ -144,6 +148,15 @@ rm -rf $mysql_datadir
 # PXB-2980 - Add a parameter to disable memory estimation during backup
 start_server
 backupdir=${topdir}/backup
+run_cmd $MYSQL $MYSQL_ARGS test <<EOF
+CREATE TABLE pxb_2710(a INT PRIMARY KEY AUTO_INCREMENT) ENGINE=InnoDB;
+INSERT INTO pxb_2710 VALUES (NULL);
```
satya-bodapati, line 153: It is still possible that before 0.5 seconds we reach the ON test. Is one insert sufficient?

altmannmarcelo: There is an insert into a loop as part of `run_insert` function. That one is the one that ensures redo log will be generated.

satya-bodapati: But I think it is a cat and mouse game marce. Cat = redo checkpoint thread mouse = pxb :smile:

altmannmarcelo: I can remove the sleep, I just did not want to introduce too much unnecessary pressure.

```
run_insert &
insert_pid=$!
```
This will be running in the background and will add 2 redo logs per second. PXB never completes in less than 1 second.

satya-bodapati (approved):

storage/innobase/xtrabackup/test/t/estimate_memory.sh
rahulmalik87: Since the test already requiredthe debug PXB binaries. Why not use debug sync to insert a record that would guarnted generated redo before the backup is complete?

altmannmarcelo: Can be done that way. Will have the same outcome as the current approach.

rahulmalik87 (approved):
LGTM

PR #1434, opened 2023-03-10
Fixed PXB-3014 - redo log consumer test fails on centos7

altmannmarcelo:
https://jira.percona.com/browse/PXB-3014

Problem:
centos7 identifies `!grep` as a command, rather than analyzing the exit status of grep.

Fix:
Add space between `!` and `grep` to make it a command.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/297/

rahulmalik87 (approved):
LGTM

PR #1435, opened 2023-03-17
PXB-2955 - Implement Dictionary cache

satya-bodapati:
PXB-2955 : Implement dictionary cache

Problem(s)
----------
1. xtrabackup --prepare requires very high ram (--use-memory doesn't help)
2. xtrabackup --prepare crashes or doesn't finish if backup dir has more than 300K tablespaces
3. xtrabackup --prepare takes long time to complete if backup dir has huge number of tablespaces (redo to be applied is small)

Before this PR(the old design)
------------------------------
xtrabackup --prepare, after applying redo logs, starts InnoDB engine to do "rollback" of uncompleted
transactions at the time of backup.

For this purpose, it has to prepare a dictionary. Rollback works by loading a table object. This is based
on "table_id" recorded in undo log record.

PXB doesn't use the same server APIs to boot DD engine. It is complicated and requires significant efforts
to load server component and initialize DD cache. PXB uses SDI (Serialized Dictionary Info) found on IBD tablespaces.

SDI is a JSON representation of the "table structure". Index information, index fields, columns, column types etc
are part of SDI.

it is possible for an user to see SDI using the below command

ibd2sdi t1.ibd > t1.sdi

PXB uses the same SDI present inside the tablespace. To build a dictionary cache (that is required for rollback). PXB
goes through all tablespaces (*.ibd files) in backup dir. For each IBD:

1. open IBD file
2. Parse SDI using SDI APIs
3. Uncompress the SDI (it is stored as zlib compressed stream)
4. Deserialize the SDI to Server table object (dd::Table)
5. Convert dd::Table to InnoDB table object (dict_table_t)

Note that we load every table in the above steps as "non-evictable". This is because InnoDB has master thread enforces
the table (dict_table_t) cache limit.

The reason for storing them as "non-evictable" is that if a table is evicted, and rollback asks to load a table using
"table_id", we don't know the "tablespace" that contains the table_id. The relation ship "table_id" to "space_id" doesnt
exist. Hence, we load the tables as "non-evictable"

This design (to not being able to evict tables) has a huge penalty. Since we have to load all tables from all IBDs,
the number of tables loaded into cache is huge. We load a table that is NOT required for rollback.

Fix (New Design):
-----------------

If we could somehow maintain the relation "table_id (key) -> space_id (value)", we can do "on-demand" loading of table

The "somehow" part:
-------------------
1. We scan mysql.indexes table to parse a DD table directly using InnoDB APIs
2. We parse the InnoDB record and extract "se_private_data"
3. A sample "se_private_data" looks like this: "id=158;root=4;space_id=6;table_id=1067;trx_id=1548;"
4. From se_private_data, extract "space_id" and "table_id". From this we know for table_id 1067, the space_id
   to look for SDi is 6.
8. For partitions we take a similar approach. we scan mysql.index_partitions and the build "table_id -> space_id" map

New Flow:
---------
1. Load all tables mysql.ibd, we need them to be able to do B-Tree scan of mysql.indexes, mysql.index_partitions, mysql.tables tables.
2. Read from mysql.indexes/index_partitions to build dictionary map (std::unordered_map) of  "table_id -> space_id"
3. On rollback, if server looks for a table_id, we first look into cache. If it doesn't exists, we load the table using the following steps.
   a. find the space_id for the given table_id
   b. Load a table_id from the above space_id (step a result)
   c. Load the table into cache
   d. Table is opened (table->n_ref_count is 1)

4. After the table object is used and undo apply is over, we close the table (table->n_ref_count is 0). Hence evictable

We save memory by
1. By not loading all tables from all tablespaces (using SDI) upfront
2. We load a table on-demand. i.e. "only" if there is rollback on the table (or on --export)

The partition conundrum:
------------------------
The above solution to use "table_id -> space_id" doesn't work with partitions. This is because Server stores SDI only in the first IBD.

Lets say, we create a partition table with 3 partitions. t1#p#p0.ibd, t1#p#p1.ibd and t1#p#p2.ibd. Server stores SDI only in t1#p#p0.ibd.
Other partitions t1#p#p1.ibd and t1#p#p2.ibd do not have the dd::Table SDI. Only dd::Tablespace SDI is present in p1 and p2.

This means, if ever there is rollback on p1 and p2, we cannot use the p1.ibd or p2.ibd. We have to look always into p0.ibd (or the first partition)

To solve this problem, we first have to identify all partition of a table. In the above example, we have to find the group is

t1#p#p0.ibd, t1#p#p1.ibd and t1#p#p2.ibd all three form one table with paritions. We could do a name scan but we chose a different way to figure out
the partitions. Again, that is by relying "partition index id"

Lets get into details. How do we find "all partitions" of table?
Answer: Scan mysql.index_partitions, we can get index_partition id and the
InnoDB space_id and also the table_id->space_id relation

table_id_space_map:
{table_id -> space_id} 
1075 -> 13 (p0.ibd)
1076 -> 14 (p1.ibd)
1077 -> 15 (p2.ibd)}
 
part_id_spaces_map:
{partition_id -> space_id} 
{296 -> 13
296 -> 14
296 -> 15 }
 
space_part_map
{space_id -> partition_id } {
13 : 296
14 : 296
15 : 296 }
 
Lets say, we are asked to load table for table_id 1077.
We look into table_id_space_map and figure out that the space_id
for 1077 is 15 (p2.ibd)

Next, check if this space belongs to a "partitioned table"
For this, we look into space_part_map. If an entry exists, it means
the space_id belongs to partitioned table.

We get the partition id. For space_id 15, the partition id is 296.

Given a partition_id 296, we can easily know all the partition IBs by using
part_id_spaces_map

For 296, we now know that the IBDs are 13, 14, 15.

We will start from lower bound(13) and start looking into all space_ids in this
group. (14, 15)

We will load SDI from space_id 13 (p0.ibd)

Other improvements:
-------------------
We now use multiple threads to load IBDs into cache. Previously, only the scan is parallel.
See Tablespace_files::open_ibds()


storage/innobase/dict/dict0dd.cc
```diff
@@ -1189,7 +1188,13 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
...
   mutex_enter(&dict_sys->mutex);
 
-  dict_table_add_to_cache(table, false);
+  // Second parameter of dict_table_add_to_cache() is to make
+  // a table evictable or not. NEVER use "false". It is meant for "DD" tables.
+  // In PXB, we handle DD tables (mysql.* tables) to be non-evictable by using
+  // table->n_ref_count. It is incremented to 1 on load and hence not evictable.
+  // The DD tables are closed at end of prepare operation. Check
+  // 'mysql_ibd_tables' in xtrabackup.cc
+  dict_table_add_to_cache(table, true);
```
rahulmalik87: Can we put a debug assert in dict_table_add_to_cache to ensuser it is never set true in PXB code path?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1231,15 +1236,33 @@ table_id_t dd_table_id_and_part(space_id_t space_id, const dd::Table &dd_table,
   return (table_id);
 }
 
-static int dd_table_load_part(table_id_t table_id, const dd::Table &dd_table,
-                              const dd::Partition *dd_part,
-                              dict_table_t *&table, THD *thd,
-                              const dd::String_type *schema_name, bool implicit,
-                              space_id_t space_id) {
+/* Convert dd::Table object to InnoDB Table object. The table can be a
```
rahulmalik87: It is PXB specific code. Can we move to #ifdef XTRABACKUP?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
-    int ret = dd_table_load_part(table_id, dd_table, dd_part, table, thd,
-                                 schema_name, implicit, space_id);
-    if (ret != 0) {
-      return ret;
+    dict_table_t *table = dd_table_load_part(table_id, dd_table, dd_part, thd,
+                                             schema_name, implicit, space_id);
+    if (table != nullptr) {
+      table_vec.push_back(table);
+    } else {
+      // Unable to load table. Error
```
rahulmalik87: Can this happen in lock-ddl=true? Will add any value to add debug assert?

storage/innobase/dict/dict0dd.cc
```diff
@@ -6344,7 +6419,7 @@ const rec_t *dd_startscan_system(THD *thd, MDL_ticket **mdl, btr_pcur_t *pcur,
   dict_index_t *clust_index;
   const rec_t *rec = nullptr;
 
-  *table = dd_table_open_on_name(thd, mdl, system_table_name, true, false);
+  *table = dd_table_open_on_name(thd, mdl, system_table_name, false, false);
```
rahulmalik87: Why are we changing to ignore_err to false?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4940,9 +4647,9 @@ static void stat_with_rec(dict_table_t *table, THD *thd,
       }
     }
   }
-  mutex_enter(&(dict_sys->mutex));
+  // mutex_enter(&(dict_sys->mutex));
```
altmannmarcelo: remove

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4886,7 +4594,6 @@ static bool xtrabackup_stats_level(dict_index_t *index, ulint level) {
 
 static void stat_with_rec(dict_table_t *table, THD *thd,
                           MDL_ticket *mdl_on_tab) {
-  mutex_exit(&(dict_sys->mutex));
```
altmannmarcelo: why the mutex is not necessary anymore?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8109,6 +7815,10 @@ int main(int argc, char **argv) {
 
   /* Logs xtrabackup generated timestamps in local timezone instead of UTC */
   opt_log_timestamps = 1;
+  /* This variable determines the size of dict_table_t* cache in InnoDB. This is
+  default and it is sufficient because rollback is single threaded and we only
+  open tables one by one */
+  table_def_size = 400;
```
altmannmarcelo: shall we add this as a parameter or inherit from open_table_cache ?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
   return (1);
 }
 
-int dd_table_load_on_dd_obj(dd::cache::Dictionary_client *client,
-                            space_id_t space_id, const dd::Table &dd_table,
-                            dict_table_t *&table, THD *thd,
-                            const dd::String_type *schema_name, bool implicit) {
-  uint64 table_id = dd_table.se_private_id();
+std::tuple<int, std::vector<dict_table_t *>> dd_table_load_on_dd_obj(
```
rahulmalik87: Can we use use  typealias `using XB_RES_TUPLE std::tuple<int, std::vector<dict_table_t *>>` or something similar for better readibility

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
-  if (table_id == dd::INVALID_OBJECT_ID) {
+  if (dd_table.se_private_id() == dd::INVALID_OBJECT_ID) {
     /* Partitioned table */
     ut_ad(dd_table_is_partitioned(dd_table));
+    auto end = dd_table.leaf_partitions().end();
+
+    if (table_id != 0) {
+      // Load specific partition based on table_id
+      auto i =
+          std::search_n(dd_table.leaf_partitions().begin(), end, 1, table_id,
```
rahulmalik87: Is it possible that dd_table.leaf_partitions can have a duplicate entry of table_id?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
+        dict_table_t *table = dd_table_load_part(
+            tid, dd_table, part, thd, schema_name, implicit, space_id);
+        if (table != nullptr) {
+          table_vec.push_back(table);
+        } else {
+          // Unable to load table. Error
+          return std::make_tuple(1, table_vec);
+        }
       }
-      dd_table_close(table, thd, nullptr, false);
```
rahulmalik87: We are not doing dd_table_close now in case of an error.

storage/innobase/dict/dict0dd.cc
```diff
@@ -1738,7 +1791,29 @@ dict_table_t *dd_table_open_on_id(table_id_t table_id, THD *thd,
...
+      // There are other paths (stats, export) that load all tables at once
+      // from tablespace (they dont ask for specific table_id). Here we always
+      // for a specific "table_id" from space_id SDI
+      auto result = xb::prepare::dict_load_tables_using_table_id(table_id);
+      dberr_t err = std::get<0>(result);
+      std::vector<dict_table_t *> vec = std::get<1>(result);
+      if (err == DB_SUCCESS)
+        if (vec.size() != 1) {
+          // Too many elements? We requested only one and
+          // it is unexpected behaviour. Return nullptr
```
rahulmalik87: Can we add some warning or message here?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
-                                   implicit, space_id);
-      if (ret != 0) {
-        return ret;
+        uint64 tid = part->se_private_id();
+        dict_table_t *table = dd_table_load_part(
+            tid, dd_table, part, thd, schema_name, implicit, space_id);
+        if (table != nullptr) {
+          table_vec.push_back(table);
+        } else {
+          // Unable to load table. Error
```
rahulmalik87: Can we add some error/warning here?

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
-void Tablespace_files::open_ibds() const {
-  for (auto path : m_ibd_paths) {
-    for (auto name : path.second) {
-      fil_open_for_xtrabackup(m_dir.path() + name,
-                              name.substr(0, name.length() - 4));
+/** Open IBD tablespaces and load them into cache fil_space_t*
+@param[in]  start   Start of slice
+@param[in]  end   End of slice
+@param[in]  thread_id Thread ID */
+void Tablespace_files::open_ibd(const Const_iter &start, const Const_iter &end,
```
rahulmalik87: open_ibd name looks confusing. Here we are still opening multiple ibds.

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
+               << " tablespaces";
   }
 }
 
+/** Open all known tablespaces. */
+void Tablespace_files::open_ibds() {
+  using std::placeholders::_1;
+  using std::placeholders::_2;
+  using std::placeholders::_3;
+  using std::placeholders::_4;
```
rahulmalik87: [This](std::placeholders::_4;) is unused?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1707,7 +1760,7 @@ dict_table_t *dd_table_open_on_id(table_id_t table_id, THD *thd,
 
 reopen:
   if (ib_table == nullptr) {
-    if (skip_missing) {
+    if (!xb::prepare::table_exists_in_dd(table_id) && skip_missing) {
```
rahulmalik87: Can we add an warning or error here?

storage/innobase/dict/dict0dd.cc
```diff
@@ -6491,7 +6559,7 @@ const char *dd_process_dd_tables_rec_and_mtr_commit(
   mtr_commit(mtr);
   THD *thd = current_thd;
 
-  *table = dd_table_open_on_id(table_id, thd, mdl, true, false, true);
```
rahulmalik87: Why are we changing dict_locked?

storage/innobase/fil/fil0fil.cc
```diff
@@ -376,8 +376,12 @@ struct Char_Ptr_Compare {
 class Tablespace_files {
  public:
   using Names = std::vector<std::string, ut::allocator<std::string>>;
-  using Paths = std::unordered_map<space_id_t, Names>;
+  /* Use std::map instead of std::unorded_map because we want to use
```
rahulmalik87: These changes are specific to PXB. Let us discuss how to handle them @satya-bodapati  @altmannmarcelo , Do we need #IFDEF XTRABACKUP or something similar to know these are changes done in PXB only?

storage/innobase/fil/fil0fil.cc
```diff
@@ -459,7 +463,14 @@ class Tablespace_files {
   }
 
   /** Open all known tablespaces. */
-  void open_ibds() const;
```
rahulmalik87: Why removed const?

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
+@param[in]  thread_id Thread ID */
+void Tablespace_files::open_ibd(const Const_iter &start, const Const_iter &end,
+                                size_t thread_id) const {
+  size_t count = 0;
+  bool printed_msg = false;
+  auto start_time = std::chrono::steady_clock::now();
+
+  for (auto it = start; it != end; ++it) {
+    for (auto name : it->second) {
+      // TODO: remove the retcode check??
```
rahulmalik87: Please address this TODO.

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,786 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+std::tuple<dberr_t, std::vector<dict_table_t *>>
+dict_load_tables_using_table_id(table_id_t table_id) {
+  auto it = table_id_space_map.find(table_id);
+  if (it == table_id_space_map.end()) {
+    // Table_id not present in any space_id. A dropped table
+    return {DB_TABLESPACE_NOT_FOUND, {}};
+  }
+
+  space_id_t space_id = it->second;
+  xb::info() << "space_id is " << space_id << " for table_id: " << table_id;
```
rahulmalik87: This will print lot of information to user which might not be useful.

storage/innobase/dict/dict0dd.cc
```diff
@@ -1738,7 +1791,29 @@ dict_table_t *dd_table_open_on_id(table_id_t table_id, THD *thd,
...
+      // we load all tables from that the space_id. Typically, we load single
+      // table at a time.
+      //
+      // There are other paths (stats, export) that load all tables at once
+      // from tablespace (they dont ask for specific table_id). Here we always
+      // for a specific "table_id" from space_id SDI
+      auto result = xb::prepare::dict_load_tables_using_table_id(table_id);
+      dberr_t err = std::get<0>(result);
+      std::vector<dict_table_t *> vec = std::get<1>(result);
+      if (err == DB_SUCCESS)
```
rahulmalik87: please add {} after if.

rahulmalik87 (changes requested):

storage/innobase/dict/dict0dd.cc
```diff
@@ -1189,7 +1188,13 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
 
   mutex_enter(&dict_sys->mutex);
 
-  dict_table_add_to_cache(table, false);
+  // Second parameter of dict_table_add_to_cache() is to make
```
altmannmarcelo: Can we move this description to the function signature?

storage/innobase/dict/dict0dd.cc
altmannmarcelo: I like the idea, but I don't know how to differentiate it from PXB vs other parts of the code. something like IF_PXB() will always compile when compiling PXB

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
+    dd::cache::Dictionary_client *client, space_id_t space_id,
+    const dd::Table &dd_table, table_id_t table_id, THD *thd,
+    const dd::String_type *schema_name, bool implicit) {
+  std::vector<dict_table_t *> table_vec;
 
-  if (table_id == dd::INVALID_OBJECT_ID) {
+  if (dd_table.se_private_id() == dd::INVALID_OBJECT_ID) {
     /* Partitioned table */
     ut_ad(dd_table_is_partitioned(dd_table));
+    auto end = dd_table.leaf_partitions().end();
```
altmannmarcelo: what datatype is dd_table.leaf_partitions() ?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
+          // Unable to load table. Error
+          return std::make_tuple(1, table_vec);
+        }
       }
-      dd_table_close(table, thd, nullptr, false);
     }
-    return 0;
   } else {
+    // Load a NON-Partition table
+    if (table_id == 0) {
```
altmannmarcelo: Can you refresh my memory why we would overwrite table_id parameter in case its 0?

storage/innobase/dict/dict0dd.cc
altmannmarcelo: This return type is a bit confusing. From what I can see, we create a dict_table_t* vector named table_vec, but we only push_back a single table to it, so we don't seem to need a vector. 
I would rather return `dict_table_t *` here, and the caller is responsible to validate if its nullptr OR receive a `dict_table_t *`  and return int as success or failure.

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,36 +1316,72 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
-      uint64 tid = part->se_private_id();
-      int ret = dd_table_load_part(tid, dd_table, part, table, thd, schema_name,
-                                   implicit, space_id);
-      if (ret != 0) {
-        return ret;
+        uint64 tid = part->se_private_id();
+        dict_table_t *table = dd_table_load_part(
+            tid, dd_table, part, thd, schema_name, implicit, space_id);
+        if (table != nullptr) {
+          table_vec.push_back(table);
```
altmannmarcelo: On places where we find the table, I would rather return right away. We will not fall into any other code branch, just direct to `{0, table_vec};`  having the return right away when its the only path makes it easier to read.

storage/innobase/fil/fil0fil.cc
altmannmarcelo: Ideally yeah, they should be withing `ifdef XTRABACKUP` on shared code.

storage/innobase/fil/fil0fil.cc
```diff
@@ -11452,10 +11520,11 @@ dberr_t fil_open_for_xtrabackup(const std::string &path,
   return (DB_SUCCESS);
 }
 
-/** Open IBD tablespaces.
+/** Open IBD tablespaces and load them to cache
 @param[in]  start   Start of slice
 @param[in]  end   End of slice
-@param[in]  thread_id Thread ID */
+@param[in]  thread_id Thread ID
+@param[in]  result  result of the open */
```
altmannmarcelo:
```suggestion
@param[in,out]  result  result of the open */
```

storage/innobase/dict/dict0dd.cc
satya-bodapati: Added

storage/innobase/dict/dict0dd.cc
satya-bodapati: Done

storage/innobase/dict/dict0dd.cc
satya-bodapati: Let me think. Let's say DDL is in progress, new indexes added, instant columns added etc.
1. ib_sdi_set() is called from DDL, which would delete mark current SDI record, and a new SDI entry is added.
2. changes to mysql.ibd is also in the same transaction.

Let's assume PXB copies a tablespace that is under DDL, we could have missed those un-flushed SDI changes but these changes are 'redo-logged' and undo-logged as well. 

The initial scan of mysql.ibd will be up to date physically because the redo phase is over. The only problem is DDL to be rolled back, then the mysql.ibd scan might have seen extra dd_index/column entries.

So if we do dictionary trx rollback before user trx rollback, we should be OK but this is a big grey area..

I am OK with adding debug assert for now.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Agree, doing it

storage/innobase/dict/dict0dd.cc
satya-bodapati: Its not ignore_err, its dict_locked. We dont need to lock the dict_sys mutex in a big loop. 
when a table is loaded, the mutex is taken. later it is ref_counted. 
once the table usage is over, we do dict_table_close() or dd_table_close().

Then it makes the table eligible for eviction.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Not possible by Server DD design

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
+  bool printed_msg = false;
+  auto start_time = std::chrono::steady_clock::now();
+
+  for (auto it = start; it != end; ++it) {
+    for (auto name : it->second) {
+      // TODO: remove the retcode check??
+      dberr_t err = fil_open_for_xtrabackup(m_dir.path() + name,
+                                            name.substr(0, name.length() - 4));
+
+      if (err != DB_SUCCESS && err != DB_TABLESPACE_EXISTS &&
```
altmannmarcelo: Where we got the list of allowed errors from?

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
+      xb::info() << "Thread# " << thread_id << " - Opened " << count << "/"
+                 << std::distance(start, end) << " tablespaces";
+
+      start_time = std::chrono::steady_clock::now();
+
+      printed_msg = true;
+    }
+  }
+
+  if (printed_msg) {
```
altmannmarcelo: Shouldn't we print the below message regardless of this thread already printed the 10 seconds message?

storage/innobase/fil/fil0fil.cc
```diff
@@ -2230,16 +2241,73 @@ size_t Tablespace_files::add(space_id_t space_id, const std::string &name) {
...
+  auto begin = std::chrono::high_resolution_clock::now();
+
+  // The parallel for loop. Loads InnoDB tablespaces into cache.
+  par_for(PFS_NOT_INSTRUMENTED, m_ibd_paths, n_threads, open);
+
+  auto end = std::chrono::high_resolution_clock::now();
+  auto elapsed =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin);
+
+  xb::info() << "Time taken to load " << m_ibd_paths.size()
```
altmannmarcelo: Perhaps we could change this to `Completed loading XXX tablespaces into cache in XXX seconds` To indicate the load is done/complete.

storage/innobase/xtrabackup/src/xb_dict.h
```diff
@@ -23,14 +22,232 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
...
+-------------
+1. Load all tables(dict_table_t*) from mysql.ibd using SDI. These tables are not
+closed because we dont want them to be evictable. Dictionary tables reside in
+mysql.ibd
+
+2. After dictionary tables are loaded, we scan mysql.indexes,
+mysql.index_partitions to create a relation "table_id->space_id".
+
+3. By default, we don't load any tables upfront (the old design). A table is
+opened only on on transaction rollback using table_id (dd_table_open_on_id()).
```
altmannmarcelo:
```suggestion
opened only on transaction rollback using table_id (dd_table_open_on_id()).
```

storage/innobase/xtrabackup/src/xb_dict.h
```diff
@@ -23,14 +22,232 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
...
+Not always. We have a special case with partition IBDs. The SDI for all the
+partition tables is not in each partition IBD. Instead, it is "one" of the
+partition IBD So given a partition table_id, we cannot use it corresponding
+"space_id" (This we got from dictionary mysql.index_partitions)
+
+The problem with partitioned tablespaces is that, SDI exists in any ONE of the
+partition IBDs. Not all. Lets say t1#p#p0.ibd, t1#p#p1.ibd, t1#p#p2.ibd.
+All the information about p1 and p2 exists only in p0.ibd
+
+From dictionary (mysql.index_partitions), for a given table_id, we know the
```
altmannmarcelo: @satya-bodapati I am wondering how the server keeps track of which partition the information is stored. Do they have any flag to indicate it or the try the next one is the approach on server too?

storage/innobase/dict/dict0dd.cc
satya-bodapati: The error case is table == nullptr but I see that in the scenario that we load all tables, there could be a problem in one partition table.

Added close_on_error()

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2698,7 +2400,9 @@ static bool innodb_init(bool init_dd, bool for_apply_log) {
     if (dict_persist->table_buffer == nullptr)
       dict_persist->table_buffer =
           ut::new_withkey<DDTableBuffer>(UT_NEW_THIS_FILE_PSI_KEY);
-    srv_dict_recover_on_restart();
+    if (!srv_apply_log_only) {
```
altmannmarcelo: This was been calling unconditionally before. Was this a mistake?
We have `for_apply_log` been passed as parameter here, is srv_apply_log_only == for_apply_log ?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2727,8 +2431,12 @@ static bool innodb_end(void) {
 
   mutex_free(&master_key_id_mutex);
 
+  for (auto table : mysql_ibd_tables) {
+    dict_table_close(table, false, false);
```
altmannmarcelo: This is only mysql.ibd tables correct?

storage/innobase/dict/dict0dd.cc
satya-bodapati: Done

storage/innobase/dict/dict0dd.cc
satya-bodapati: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7330,49 +7036,45 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
...
-
-      mutex_enter(&(dict_sys->mutex));
-      if (!table) {
+      auto result = xb::prepare::dict_load_from_spaces_sdi(space->id);
+      dberr_t err = std::get<0>(result);
+      auto table_vec = std::get<1>(result);
+      if (err != DB_SUCCESS) {
         xb::error() << "cannot find dictionary record of table " << space->name;
-        goto next_node;
+        continue;
```
altmannmarcelo: Any possibility that we have a table_vec and return err != DB_SUCCESS? This would leave one dangling table without closing.

storage/innobase/fil/fil0fil.cc
satya-bodapati: open_ibds is already used :)

storage/innobase/fil/fil0fil.cc
satya-bodapati: Only 4th is unused. Remain are binded as function parameters. See open = std::bind

storage/innobase/dict/dict0dd.cc
satya-bodapati: skip_missing callers are export and  stats.They print a message already

storage/innobase/dict/dict0dd.cc
satya-bodapati: I answered above, when you wrote I changed "ignore_err". Please check

storage/innobase/fil/fil0fil.cc
satya-bodapati: Fixed. I think there was some weird compilation error and the only was to remove const.
But adding back doesn't cause the issues now.

Brought it back

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4265,7 +3973,7 @@ void xtrabackup_backup_func(void) {
 
   recv_is_making_a_backup = true;
   bool data_copying_error = false;
-  std::shared_ptr<xb::dd_tablespaces> xb_dd_space;
+  std::shared_ptr<xb::backup::dd_space_ids> xb_dd_spaces;
```
altmannmarcelo: @satya-bodapati I know this is not new to your patch, but do we need a shared_ptr here? I think an unique_ptr and later move the ownership to datafiles_iter_new. We don't use it on xtrabackup_backup_func anymore.

storage/innobase/fil/fil0fil.cc
satya-bodapati: already addressed. The thing is fil_open_for_xtrabackup() return type is ignored during open.
My new code was not "ignoring". I later fixed it to ignore the return type.

One difference, I print a message now.

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: Ah, I missed it.

I will now use DBUG_LOG("xb_dd"),

It has an added advantage compared to DBUG_PRINT("xb_dd" , message).

We can use c++ style prints in DBUG_LOG

DBUG_LOG("xb_dd",
           "space_id is " << space_id << " for table_id: " << table_id);

no need to use format specifiers %s, %d.. etc

@altmannmarcelo FYI:  ^

storage/innobase/dict/dict0dd.cc
satya-bodapati: Done. rewrote it a bit

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,786 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+This map is used to handle duplicate SDI */
+static std::map<std::pair<int, std::string>, uint64> sdi_id_map;
+
+/* map of schema name and schema id
+This map is used to handle duplicate SDI */
+static std::map<std::string, uint64> dd_schema_map;
+
+using dd_Table_Ptr = std::unique_ptr<dd::Table>;
+
+/** @return true if table_id is found in dd map. This map
```
altmannmarcelo: Really optional, but seems strange to start with return it normally goes in the end of the comment.

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,786 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+  ut_ad(is_space_partitioned(space_id));
+  auto it = space_part_map.find(space_id);
+  return (it != space_part_map.end() ? it->second : 0);
+}
+
+/** Scan the SDI id from DD table "mysql.tables"
+@param[in]  name       tablespace name database/name
+@param[out] sdi_id     id of table
+@param[out] table_name name of the table
+@param[in]  thd        THD */
```
altmannmarcelo: return missing?

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,786 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+
+    if (err != DB_SUCCESS) {
+      goto error;
+    }
+  }
+
+  ut::free(compressed_sdi);
+  ut::free(sdi);
+
+  return (DB_SUCCESS);
```
altmannmarcelo: load_table_cb (both `load_table_func` & `load_table_func_cb` ) will return DB_SUCCESS, we can remove this code and adjust `error:` to something like `cleanup:` and let it return `err`

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,786 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+
+  ulint *offsets = rec_get_offsets(rec, dd_indexes->first_index(), nullptr,
+                                   ULINT_UNDEFINED, UT_LOCATION_HERE, &heap);
+
+  field = (const byte *)rec_get_nth_field(
+      nullptr, rec, offsets, dd_indexes->field_number("index_id"), &len);
+
+  uint64_t part_index_id = mach_read_from_8(field);
+  ut_ad(len == 8);
+  ut_d(xb::info()
```
altmannmarcelo: ut_d ? do we compile it only if debug?

storage/innobase/xtrabackup/test/t/dictionary.sh
```diff
@@ -0,0 +1,291 @@
...
+   INSERT INTO $TABLE (id) SELECT id + (SELECT MAX(id) FROM $TABLE) FROM $TABLE;\
+   SELECT SLEEP (10000);
+ " test &
+ uncommitted_trx=$!
+
+ wait_for_bg_trx
+
+ vlog "Backup"
+ rm -rf $topdir/backup
+ xtrabackup --backup --target-dir=$topdir/backup 2>&1 | tee $topdir/pxb.log
```
altmannmarcelo: Can we add a trap to execute `kill_bg_trx $uncommitted_trx` in case xtrabackup fails, we will never reach `kill_bg_trx $uncommitted_trx`

storage/innobase/xtrabackup/test/t/dictionary.sh
```diff
@@ -0,0 +1,291 @@
...
+   DELETE FROM t1 WHERE a>$START AND a < $END; \
+   SELECT SLEEP (10000);
+ " test &
+ uncommitted_trx=$!
+
+ wait_for_bg_trx
+
+ vlog "Backup"
+ rm -rf $topdir/backup
+ xtrabackup --backup --target-dir=$topdir/backup 2>&1 | tee $topdir/pxb.log
```
altmannmarcelo: same trap comment here.

storage/innobase/xtrabackup/test/t/dictionary.sh
```diff
@@ -0,0 +1,291 @@
...
+   DELETE FROM pt1 WHERE a < 2500;
+   SELECT SLEEP (10000);
+ " test &
+ uncommitted_trx=$!
+
+ wait_for_bg_trx
+
+ vlog "Backup"
+ rm -rf $topdir/backup
+ xtrabackup --backup --target-dir=$topdir/backup 2>&1 | tee $topdir/pxb.log
```
altmannmarcelo: same trap comment here

altmannmarcelo (changes requested):
@satya-bodapati first of all, thanks a lot for all the effort on this PR. This is a huge improvement and a long term pain we had in PXB.

I have a few minor comments apart from the std::tuple return code. We might discuss offline how big of a deal is to adjust it to return db_err and get the vector/dd_table as an in,out parameter.

storage/innobase/dict/dict0dd.cc
satya-bodapati: I've fixed this in the commit "Fix dict_sys mutex acquisitions"

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Fixed

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Fixed in commit "Fix dict_sys mutex acquistions".

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: You mean, add a new option to PXB? and read from it?

storage/innobase/dict/dict0dd.cc
satya-bodapati: Done

satya-bodapati:
https://github.com/percona/percona-xtrabackup/pull/1435#discussion_r1149603277

Added 

```
  // In PXB, we always load tables as evictable. We do not have create table
  // and upgrade from 5.7 scenarios. These paths can load table as non-evictable
  IF_XB(ut_ad(can_be_evicted);)
```

storage/innobase/dict/dict0dd.cc
satya-bodapati: Partition_leaf_vector
typedef std::vector<Partition *> Partition_leaf_vector;

storage/innobase/dict/dict0dd.cc
satya-bodapati: Sure, I wrote it as function comment of  dd_table_load_on_dd_obj in storage/innoabse/include/dict0dd.h

@param[in]  table_id  Table_id of the InnoDB table to be loaded. If this is 0,
                      all partitions will be loaded

storage/innobase/dict/dict0dd.cc
satya-bodapati: marce, Rollback path asks for single table_id in space_id.
For export and stats, for one space_id, we load all tables in space_id. Then for each table object, we do export or stats.

Since we return multiple tables, we need a vector.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Fixed

storage/innobase/fil/fil0fil.cc
satya-bodapati: Fixed

storage/innobase/fil/fil0fil.cc
satya-bodapati: This just me being nice :) Previously we completely ignored the return type from fil_open_for_xtrabackup() during prepare. 
I just filtered common error codes (that are acceptable) and anything else I print message.

I can either make it debug only or remove it if you wish

storage/innobase/fil/fil0fil.cc
satya-bodapati: Done

storage/innobase/xtrabackup/src/xb_dict.h
satya-bodapati: Done

storage/innobase/xtrabackup/src/xb_dict.h
satya-bodapati: Server doesn't use SDI at all to load tables. It deserialized dd::Table from mysql.tables/indexes/partitions/etc
On the partition to write SDI, the server always uses the "first' available partition.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Yes, it is a mistake before to do rollback of dictionary trxs in apply_log only mode.

Regarding the for_apply_log, it seems the callers are

```
Functions calling this function: innodb_init

  File          Function                Line
0 xtrabackup.cc xtrabackup_stats_func   4735 if (innodb_init(true, false)) exit(EXIT_FAILURE);
1 xtrabackup.cc xtrabackup_prepare_func 6991 if (innodb_init(true, true)) {
```
so they are not same. May we didn't want to initiate recovery when operating on server log data directory with redo files..
Name could have been better

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Right, only tables in mysql.ibd

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: if we cannot open table, it should be empty vector. Added debug assert

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Right, I had the same question to @rahulmalik87
I think he told me that it is required because the structure is passed to the parallel copy threads.

But after analyzing it deep, it seems that we use it at the time of populating the iterator and not after that.
So it is not used by copy threads.

@rahulmalik87 can you please check and confirm. We can make this unique_ptr or not?

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: it seems it is a valid doxygen syntax :)

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: Fixed. Looks like old code missed it too

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: there is a possibility of an error from load_table_func() in dict_load_tables_from_space_id.

```
std::tie(ret, tables) = dd_table_load_on_dd_obj(
        dc, space_id, *dd_table.get(), table_id, thd, &schema_name, implicit);
    if (ret != 0) {
      return (DB_ERROR);
```

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: yes, we do have ut_d(); simply means compile only on debug. It looks like an assertion, but it is not :) 

btw, I changed this to DBUG_LOG("xb_dd",..)

storage/innobase/xtrabackup/test/t/dictionary.sh
satya-bodapati: Done

storage/innobase/xtrabackup/test/t/dictionary.sh
satya-bodapati: Done

storage/innobase/xtrabackup/test/t/dictionary.sh
satya-bodapati: Done

storage/innobase/xtrabackup/src/xb_dict.cc
altmannmarcelo: returning `err` in this case will be sufficient right?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,38 +1312,109 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
+
+    if (table_id != 0) {
+      // Load specific partition based on table_id
+      auto i =
+          std::search_n(dd_table.leaf_partitions().begin(), end, 1, table_id,
+                        [](const dd::Partition *p, table_id_t id) {
+                          return (p->se_private_id() == id);
+                        });
+      if (i == end) {
+        // The partition table_id doesn't exist in this Table partitions.
```
rahulmalik87: If it is unexpected can we add some warning/error?

storage/innobase/fil/fil0fil.cc
rahulmalik87: open_ibds_thread or some other name ?

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,789 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+/** Load a specific table from space_id
+@param[in] space_id InnoDB tablespace_id
+@param[in] table_id InnoDB table id
+@return tuple <a,b>
+a - DB_SUCCESS on success, other DB_ on errors
+b - std::vector<dict_table_t*>, empty on errors */
+static xb_dict_tuple dict_load_tables_from_space_id_wrapper(
+    space_id_t space_id, table_id_t table_id) {
+  fil_space_t *space = fil_space_get(space_id);
+  if (space == nullptr) {
```
rahulmalik87: some error or warning if we space== nullptr?

storage/innobase/xtrabackup/src/xb_dict.cc
```diff
@@ -48,4 +70,789 @@ std::shared_ptr<dd_tablespaces> build_space_id_set(MYSQL *connection) {
...
+  }
+
+  // For partition tables, SDI exists in only one partition IBD, loop
+  // through such IBDs to find the required table_id
+  uint64_t part_id = get_part_id_for_space(space_id);
+  if (part_id == 0) return {DB_TABLESPACE_NOT_FOUND, {}};
+
+  auto low = part_id_spaces_map.lower_bound(part_id);
+  auto high = part_id_spaces_map.upper_bound(part_id);
+  while (low != high) {
```
rahulmalik87: What would happen in case of a single partition?

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: what was the consequence of calling rollbackup of dictionary trx?  Do we have a test case to show problem?

storage/innobase/xtrabackup/src/xtrabackup.cc
rahulmalik87: yeah @satya-bodapati  right, we used that for populating the iterator and not after that so no parallel use.  
May be I changed some thing after initial patch. 
 
DO you want we can create a separate Jira ticket to fix that ?

storage/innobase/dict/dict0dd.cc
```diff
@@ -1299,38 +1312,109 @@ int dd_table_open_on_dd_obj(dd::cache::Dictionary_client *client,
...
+      ut_ad(0);
+      xb::warn() << "Unable to convert dd::Table to dict_table for table: ["
+                 << schema_name << "/" << dd_table.name() << "]";
+      return {1, {}};
     }
-    dd_table_close(table, thd, nullptr, false);
-    return ret;
   }
+
+  // Shouldn't reach here because on success or error, we should have returned
```
rahulmalik87: Given we are returning from both if and else block do we really need this code? Will compiler complain?

rahulmalik87 (changes requested):
Some minors comments. Almost LGTM . Please submit to jenkins.

storage/innobase/dict/dict0dd.cc
satya-bodapati: I think this can happen with lock-ddl=false. Adding a debug assert and warning for release build

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: rollback can be requested on many dropped tables and I think error or warning causes more confusion.

A tablespace not found is expected and no action to be done

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: high to point to upper bound, we will still go in the while loop

storage/innobase/fil/fil0fil.cc
satya-bodapati: Used open_ibds_slice

storage/innobase/fil/fil0fil.cc
satya-bodapati: The caller prints a final message after loading all. This is just the thread progress

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: we will need an extra patch to catch the redo log generated during apply log phase.  Since it is touching the dictionary-related area, I fixed this too.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Yes,  all paths should have a return type.
Also, its a good practice to crash on the un-expected path (debug builds)

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: Rewrote this code using smart pointers

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: I mean make it read from https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_open_files . If I am not mistaken, we already parse and expose all innodb_* variables.

storage/innobase/dict/dict0dd.cc
altmannmarcelo: @satya-bodapati got that we need a vector of dict_table_t, but what I mean is the return type bringing two information. 
I would rather have the function to return db_err and the vector be a in/out parameter

altmannmarcelo: Something like:

```
dberr_t dd_table_load_on_dd_obj(dd::cache::Dictionary_client *client,
                                      space_id_t space_id,
                                      const dd::Table &dd_table,
                                      table_id_t table_id, THD *thd,
                                      const dd::String_type *schema_name,
                                      bool implicit,
                                      std::vector<dict_table_t *> *table_vec);
```

storage/innobase/fil/fil0fil.cc
altmannmarcelo: I would play safe here and make it debug_only .What you think?

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: I am ok with a JIRA

storage/innobase/xtrabackup/src/xb_dict.cc
altmannmarcelo: FYI: we ported the ut_make_unique_ptr_zalloc_nokey from PS. But they have got rid of it because it was not used in 8.0.26. Any concern on using it ?

storage/innobase/xtrabackup/test/t/dictionary.sh
```diff
@@ -0,0 +1,306 @@
...
+ vlog "Restore"
+ stop_server
+ rm -rf $MYSQLD_DATADIR/*
+ xtrabackup --move-back --target-dir=${topdir}/backup
+ start_server
+ verify_db_state test
+}
+
+function create_part_table() {
+  vlog "Create partition table"
```
rahulmalik87: Can you please create table with subpartition and do some uncommitted transaction. so we have rollbackup on partition where SDI is not stored

rahulmalik87 (changes requested):

storage/innobase/xtrabackup/src/xb_dict.cc
rahulmalik87: Does we need multiple partitions to table_id mapping because of sub partition?

altmannmarcelo:
@satya-bodapati another sugestion:

```
2023-04-27T17:37:17.569845-00:00 0 [Note] [MY-012203] [InnoDB] Directories to scan './'
2023-04-27T17:37:17.569871-00:00 0 [Note] [MY-012204] [InnoDB] Scanning './'
2023-04-27T17:37:20.375644-00:00 0 [Note] [MY-012207] [InnoDB] Using 17 threads to scan 1000003 tablespace files
2023-04-27T17:37:30.376407-00:00 0 [Note] [MY-012200] [InnoDB] Thread# 4 - Checked 1237/62500 files
. . . 
2023-04-27T17:48:11.942890-00:00 0 [Note] [MY-012200] [InnoDB] Thread# 14 - Checked 61695/62500 files
2023-04-27T17:48:13.455207-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:14.975996-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:16.102324-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:16.245244-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:16.643509-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:17.023574-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:17.356843-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:17.674888-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:17.744431-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:17.944624-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.221327-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.236320-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.292918-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.327884-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.375826-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.472105-00:00 0 [Note] [MY-012201] [InnoDB] Checked 62500 files
2023-04-27T17:48:18.473312-00:00 0 [Note] [MY-012208] [InnoDB] Completed space ID check of 1000005 files.
2023-04-27T17:48:18.642909-00:00 0 [Note] [MY-012955] [InnoDB] Initializing buffer pool, total size = 128.000000M, instances = 1, chunk size =128.000000M 
2023-04-27T17:48:18.668298-00:00 0 [Note] [MY-012957] [InnoDB] Completed initialization of buffer pool
2023-04-27T17:48:18.696671-00:00 0 [Note] [MY-011951] [InnoDB] page_cleaner coordinator priority: -20
2023-04-27T17:48:18.705595-00:00 0 [Note] [MY-011954] [InnoDB] page_cleaner worker priority: -20
2023-04-27T17:48:18.705658-00:00 0 [Note] [MY-011954] [InnoDB] page_cleaner worker priority: -20
2023-04-27T17:48:18.705768-00:00 0 [Note] [MY-011954] [InnoDB] page_cleaner worker priority: -20
2023-04-27T17:48:18.718376-00:00 0 [Note] [MY-013883] [InnoDB] The latest found checkpoint is at lsn = 775179431 in redo log file ./#innodb_redo/#ib_redo0.
2023-04-27T17:48:18.718458-00:00 0 [Note] [MY-012560] [InnoDB] The log sequence number 775118558 in the system tablespace does not match the log sequence number 775179431 in the redo log files!
2023-04-27T17:48:18.718479-00:00 0 [Note] [MY-012551] [InnoDB] Database was not shutdown normally!
2023-04-27T17:48:18.718496-00:00 0 [Note] [MY-012552] [InnoDB] Starting crash recovery.
2023-04-27T17:48:18.720091-00:00 0 [Note] [MY-013086] [InnoDB] Starting to parse redo log at lsn = 775179310, whereas checkpoint_lsn = 775179431 and start_lsn = 775179264
2023-04-27T17:48:18.720122-00:00 0 [Note] [MY-012550] [InnoDB] Doing recovery: scanned up to log sequence number 775188626
2023-04-27T17:48:18.736647-00:00 0 [Note] [MY-013083] [InnoDB] Log background threads are being started...
2023-04-27T17:48:18.737040-00:00 0 [Note] [MY-012532] [InnoDB] Applying a batch of 24 redo log records ...
2023-04-27T17:48:18.748860-00:00 0 [Note] [MY-012533] [InnoDB] 100%
2023-04-27T17:48:19.249089-00:00 0 [Note] [MY-012535] [InnoDB] Apply batch completed!
2023-04-27T17:48:19.356508-00:00 0 [Note] [MY-013084] [InnoDB] Log background threads are being closed...
2023-04-27T17:48:19.360127-00:00 0 [Note] [MY-013041] [InnoDB] Resizing redo log from 8M to 1024M (LSN=775188626) synchronously. If this takes too long, consider starting the server with large --innodb_redo_log_capacity, and resizing the redo log online using SET.
2023-04-27T17:48:19.363494-00:00 0 [Note] [MY-012968] [InnoDB] Starting to delete and rewrite redo log files.
2023-04-27T17:48:19.363591-00:00 0 [Note] [MY-011825] [InnoDB] Removing redo log file: ./#innodb_redo/#ib_redo0
2023-04-27T17:48:19.387507-00:00 0 [Note] [MY-011825] [InnoDB] Creating redo log file at ./#innodb_redo/#ib_redo0_tmp with file_id 0 with size 33554432 bytes
2023-04-27T17:48:19.393282-00:00 0 [Note] [MY-011825] [InnoDB] Renaming redo log file from ./#innodb_redo/#ib_redo0_tmp to ./#innodb_redo/#ib_redo0
2023-04-27T17:48:19.398739-00:00 0 [Note] [MY-012893] [InnoDB] New redo log files created, LSN=775189004
2023-04-27T17:48:19.398814-00:00 0 [Note] [MY-013083] [InnoDB] Log background threads are being started...
2023-04-27T17:48:19.399238-00:00 0 [Note] [MY-013252] [InnoDB] Using undo tablespace './undo_001'.
2023-04-27T17:48:19.399302-00:00 0 [Note] [MY-013252] [InnoDB] Using undo tablespace './undo_002'.
2023-04-27T17:48:19.399659-00:00 0 [Note] [MY-012910] [InnoDB] Opened 2 existing undo tablespaces.
2023-04-27T17:48:19.402134-00:00 0 [Note] [MY-011980] [InnoDB] GTID recovery trx_no: 26641536
2023-04-27T17:48:19.403994-00:00 0 [Note] [MY-013776] [InnoDB] Parallel initialization of rseg complete
2023-04-27T17:48:19.404010-00:00 0 [Note] [MY-013777] [InnoDB] Time taken to initialize rseg using 4 thread: 1887 ms.
2023-04-27T17:48:19.404049-00:00 0 [Note] [MY-013023] [InnoDB] 1 transaction(s) which must be rolled back or cleaned up in total 1 row operations to undo
2023-04-27T17:48:19.404057-00:00 0 [Note] [MY-013024] [InnoDB] Trx id counter is 26642014
2023-04-27T17:48:19.405286-00:00 0 [Note] [MY-012923] [InnoDB] Creating shared tablespace for temporary tables
2023-04-27T17:48:19.405363-00:00 0 [Note] [MY-012265] [InnoDB] Setting file './ibtmp1' size to 12 MB. Physically writing the file full; Please wait ...
2023-04-27T17:48:19.453727-00:00 0 [Note] [MY-012266] [InnoDB] File './ibtmp1' size is now 12 MB.
2023-04-27T17:48:19.454918-00:00 0 [Note] [MY-013627] [InnoDB] Scanning temp tablespace dir:'./#innodb_temp/'
2023-04-27T17:48:19.492922-00:00 0 [Note] [MY-013018] [InnoDB] Created 128 and tracked 128 new rollback segment(s) in the temporary tablespace. 128 are now active.
2023-04-27T17:48:19.493281-00:00 0 [Note] [MY-012976] [InnoDB] 8.0.32 started; log sequence number 775189014
2023-04-27T17:48:19.890094-00:00 0 [Warning] [MY-012091] [InnoDB] Allocated tablespace ID 1 for sys/sys_config, old maximum was 0
2023-04-27T17:48:29.897269-00:00 0 [Note] [MY-011825] [Xtrabackup] Thread# 0 - Opened 2138/62500 tablespaces
2023-04-27T17:48:29.899308-00:00 0 [Note] [MY-011825] [Xtrabackup] Thread# 1 - Opened 2133/62500 tablespaces
. . .
2023-04-27T17:55:28.167103-00:00 0 [Note] [MY-011825] [Xtrabackup] Completed loading of 1000003 tablespaces into cache in 428.258 seconds
2023-04-27T17:55:38.790838-00:00 0 [Note] [MY-011825] [Xtrabackup] Time taken to build dictionary: 10.5653 seconds
2023-04-27T17:55:38.837142-00:00 0 [Note] [MY-013014] [InnoDB] Starting in background the rollback of uncommitted transactions
2023-04-27T17:55:38.837181-00:00 0 [Note] [MY-013011] [InnoDB] Rolling back trx with id 26641530, 1 rows to undo
2023-04-27T17:55:38.844881-00:00 0 [Note] [MY-013012] [InnoDB] Rollback of trx with id 26641530 completed
2023-04-27T17:55:38.844927-00:00 0 [Note] [MY-013015] [InnoDB] Rollback of non-prepared transactions completed
``` 

We are iterating over all the tablespaces twice, can we combine both?

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: This cache limit doesn't translate to open file descriptors. See more information here: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html

There is a max cap of 2000 files based on the auto-sizing setting. We can leave it hardcoded to 2000 for most use cases. Currently, I am using 4000 as a cache limit.

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: imho, we should have kept it in PS too. No issues in PXB, because we retained that code.

storage/innobase/xtrabackup/src/xb_dict.cc
satya-bodapati: Not required, every partition (sub or not) has unique table_id assigned.

satya-bodapati:
> 
@altmannmarcelo 
The first scan creates a vector of space_id -> file_name relation. At this point,  tablespace is NOT opened. In-memory object fil_space_t is also not created.

Second scan does the page 0 read and creates the fil_space_t object. We cannot combine these two because, we need redo to be applied to page 0.

Steps are:
1. files scan (create space_id->file_name mapping)
2. apply redo (page 0 is consistent now)
3. files scan again( this time create fil_space_t)

Another idea is to avoid this second scan and load only the fil_space_ts required for rollback. Its not straightforward, though. 
See https://jira.percona.com/browse/PXB-3027

It can be handled separately.

storage/innobase/xtrabackup/test/t/dictionary.sh
satya-bodapati: subpartitions are not any different from the rollback perspective.

PR #1437, opened 2023-03-23
PXB-3022 : From PXB 8.0.30 version, prepare requires more memory than previous versions

satya-bodapati:
Problem:
--------
This is because we now store "dd::Tables" in the global map for the purpose of extracting the "instant column" info required for creating a ".cfg" file for --export. We store these dd::Tables in the global map by default, even if --export is not used. The memory occupied by dd::Table is not freed until end of prepare

Fix:
----
Store these dd::Tables in global map only if --export is specified during prepare


altmannmarcelo (approved):
LGTM. Please adjust the merge branch to Release branch

PR #1443, opened 2023-03-31
Add RISC-V support

hack3ric:
Adds various memory barrier methods:

- mb(), rmb(), wmb() according to [RISC-V unpriviledged specs](https://riscv.org/wp-content/uploads/2019/12/riscv-spec-20191213.pdf) section A.5
- xcng() from [Linux kernel source code](https://github.com/torvalds/linux/blob/master/arch/riscv/include/asm/cmpxchg.h#L124)
- cpu_pause() using Zihintpause's pause, with checks included.

See also https://github.com/percona/percona-server/pull/5030.

### DIFF
--- a/configure.cmake
+++ b/configure.cmake
@@ -496,6 +496,17 @@ IF(NOT CMAKE_CROSSCOMPILING AND NOT MSVC)
     }
     " HAVE_HMT_PRIORITY_INSTRUCTION)
   ENDIF()
+
+  # Check for RISC-V Zihintpause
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
Hi @hack3ric .

Thanks for your contribution. However, we inherit those files from upstream [mysql/mysql-server ](https://github.com/mysql/mysql-server). Once upstream incorporates and supports those changes, we will seamlessly do the same. 

Thanks once again for your PR.

PR #1447, opened 2023-04-05
Implemented PXB-2514 - multiple FIFO streams

altmannmarcelo:
https://jira.percona.com/browse/PXB-2514

New FIFO datasink for xtrabackup / xbcloud / xbstream that uses true
multi-thread.

-- FIFO datasink --
FIFO/named pipes is a new data sink that transfers data
between xtrabackup and xbcloud. The main advantage of using FIFO is
that it allows multiple threads to truly write in parallel, which is
not possible with STDIN/STDOUT. FIFO works by creating a FIFO file and
opening it with read and write mode. By default, FIFO is opened in
blocking mode. This is what we want for when both sides have been open
because it will ensure that:
- Read end will block while there is no data to read.
- Write end will block if the reader is not reading fast enough.
- This avoids the threads having to use epool/select to validate if the
other end is ready.

However, opening a FIFO in blocking mode can make the process hang
forever if the other end is not opened. This is the case when xtrabackup
is writing to FIFO and xbcloud is not reading from it. To avoid this, we
open the FIFO in non-blocking mode. This way, we can set a timeout and
keep validating if the other side has been opened. If the timeout is
reached, we close the FIFO and exit. If both sides have been opened, we
set the FIFO back to blocking mode.

To support multiple parallel streams, a new struct called
ds_parallel_stream_ctxt_t has been created. This struct is used to
maintain a list of ds_stream_ctxt_t (either a single-entry list
pointing to STDOUT or multiple entries pointing to FIFO files).

When a thread calls xbstream_open we get the first entry of the
ds_parallel_stream_ctxt_t list and assign it to ds_stream_ctxt_t object
that will be used to form the return file. Once we pick the first entry
from the parallel stream list, we reenter it at the end of the list.
This way we maintain a LRU list and multiple copy threads will evenly
utilize the available FIFO ds_stream_ctxt_t.

Once we have our ds_stream_ctxt_t assigned, we will check if it has a
destination file assigned to it, if not we will do it via ds_open this
time passing out destination datasink (either FIFO or STDOUT).
FIFO datasink has a list of FIFO files it can assign on an open operation.
When xbstream::ds_open calls FIFO::fifo_open we assign the first FIFO file
in the list and from this point forward xbstream ds_stream_ctxt_t will be
bound to this particular FIFO file. We do not assign the fifo file back to
the list of fifo files as xbstream ds_stream_ctxt_t->fifo is a 1:1 mapping.

When using FIFO, ddded logic to ensure --parallel is at least equal to the
number FIFO files. Otherwise, there will be FIFO files without data been
written to it.

-- xbcloud Download --
Xbstream format has a limitation in that it requires that each chunk of
a file is downloaded/extracted in order. For example, chunk 3 of the
file X cannot arrive before chunks 0, 1 and 2.
More details at https://jira.percona.com/browse/PXB-2950

Each thread has its own download state instance. With that in mind, to
download from xbcloud we keep 3 lists of files:

* cntx->global_list->files - Global list of files. Shared among
all threads. This list is populated by the main thread by querying the
object storage bucket. Each file has a next_chunk field that indicates
the next chunk and last_chunk field that indicates the last chunk of
the file.

* thread_state->in_progress_files - List of files that are currently
being downloaded by this thread. This list is intended to synchronize the
chunks of a file. If a file has in-progress chunks, we do not start the
next chunk of this file.

* thread_state->file_list - List of files that are assigned to
current thread.

Example:

cntx->global_list->files {
["bkp/performance_schema/events_stages_su_125.sdi"] = {
last_chunk = 1,
next_chunk = 0,
name = "bkp/performance_schema/events_stages_su_125.sdi"
},
["bkp/performance_schema/events_statement_135.sdi"] = {
last_chunk = 1,
next_chunk = 0,
name = "bkp/performance_schema/events_statement_135.sdi"
}
}

thread_state->in_progress_files {
[0] = "bkp/performance_schema/accounts_154.sdi",
[1] = "bkp/performance_schema/events_stages_su_126.sdi",
[2] = "bkp/performance_schema/events_statement_136.sdi",
[3] = "bkp/sbtest/sbtest5.ibd",
[4] = "bkp/sbtest/sbtest7.ibd"
}

thread_state->file_list {
["bkp/performance_schema/events_stages_su_126.sdi"] = {
last_chunk = 1,
next_chunk = 0,
name = "bkp/performance_schema/events_stages_su_126.sdi"
},
["bkp/performance_schema/events_statement_136.sdi"] = {
last_chunk = 1,
next_chunk = 1,
name = "bkp/performance_schema/events_statement_136.sdi"
},
["bkp/performance_schema/accounts_154.sdi"] = {
last_chunk = 1,
next_chunk = 0,
name = "bkp/performance_schema/accounts_154.sdi"
},
["bkp/sbtest/sbtest7.ibd"] = {
last_chunk = 24,
next_chunk = 4,
name = "bkp/sbtest/sbtest7.ibd"
},
["bkp/sbtest/sbtest5.ibd"] = {
last_chunk = 24,
next_chunk = 8,
name = "bkp/sbtest/sbtest5.ibd"
}
}

Download flow is as follow:
* We check that we don't have more in progress file than --parallel
option.
* if the global list of files and the thread list of files we completed
the flow
* We attempt to get the next file from per thread list
(thread_state->file_list). Where we check if the file is not part
of the in progress list. If it is, we skip it.
* If there is no available files from per thread list, we pick a new
one from global list. If we get a file from this list, the file is
removed from global list.
* Once we have a file, we indicate that we are downloading it by adding
it to the in progress list and per thread list.
* We make an async request to download the current chunk of the file.
The event handler thread will handle this via a callback.
* Event handler callback, will complete the download of the chunk.
It will remove the file from in progress list and advance the next
chunk of the file. In case we just completed the last chunk of the file,
we remove the file from the per-thread list.

-- xbcloud upload --

Upload is a bit more simple. We keep reading more data on STDIN or
FIFO file. Until the other side (xtrabackup) closes the pipe.
For each chunk of data, we create one async request that will be handled
by the event handler.

-- xbstream --
xbstream changes are related to where to read the data from. Similar to
other tools, we read from STDIN in case of a single thread and from the
FIFO files in case of multiple threads.

Introduced 3 new parameters:
--fifo-streams=# Number of FIFO files to use for parallel datafiles
                stream. Setting this parameter to 1 disables FIFO and
                stream is sent to STDOUT.
--fifo-dir=name   Directory to write Named Pipe.
--fifo-timeout=#  How many seconds to wait for other end to open the
                stream for reading. Default 60 seconds

-- pthread refactor --
Moved from pthread to std::thread and from  ptread_mutex to std::mutex

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_fifo.cc
+++ b/storage/innobase/xtrabackup/src/ds_fifo.cc
@@ -1,0 +1,199 @@
+/******************************************************
+Copyright (c) 2023 Percona LLC and/or its affiliates.
+
+FIFO datasink implementation for XtraBackup.
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; version 2 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+*******************************************************/
+
+#include <my_base.h>
+#include <my_io.h>
+#include <my_thread_local.h>
+#include <mysql/service_mysql_alloc.h>
+#include <mysys_err.h>
+#include <mutex>
+#include <unordered_map>
+#include "common.h"
+#include "datasink.h"
+#include "file_utils.h"
+#include "msg.h"
+
+typedef struct {
+  /* List of FIFO files to be used for stream */
+  std::unordered_map<std::string, File> FIFO_list;
+  /* Mutex protecting FIFO list */
+  std::mutex fifo_mutex;
+
+  /* Add a new pair of fullpath and fd to FIFO_list */
+  void populate_list(std::string fullpath, File fd) {
+    std::lock_guard<std::mutex> g(fifo_mutex);
+    FIFO_list.insert({fullpath, fd});
+  }
+
+  /**
+  Get the first file from FIFO_list. This call will remove the
+  file from the list.
+  @return false in case of file found from list, true in case of error */
+  bool allocate_from_list(std::string &fullpath, File &fd) {
+    std::lock_guard<std::mutex> g(fifo_mutex);
+    auto fifo_it = FIFO_list.begin();
+    if (fifo_it == FIFO_list.end()) {
+      return true;
+    }
+    fullpath = fifo_it->first;
+    fd = fifo_it->second;
+    FIFO_list.erase(fifo_it);
+    return false;
+  }
+} ds_fifo_ctxt_t;
+
+typedef struct {
+  File fd;
+  char *path;
+  ds_fifo_ctxt_t *fifo_context;
+} ds_fifo_file_t;
+
+extern uint xtrabackup_fifo_streams;
+extern uint xtrabackup_fifo_timeout;
+static ds_ctxt_t *fifo_init(const char *root);
+static ds_file_t *fifo_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *mystat);
+static int fifo_write(ds_file_t *file, const void *buf, size_t len);
+static int fifo_close(ds_file_t *file);
+static void fifo_deinit(ds_ctxt_t *ctxt);
+
+datasink_t datasink_fifo = {&fifo_init, &fifo_open,  &fifo_write,
+                            nullptr,    &fifo_close, &fifo_deinit};
+
+static void cleanup_on_error(int idx, const char *root) {
+  std::string path;
+  char fullpath[FN_REFLEN];
+  for (int i = 0; i <= idx; i++) {
+    path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    unlink(fullpath);
+  }
+}
+
+/**
+Initialize FIFO datasink. This function is responsible for creating stream dir,
+and the fifo files
+
+@param [in]  root  path to create FIFO files.
+
+@return ds_ctxt_t object with a ptr to ds_fifo_ctxt_t object.
+NULL in case of error. */
+static ds_ctxt_t *fifo_init(const char *root) {
+  ds_fifo_ctxt_t *fifo_context = new ds_fifo_ctxt_t;
+  File fd;
+  ds_ctxt_t *ctxt = new ds_ctxt_t;
+  char fullpath[FN_REFLEN];
+
+  if (my_mkdir(root, 0600, MYF(0)) < 0 && my_errno() != EEXIST &&
+      my_errno() != EISDIR) {
+    char errbuf[MYSYS_STRERROR_SIZE];
+    my_error(EE_CANT_MKDIR, MYF(0), root, my_errno(),
+             my_strerror(errbuf, sizeof(errbuf), my_errno()));
+    return NULL;
+  }
+
+  for (uint i = 0; i < xtrabackup_fifo_streams; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    if (mkfifo(fullpath, 0600) < 0) {
+      msg_ts("mkfifo(%s) failed with error %d\n", fullpath, errno);
+      if (errno == EEXIST) {
+        msg_ts(
+            "FIFO file %s already exists. Please ensure you don't have other "
+            "xtrabackup instance running, remove the file(s) and try again.\n",
+            fullpath);
+      } else {
+        cleanup_on_error(i, root);
+      }
+      return NULL;
+    }
+  }
+
+  for (uint i = 0; i < xtrabackup_fifo_streams; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    fd = open_fifo_for_write_with_timeout(fullpath, xtrabackup_fifo_timeout);
+    if (fd < 0) {
+      cleanup_on_error(i, root);
+      return NULL;
+    }
+    fifo_context->populate_list(fullpath, fd);
+  }
+
+  ctxt->ptr = fifo_context;
+  ctxt->root = my_strdup(PSI_NOT_INSTRUMENTED, root, MYF(MY_FAE));
+
+  return ctxt;
+}
+
+static ds_file_t *fifo_open(ds_ctxt_t *ctxt,
+                            const char *path __attribute__((unused)),
+                            MY_STAT *mystat __attribute__((unused))) {
+  ds_fifo_ctxt_t *fifo_context = (ds_fifo_ctxt_t *)ctxt->ptr;
+  std::string fifo_path;
+  File fd;
+  if (fifo_context->allocate_from_list(fifo_path, fd)) return NULL;
+
+  size_t path_len = fifo_path.length() + 1; /* terminating '\0' */
+
+  ds_file_t *file = (ds_file_t *)my_malloc(
+      PSI_NOT_INSTRUMENTED,
+      sizeof(ds_file_t) + sizeof(ds_fifo_file_t) + path_len, MYF(MY_FAE));
+  ds_fifo_file_t *fifo_file = (ds_fifo_file_t *)(file + 1);
+
+  fifo_file->fd = fd;
+  fifo_file->fifo_context = fifo_context;
+  fifo_file->path = (char *)fifo_path.c_str();
+  file->path = (char *)fifo_file + sizeof(ds_fifo_file_t);
+  memcpy(file->path, fifo_path.c_str(), path_len);
+
+  file->ptr = fifo_file;
+
+  return file;
+}
+
+static int fifo_write(ds_file_t *file, const void *buf, size_t len) {
+  File fd = ((ds_fifo_file_t *)file->ptr)->fd;
+
+  if (!my_write(fd, static_cast<const uchar *>(buf), len,
+                MYF(MY_WME | MY_NABP))) {
+    posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED);
+    return 0;
+  }
+
+  return 1;
+}
+
+static int fifo_close(ds_file_t *file) {
+  ds_fifo_ctxt_t *fifo_context = ((ds_fifo_file_t *)file->ptr)->fifo_context;
+  File fd = ((ds_fifo_file_t *)file->ptr)->fd;
+  my_close(fd, MYF(MY_WME));
+  fifo_context->populate_list(((ds_fifo_file_t *)file)->path, fd);
+  my_free(file);
+
+  return 0;
+}
+
+static void fifo_deinit(ds_ctxt_t *ctxt) {
+  ds_fifo_ctxt_t *fifo_context = (ds_fifo_ctxt_t *)ctxt->ptr;
+  assert(fifo_context->FIFO_list.size() == xtrabackup_fifo_streams);
+  delete fifo_context;
+  my_free(ctxt->root);
+  delete ctxt;
+}
--- a/storage/innobase/xtrabackup/src/ds_fifo.h
+++ b/storage/innobase/xtrabackup/src/ds_fifo.h
@@ -1,0 +1,70 @@
+/******************************************************
+Copyright (c) 2023 Percona LLC and/or its affiliates.
+
+FIFO datasink interface for XtraBackup.
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; version 2 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+*******************************************************/
+
+/* FIFO Datasink implementation
+
+at xtrabackup_init_datasinks we initialize ds_data as DS_TYPE_FIFO, and create a ds object as XB_STREAM_FMT_XBSTREAM (xbstream).
+We adjust ds saying it has to pipe its data to DS_TYPE_FIFO and we adjust ds_data to xbstream.
+From this point forward we know we need to first convert the raw data we received from copy threads to xbstream and then later pipe the transformed data (xbstream format) to FIFO (same was there before but fixed to STDOUT).
+
+(gdb) p *ds_data
+$2 = {
+ datasink = 0x55555d670dc0 <datasink_xbstream>,
+ root = 0x0,
+ ptr = 0x55555e5af190,
+ pipe_ctxt = 0x55555e1e43f0,
+ fs_support_punch_hole = 105
+}
+(gdb) p *ds_data->pipe_ctxt
+$3 = {
+ datasink = 0x55555d670cc0 <datasink_fifo>,
+ root = 0x55555e15ae10 "/tmp/stream",
+ ptr = 0x55555e15ae50,
+ pipe_ctxt = 0x666e632e706d75,
+ fs_support_punch_hole = false
+}
+
+fifo_init - responsible for creating the stream dir and all the fifo files (thread_XX).
+Once created, we open them for writing at the init function despite the fact that we have an open function, this is because with FIFO / STDOUT we do not actually close the FD when a copy thread finishes its works on a specific data file, so ds_open / ds_close operate differently, ds_open just returns a ds_fifo_file_t as file(ds_file_t)->ptr back to xbstream datasink
+xbstream_init will create a list of ds_stream_ctxt_t, those stream contexts will later receive a fifo FD assigned to it. this list is called ds_parallel_stream_ctxt_t.
+at xtrabackup_copy_datafile (aka copy threads) we call ds_open passing ds_data (xbstream) as a parameter - xbstream_open will pick the first ds_stream_ctxt_t from the ds_parallel_stream_ctxt_t list and rotate it to the end of the list, so the next copy thread gets the next item on the list. Once we have our stream context assigned, we will check if it has a destination file assigned to it, if not we will do it via ds_open this time passing out destination datasink (either FIFO or STDOUT).
+FIFO datasink has a list of FIFO files it can assign on an open operation, when xbstream::ds_open calls FIFO::ds_open we assign the first FIFO file in the list and from this point forward that xbstream ds_stream_ctxt_t will be bound to this particular FIFO file. We do not assign the fifo file back to the list of fifo files as xbstream ds_stream_ctxt_t->fifo is a 1:1 mapping
+Note: an xbstream_close operation only ensures we write any remaining buffer data to FIFO/STDOUT, we do not close the FD assigned to that stream context. We only remove the binding of stream context to FIFO file at xbstream_deinit
+next xtrabackup_copy_datafile thread, will then go again to xbstream_open and this time it will get the next stream context from the list and same workflow will happen until the first stream context is reached again. When it happens, we will get a stream context assigned that already has a stream_ctxt->dest_file assigned and this is where we will start having multiple copy threads writing in parallel to the same stream
+In summary, the relationship is:
+ * Copy threads to stream context - N:1 - Multiple copy threads are assigned to the same xbstream context (this is done by the LRU pop_from & push_back)
+ * xbstream context -> FIFO file - 1:1 - One stream context to one fifo file, once a stream context has a fifo file assigned to it, it will remain bound to this FIFO until deinit
+Workflow:
+Before:
+ * copy_thread->stream_context->stdout
+Now:
+ * copy_thread->parallel_stream_ctxt->stream_context->fifo
+
+*/
+#ifndef DS_FIFO_H
+#define DS_FIFO_H
+
+#include "datasink.h"
+
+extern uint ds_fifo_threads;
+
+extern datasink_t datasink_fifo;
+
+#endif
--- a/storage/innobase/xtrabackup/src/ds_xbstream.cc
+++ b/storage/innobase/xtrabackup/src/ds_xbstream.cc
@@ -1,5 +1,5 @@
 /******************************************************
-Copyright (c) 2011-2013 Percona LLC and/or its affiliates.
+Copyright (c) 2011-2023 Percona LLC and/or its affiliates.
 
 Streaming implementation for XtraBackup.
 
@@ -22,6 +22,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include <my_io.h>
 #include <mysql/service_mysql_alloc.h>
 #include <mysql_version.h>
+#include <list>
+#include <mutex>
 #include "common.h"
 #include "datasink.h"
 #include "msg.h"
@@ -30,14 +32,20 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 typedef struct {
   xb_wstream_t *xbstream;
   ds_file_t *dest_file;
-  pthread_mutex_t mutex;
+  std::mutex mutex;
 } ds_stream_ctxt_t;
+
+typedef struct {
+  std::list<ds_stream_ctxt_t *> ctx_list;
+  std::mutex mutex;
+} ds_parallel_stream_ctxt_t;
 
 typedef struct {
   xb_wstream_file_t *xbstream_file;
   ds_stream_ctxt_t *stream_ctxt;
 } ds_stream_file_t;
 
+extern uint xtrabackup_fifo_streams;
 /***********************************************************************
 General streaming interface */
 
@@ -75,28 +83,29 @@ static ssize_t my_xbstream_write_callback(xb_wstream_file_t *f
 
 static ds_ctxt_t *xbstream_init(const char *root __attribute__((unused))) {
   ds_ctxt_t *ctxt;
-  ds_stream_ctxt_t *stream_ctxt;
-  xb_wstream_t *xbstream;
+  ds_parallel_stream_ctxt_t *parallel_stream_ctxt =
+      new ds_parallel_stream_ctxt_t;
 
   ctxt = static_cast<ds_ctxt_t *>(
       my_malloc(PSI_NOT_INSTRUMENTED,
-                sizeof(ds_ctxt_t) + sizeof(ds_stream_ctxt_t), MYF(MY_FAE)));
-  stream_ctxt = (ds_stream_ctxt_t *)(ctxt + 1);
+                sizeof(ds_ctxt_t) + sizeof(ds_parallel_stream_ctxt_t) +
+                    (sizeof(ds_stream_ctxt_t) * (xtrabackup_fifo_streams + 1)),
+                MYF(MY_FAE)));
 
-  if (pthread_mutex_init(&stream_ctxt->mutex, NULL)) {
-    msg("xbstream_init: pthread_mutex_init() failed.\n");
-    goto err;
+  for (uint i = 0; i < xtrabackup_fifo_streams; i++) {
+    ds_stream_ctxt_t *stream_ctxt = new ds_stream_ctxt_t;
+    xb_wstream_t *xbstream;
+    xbstream = xb_stream_write_new();
+    if (xbstream == NULL) {
+      msg("xb_stream_write_new() failed.\n");
+      goto err;
+    }
+    stream_ctxt->xbstream = xbstream;
+    stream_ctxt->dest_file = NULL;
+    parallel_stream_ctxt->ctx_list.push_back(stream_ctxt);
   }
 
-  xbstream = xb_stream_write_new();
-  if (xbstream == NULL) {
-    msg("xb_stream_write_new() failed.\n");
-    goto err;
-  }
-  stream_ctxt->xbstream = xbstream;
-  stream_ctxt->dest_file = NULL;
-
-  ctxt->ptr = stream_ctxt;
+  ctxt->ptr = parallel_stream_ctxt;
 
   return ctxt;
 
@@ -108,6 +117,7 @@ err:
 static ds_file_t *xbstream_open(ds_ctxt_t *ctxt, const char *path,
                                 MY_STAT *mystat) {
   ds_file_t *file;
+  ds_parallel_stream_ctxt_t *parallel_stream_ctxt;
   ds_stream_file_t *stream_file;
   ds_stream_ctxt_t *stream_ctxt;
   ds_ctxt_t *dest_ctxt;
@@ -117,16 +127,21 @@ static ds_file_t *xbstream_open(ds_ctxt_t *ctxt, const char *path,
   xb_ad(ctxt->pipe_ctxt != NULL);
   dest_ctxt = ctxt->pipe_ctxt;
 
-  stream_ctxt = (ds_stream_ctxt_t *)ctxt->ptr;
+  parallel_stream_ctxt = (ds_parallel_stream_ctxt_t *)ctxt->ptr;
+  parallel_stream_ctxt->mutex.lock();
+  stream_ctxt = parallel_stream_ctxt->ctx_list.front();
+  parallel_stream_ctxt->ctx_list.pop_front();
+  parallel_stream_ctxt->ctx_list.push_back(stream_ctxt);
+  parallel_stream_ctxt->mutex.unlock();
 
-  pthread_mutex_lock(&stream_ctxt->mutex);
+  stream_ctxt->mutex.lock();
   if (stream_ctxt->dest_file == NULL) {
     stream_ctxt->dest_file = ds_open(dest_ctxt, path, mystat);
     if (stream_ctxt->dest_file == NULL) {
       return NULL;
     }
   }
-  pthread_mutex_unlock(&stream_ctxt->mutex);
+  stream_ctxt->mutex.unlock();
 
   file = (ds_file_t *)my_malloc(PSI_NOT_INSTRUMENTED,
                                 sizeof(ds_file_t) + sizeof(ds_stream_file_t),
@@ -210,20 +225,23 @@ static int xbstream_close(ds_file_t *file) {
 }
 
 static void xbstream_deinit(ds_ctxt_t *ctxt) {
+  ds_parallel_stream_ctxt_t *parallel_stream_ctxt =
+      (ds_parallel_stream_ctxt_t *)ctxt->ptr;
   ds_stream_ctxt_t *stream_ctxt;
 
-  stream_ctxt = (ds_stream_ctxt_t *)ctxt->ptr;
+  for (uint i = 0; i < xtrabackup_fifo_streams; i++) {
+    stream_ctxt = parallel_stream_ctxt->ctx_list.front();
+    parallel_stream_ctxt->ctx_list.pop_front();
+    if (xb_stream_write_done(stream_ctxt->xbstream)) {
+      msg("xb_stream_done() failed.\n");
+    }
 
-  if (xb_stream_write_done(stream_ctxt->xbstream)) {
-    msg("xb_stream_done() failed.\n");
+    if (stream_ctxt->dest_file) {
+      ds_close(stream_ctxt->dest_file);
+      stream_ctxt->dest_file = NULL;
+    }
+    delete stream_ctxt;
   }
-
-  if (stream_ctxt->dest_file) {
-    ds_close(stream_ctxt->dest_file);
-    stream_ctxt->dest_file = NULL;
-  }
-
-  pthread_mutex_destroy(&stream_ctxt->mutex);
-
+  delete parallel_stream_ctxt;
   my_free(ctxt);
 }
--- a/storage/innobase/xtrabackup/src/file_utils.cc
+++ b/storage/innobase/xtrabackup/src/file_utils.cc
@@ -1,5 +1,27 @@
+/******************************************************
+Copyright (c) 2021-2023 Percona LLC and/or its affiliates.
+
+Streaming implementation for XtraBackup.
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; version 2 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+*******************************************************/
+
 #include "file_utils.h"
 #include <mysql/service_mysql_alloc.h>
+#include <sys/epoll.h>
+#include <thread>
 #include "common.h"
 #include "msg.h"
 #include "my_dir.h"
@@ -314,4 +336,63 @@ bool restore_sparseness(const char *src_file_path, uint buffer_size,
   }
   datafile_close(&cursor);
   return true;
+}
+
+File open_fifo_for_write_with_timeout(const char *path, uint timeout) {
+  File fd;
+  uint attempt = 0;
+  do {
+    fd = my_open(path, O_WRONLY | O_NONBLOCK, MYF(0));
+    if (fd < 0) {
+      attempt++;
+      std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+
+  } while (fd < 0 && attempt < timeout);
+  if (fd < 0) {
+    return -1;
+  }
+  /* Adjust file flags to blocking - we are in sync mode now. */
+  if (fcntl(fd, F_SETFL, O_WRONLY) == -1) {
+    my_close(fd, MYF(0));
+    return -1;
+  }
+
+  return fd;
+}
+
+File open_fifo_for_read_with_timeout(const char *path, uint timeout) {
+  int fd;
+  uint attempt = 0;
+  do {
+    fd = open(path, O_RDONLY | O_NONBLOCK);
+    if (fd < 0) {
+      attempt++;
+      std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+  } while (fd < 0 && attempt < timeout);
+
+  if (fd < 0) {
+    return -1;
+  }
+
+  /* File was open, lets check its open on the other side */
+  int epoll_fd = epoll_create1(0);
+  if (epoll_fd < 0) {
+    return -1;
+  }
+  struct epoll_event ev;
+  ev.events = EPOLLIN;
+  ev.data.fd = fd;
+  if (epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &ev) < 0) {
+    return -1;
+  }
+  int ret = epoll_wait(epoll_fd, &ev, 1, timeout * 1000);
+  if (ret <= 0) {
+    return -1;
+  }
+  if (fcntl(fd, F_SETFL, O_RDONLY) == -1) {
+    return -1;
+  }
+  return fd;
 }
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -1,5 +1,5 @@
 /******************************************************
-Copyright (c) 2014, 2021 Percona LLC and/or its affiliates.
+Copyright (c) 2014, 2023 Percona LLC and/or its affiliates.
 
 xbcloud utility. Manage backups on cloud storage services.
 
@@ -23,11 +23,13 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include <my_dir.h>
 #include <my_getopt.h>
 #include <my_sys.h>
+#include <my_thread_local.h>
 #include <mysql/service_mysql_alloc.h>
 #include <signal.h>
 #include <typelib.h>
 #include <fstream>
 #include <iostream>
+#include <list>
 #include <set>
 #include <string>
 #include <unordered_map>
@@ -36,6 +38,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include <curl/curl.h>
 
 #include "common.h"
+#include "file_utils.h"
 #include "xbstream.h"
 #include "xtrabackup_version.h"
 
@@ -47,6 +50,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "xbcloud/s3_ec2.h"
 #include "xbcloud/swift.h"
 #include "xbcloud/util.h"
+#include "xbcloud/xbcloud.h"
 #include "xbcrypt_common.h"
 
 using namespace xbcloud;
@@ -57,6 +61,8 @@ using namespace xbcloud;
 /*****************************************************************************/
 
 const char *config_file = "my"; /* Default config file */
+
+const static int chunk_index_prefix_len = 20;
 
 const char *azure_development_access_key =
     "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/"
@@ -119,6 +125,9 @@ static bool opt_azure_development_storage = 0;
 static std::string backup_name;
 static char *opt_cacert = nullptr;
 static ulong opt_parallel = 1;
+static ulong opt_threads = 1;
+static char *opt_fifo_dir = nullptr;
+static ulong opt_fifo_timeout = 60;
 static ulong opt_max_retries = 10;
 static u_int32_t opt_max_backoff = 300000;
 
@@ -190,6 +199,9 @@ enum {
   OPT_GOOGLE_BUCKET,
 
   OPT_PARALLEL,
+  OPT_THREADS,
+  OPT_FIFO_DIR,
+  OPT_FIFO_TIMEOUT,
   OPT_MAX_RETRIES,
   OPT_MAX_BACKOFF,
   OPT_CACERT,
@@ -293,6 +305,22 @@ static struct my_option my_long_options[] = {
     {"parallel", OPT_PARALLEL, "Number of parallel chunk uploads.",
      &opt_parallel, &opt_parallel, 0, GET_ULONG, REQUIRED_ARG, 1, 1, ULONG_MAX,
      0, 0, 0},
+
+    {"fifo-streams", OPT_THREADS, "Number of parallel fifo stream threads.",
+     &opt_threads, &opt_threads, 0, GET_ULONG, REQUIRED_ARG, 1, 1, ULONG_MAX, 0,
+     0, 0},
+
+    {"fifo-dir", OPT_FIFO_DIR,
+     "Directory to read/write Named Pipe. On put mode, xbcloud read from named "
+     "pipes. On get mode, xbcloud writes to named pipes.",
+     &opt_fifo_dir, &opt_fifo_dir, 0, GET_STR_ALLOC, REQUIRED_ARG, 0, 0, 0, 0,
+     0, 0},
+
+    {"fifo-timeout", OPT_FIFO_TIMEOUT,
+     "How many seconds to wait for other end to open the stream. "
+     "Default 60 seconds",
+     &opt_fifo_timeout, &opt_fifo_timeout, 0, GET_INT, REQUIRED_ARG, 60, 1,
+     INT_MAX, 0, 0, 0},
 
     {"max-retries", OPT_MAX_RETRIES,
      "Number of retries of chunk uploads/downloads after a failure (Default "
@@ -708,59 +736,85 @@ static bool parse_args(int argc, char **argv) {
   return (0);
 }
 
-bool xbcloud_put(Object_store *store, const std::string &container,
-                 const std::string &backup_name, bool upload_md5) {
-  bool exists;
+typedef struct {
+  /* used to create FIFO file and print thread number to log */
+  uint thread_id;
+  std::atomic<bool> *has_errors;
+  struct global_list_t *global_list;
+  const std::string *container;
+  Object_store *store;
+} download_thread_ctxt_t;
 
-  if (!store->container_exists(container, exists)) {
-    return false;
-  }
+typedef struct {
+  /* used to create FIFO file and print thread number to log */
+  uint thread_id;
+  std::atomic<bool> *has_errors;
+  Http_buffer *buf_md5;
+  const std::string *container;
+  Object_store *store;
+} put_thread_ctxt_t;
 
-  if (!exists) {
-    if (!store->create_container(container)) {
-      return false;
-    }
-  }
+struct file_entry_t {
+  my_off_t chunk_idx;
+  my_off_t offset;
+  std::string path;
+};
 
-  std::vector<std::string> object_list;
-  if (!store->list_objects_in_directory(container, backup_name, object_list)) {
-    return false;
-  }
+/**
+  Build filename with proper chunk index length.
 
-  if (!object_list.empty()) {
-    msg_ts("%s: error: backup named %s already exists!\n", my_progname,
-           backup_name.c_str());
-    return false;
-  }
+@param [in]    file_name   filename part of the chunk
+@param [in]    idx         chunk index
 
-  struct file_entry_t {
-    my_off_t chunk_idx;
-    my_off_t offset;
-    std::string path;
-  };
+@return string containing the file name */
+std::string build_file_name(const std::string &file_name, my_off_t idx) {
+  std::stringstream file_name_s;
+  file_name_s << file_name << "." << std::setw(chunk_index_prefix_len)
+              << std::setfill('0') << idx;
+  return file_name_s.str();
+}
 
+void put_func(put_thread_ctxt_t &cntx) {
+  std::thread ev;
+  Event_handler h(opt_parallel > 0 ? opt_parallel : 1);
   std::unordered_map<std::string, std::unique_ptr<file_entry_t>> filehash;
-  xb_rstream_t *stream = xb_stream_read_new();
-  if (stream == nullptr) {
-    msg_ts("%s: xb_stream_read_new() failed.", my_progname);
-    return false;
+  xb_rstream_t *stream;
+  if (opt_threads > 1) {
+    char filename[FN_REFLEN];
+    snprintf(filename, sizeof(filename), "%s%s%lu", opt_fifo_dir, "/thread_",
+             (ulong)cntx.thread_id);
+    stream = xb_stream_read_new_fifo(filename, opt_fifo_timeout);
+    if (stream == nullptr) {
+      msg_ts(
+          "%s: xb_stream_read_new_fifo() failed for thread %d. Possibly sender "
+          "did "
+          "not start.\n",
+          my_progname, cntx.thread_id);
+      cntx.has_errors->store(true);
+      goto end;
+    }
+
+  } else {
+    stream = xb_stream_read_new_stdin();
+    if (stream == nullptr) {
+      msg_ts("%s: xb_stream_read_new_stdin() failed.\n", my_progname);
+      cntx.has_errors->store(true);
+      goto end;
+    }
   }
 
   xb_rstream_chunk_t chunk;
   xb_rstream_result_t res;
 
-  bool has_errors = false;
 
   memset(&chunk, 0, sizeof(chunk));
 
-  Http_buffer buf_md5;
 
-  Event_handler h(opt_parallel > 0 ? opt_parallel : 1);
   if (!h.init()) {
     msg_ts("%s: Failed to initialize event handler.\n", my_progname);
-    return false;
+    goto end;
   }
-  auto thread = h.run();
+  ev = h.run();
 
   do {
     res = xb_stream_read_chunk(stream, &chunk);
@@ -796,10 +850,7 @@ bool xbcloud_put(Object_store *store, const std::string &container,
       }
     }
 
-    std::stringstream file_name_s;
-    file_name_s << chunk.path << "." << std::setw(20) << std::setfill('0')
-                << entry->chunk_idx;
-    std::string file_name = file_name_s.str();
+    std::string file_name = build_file_name(chunk.path, entry->chunk_idx);
     std::string object_name = backup_name;
     object_name.append("/").append(file_name);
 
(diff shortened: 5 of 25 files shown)


storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -315,3 +317,62 @@ bool restore_sparseness(const char *src_file_path, uint buffer_size,
...
+  /* Adjust file flags to blocking - we are in sync mode now. */
+  if (fcntl(fd, F_SETFL, O_WRONLY) == -1) {
+    my_close(fd, MYF(0));
+    return -1;
+  }
+
+  return fd;
+}
+
+int open_fifo_for_read_with_timeout(const char *path, uint timeout) {
```
rahulmalik87: declaration and definition for open_fifo_for_read_with_timeout does not match.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3517,9 +3543,20 @@ files first, and then streams them in a serialized way when closed. */
 static void xtrabackup_init_datasinks(void) {
   /* Start building out the pipelines from the terminus back */
   if (xtrabackup_stream) {
-    /* All streaming goes to stdout */
-    ds_data = ds_meta = ds_redo =
-        ds_create(xtrabackup_target_dir, DS_TYPE_STDOUT);
+    if (xtrabackup_parallel_stream > 1) {
+      /* Use Named PIPEs */
+      xb::info() << "Created " << xtrabackup_parallel_stream
```
rahulmalik87: Is stream created as part of ds_correct right ? if so, this message should be printer after creating the stream. or we can mention creating the stream and would wait for.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8127,6 +8175,12 @@ int main(int argc, char **argv) {
     exit(EXIT_FAILURE);
   }
 
+  if (xtrabackup_parallel_stream_set && xtrabackup_stream_dir == NULL &&
+      strcmp(xtrabackup_target_dir, "./xtrabackup_backupfiles/") == 0) {
```
rahulmalik87, line 8188: why are we comparing if `xtrabackup_target_dir` is `xtrabackup_backupfiles`

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8127,6 +8175,12 @@ int main(int argc, char **argv) {
...
+    xb::error() << "Option --parallel-stream requires --stream-dir to be set.";
+    exit(EXIT_FAILURE);
```
rahulmalik87: what if directory contains old residue files of some killed backup?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,167 @@
...
+
+static ds_ctxt_t *fifo_init(const char *root) {
+  ds_fifo_ctxt_t *fifo_context = new ds_fifo_ctxt_t;
+  File fd;
+  ds_ctxt_t *ctxt = new ds_ctxt_t;
+  char fullpath[FN_REFLEN];
+
+  if (pthread_mutex_init(&fifo_context->fifo_mutex, NULL)) {
+    msg("fifo_init: pthread_mutex_init() failed.\n");
+    return NULL;
```
rahulmalik87: Please use `nullptr` instead of `null`

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,167 @@
...
+    char errbuf[MYSYS_STRERROR_SIZE];
+    my_error(EE_CANT_MKDIR, MYF(ME_BELL), root, my_errno(),
+             my_strerror(errbuf, sizeof(errbuf), my_errno()));
+    return NULL;
+  }
+
+  for (uint i = 0; i < xtrabackup_parallel_stream; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    mkfifo(fullpath, 0666);
```
rahulmalik87: Handle error in case mkfifo FAIL.

storage/innobase/xtrabackup/src/ds_xbstream.cc
```diff
@@ -75,28 +82,38 @@ static ssize_t my_xbstream_write_callback(xb_wstream_file_t *f
...
-  stream_ctxt = (ds_stream_ctxt_t *)(ctxt + 1);
+  ctxt = static_cast<ds_ctxt_t *>(my_malloc(
+      PSI_NOT_INSTRUMENTED,
+      sizeof(ds_ctxt_t) + sizeof(ds_parallel_stream_ctxt_t) +
+          (sizeof(ds_stream_ctxt_t) * (xtrabackup_parallel_stream + 1)),
+      MYF(MY_FAE)));
 
-  if (pthread_mutex_init(&stream_ctxt->mutex, NULL)) {
+  if (pthread_mutex_init(&parallel_stream_ctxt->mutex, NULL)) {
     msg("xbstream_init: pthread_mutex_init() failed.\n");
```
rahulmalik87: Can we write a different message when fail to acquire `parallel_stream_ctxt` vs `stream_ctxt` mutext

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  for (uint i = 0; i < xtrabackup_parallel_stream; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    fd = open_fifo_for_write_with_timeout(fullpath, xtrabackup_stream_timeout);
```
rahulmalik87: Any warning or error it it gets timeout?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1368,6 +1375,22 @@ struct my_option xb_client_options[] = {
      (G_PTR *)&xtrabackup_parallel, (G_PTR *)&xtrabackup_parallel, 0, GET_INT,
      REQUIRED_ARG, 1, 1, INT_MAX, 0, 0, 0},
 
+    {"parallel-stream", OPT_XTRA_PARALLEL_STREAM,
+     "Number of FIFO files to use for parallel datafiles stream. Setting this "
+     "parameter to 1 disables FIFO and stream is sent to STDOUT.",
+     (G_PTR *)&xtrabackup_parallel_stream, (G_PTR *)&xtrabackup_parallel_stream,
+     0, GET_INT, REQUIRED_ARG, 1, 1, INT_MAX, 0, 0, 0},
+
+    {"stream-dir", OPT_XTRA_STREAM_DIR, "Directory to write Named Pipe.",
```
rahulmalik87: if use has specified --target-dir then it takes that value correct? Can we specify the same.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3517,9 +3543,20 @@ files first, and then streams them in a serialized way when closed. */
 static void xtrabackup_init_datasinks(void) {
   /* Start building out the pipelines from the terminus back */
   if (xtrabackup_stream) {
-    /* All streaming goes to stdout */
-    ds_data = ds_meta = ds_redo =
-        ds_create(xtrabackup_target_dir, DS_TYPE_STDOUT);
+    if (xtrabackup_parallel_stream > 1) {
+      /* Use Named PIPEs */
+      xb::info() << "Created " << xtrabackup_parallel_stream
+                 << " Named Pipes(FIFO). Waiting up to "
```
rahulmalik87: Can we also mention the directory where the stream is created?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4416,6 +4453,17 @@ void xtrabackup_backup_func(void) {
     exit(EXIT_FAILURE);
   }
 
+  if (xtrabackup_parallel_stream_set && xtrabackup_stream_dir != NULL) {
+    xtrabackup_target_dir = xtrabackup_stream_dir;
+  }
+
+  if (xtrabackup_parallel_stream_set && !xtrabackup_stream) {
+    xb::info() << "Option --parallel-stream require xbstream format. Setting "
```
rahulmalik87: Should we error out if some one is using option --parallel-stream without specifying --stream=xbstream?

storage/innobase/xtrabackup/src/ds_fifo.cc
rahulmalik87: Do we delete this FIFO file at the end of the backup?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -836,29 +888,91 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+      PSI_NOT_INSTRUMENTED, sizeof(put_thread_ctxt_t) * (opt_threads + 1),
+      MYF(MY_FAE));
+  threads = new pthread_t[opt_threads];
+  for (uint i = 0; i < (uint)opt_threads; i++) {
+    data_threads[i].num = i;
+    data_threads[i].has_errors = &has_errors;
+    data_threads[i].store = store;
+    data_threads[i].container = &container;
+    data_threads[i].buf_md5 = new Http_buffer();
+    pthread_create(&threads[i], nullptr, &put_func, data_threads + i);
```
rahulmalik87: handle the error if pthread_create fails.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -960,136 +1084,187 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
...
+      PSI_NOT_INSTRUMENTED, sizeof(download_thread_ctxt_t) * (opt_threads + 1),
+      MYF(MY_FAE));
+  threads = new pthread_t[opt_threads + 1];
+  for (uint i = 0; i < (uint)opt_threads; i++) {
+    data_threads[i].num = i;
+    data_threads[i].has_errors = &has_errors;
+    data_threads[i].file_list = file_list;
+    data_threads[i].store = store;
+    data_threads[i].container = &container;
+    pthread_create(&threads[i], nullptr, &download_func, data_threads + i);
```
rahulmalik87: handle error if pthread_create fails.

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -668,31 +726,27 @@ static int mode_extract(int n_threads, int argc __attribute__((unused)),
...
-  ctxt.mutex = &mutex;
-
-  tids = new pthread_t[n_threads];
-  retvals = new void *[n_threads];
+  tids = new pthread_t[n_threads + 1];
+  retvals = new void *[n_threads + 1];
 
   for (i = 0; i < n_threads; i++)
-    pthread_create(tids + i, NULL, extract_worker_thread_func, &ctxt);
+    pthread_create(tids + i, NULL, extract_worker_thread_func,
```
rahulmalik87: handle error of pthread_create

rahulmalik87 (changes requested):
Thank you @altmannmarcelo for your work.   The design and usability are really great. some minor comments.

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Right, `ds_data = ds_meta = ds_redo = ds_create(xtrabackup_target_dir, DS_TYPE_FIFO);` Will hang until the other side has opened the FIFO for read. So the message needs to be issued before `ds_create`. I changed to `Creating xx Named Pipes... Waiting up to...`

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Because that is the default value for xtrabackup_target_dir (it inherits from xtrabackup_real_target_dir). Here we are erroring out if neither --stream-dir nor --target-dir has been specified.

--stream-dir I have added to be compatible with xbcloud and xbstream, so all tree tools use the same parameters
--target-dir I left as an option for backward compatibility.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: If the target dir has files, we will continue creating the thread_x FIFO file. If a file with the same name exists and we can open it for writing we will use it.

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: This requires a broader refactor.

ds_create from datasink.cc calls ds->init() and later it tests if the return is NULL. To fix this we need to adjust all datasinks.

For reference - https://github.com/percona/percona-xtrabackup/blob/release-8.0.32-26/storage/innobase/xtrabackup/src/datasink.cc#L92

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: > Handle error in case mkfifo FAIL.

Done

> Do we delete this FIFO file at the end of the backup?

The reader side does it (xbcloud put or xbstream)

storage/innobase/xtrabackup/src/ds_xbstream.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Yes, we print the error in the caller due to the return of fifo_init been NULL

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: I would rather do it the other way around. During developing this feature, I found it redundant to have to specify --stream=xbstream. If I am setting parallel-stream I want the output to be xbstream.
Another thing that I changed as part of this JIRA is to allow --stream without a parameter, we have only 1 stream format on 8.0, so we should assume it if someone sets --stream without an argument.

In other words, those changes enhance the user experience. (don't make me type two parameters if I can type only one ;) )

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done. `has_errors` will signal already started threads to cleanup and exit.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done. `has_errors` will signal already started threads to cleanup and exit.

storage/innobase/xtrabackup/src/xbstream.cc
altmannmarcelo: Done. `pthread_cancel` will signal already started threads to cleanup and exit.

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+#include <mysys_err.h>
+#include <map>
+#include "common.h"
+#include "datasink.h"
+#include "file_utils.h"
+#include "msg.h"
+
+typedef struct {
+  /* List of FIFO files to be used for stream */
+  std::map<std::string, File> FIFO_list;
```
satya-bodapati: do we do random access based on key? Do we need the list to be sorted? If the answer to these questions is no, should we consider  unordered_map? std::vector?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+      msg("mkfifo(%s) failed with error %d", fullpath, errno);
+      return NULL;
+    }
+  }
+
+  for (uint i = 0; i < xtrabackup_parallel_stream; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    fd = open_fifo_for_write_with_timeout(fullpath, xtrabackup_stream_timeout);
+    if (fd < 0) return NULL;
```
satya-bodapati: If there is an error, should we ensure all opened fifo so far are closed?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+    char errbuf[MYSYS_STRERROR_SIZE];
+    my_error(EE_CANT_MKDIR, MYF(ME_BELL), root, my_errno(),
+             my_strerror(errbuf, sizeof(errbuf), my_errno()));
+    return NULL;
+  }
+
+  for (uint i = 0; i < xtrabackup_parallel_stream; i++) {
+    std::string path = "thread_" + std::to_string(i);
+    fn_format(fullpath, path.c_str(), root, "", MYF(MY_RELATIVE_PATH));
+    if (mkfifo(fullpath, 0666) < 0) {
```
satya-bodapati: If there is an error in the middle, should we ensure all FIFO created are destroyed?

satya-bodapati: We can use RAII sentry object.

sql/raii/sentry.h

satya-bodapati: raii::Sentry<> delete_on_error{[&]() {
 for...
}};

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  File fd;
+  ds_ctxt_t *ctxt = new ds_ctxt_t;
+  char fullpath[FN_REFLEN];
+
+  if (pthread_mutex_init(&fifo_context->fifo_mutex, NULL)) {
+    msg("fifo_init: pthread_mutex_init() failed.\n");
+    return NULL;
+  }
+
+  if (my_mkdir(root, 0777, MYF(0)) < 0 && my_errno() != EEXIST &&
```
satya-bodapati: directory is created with 777 permission ?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+static ds_file_t *fifo_open(ds_ctxt_t *ctxt,
+                            const char *path __attribute__((unused)),
+                            MY_STAT *mystat __attribute__((unused))) {
+  char fullpath[FN_REFLEN];
+  size_t path_len;
+  ds_fifo_file_t *fifo_file;
+  ds_file_t *file;
+  ds_fifo_ctxt_t *fifo_context = (ds_fifo_ctxt_t *)ctxt->ptr;
+  File fd;
+  std::string fifo_path;
```
satya-bodapati: Move these variable declarations close to their usage. Only goto statements create problems with late declarations.

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+
+  if (pthread_mutex_init(&fifo_context->fifo_mutex, NULL)) {
+    msg("fifo_init: pthread_mutex_init() failed.\n");
+    return NULL;
+  }
+
+  if (my_mkdir(root, 0777, MYF(0)) < 0 && my_errno() != EEXIST &&
+      my_errno() != EISDIR) {
+    char errbuf[MYSYS_STRERROR_SIZE];
+    my_error(EE_CANT_MKDIR, MYF(ME_BELL), root, my_errno(),
```
satya-bodapati: ME_BELL?
#define ME_BELL 4          /* DEPRECATED: Ring bell then printing message */
ME_ERRORLOG is sufficient?

Would anything be logged if we use MYF(0)?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+static ds_ctxt_t *fifo_init(const char *root);
+static ds_file_t *fifo_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *mystat);
+static int fifo_write(ds_file_t *file, const void *buf, size_t len);
+static int fifo_close(ds_file_t *file);
+static void fifo_deinit(ds_ctxt_t *ctxt);
+
+datasink_t datasink_fifo = {&fifo_init, &fifo_open,  &fifo_write,
+                            nullptr,    &fifo_close, &fifo_deinit};
+
+static ds_ctxt_t *fifo_init(const char *root) {
```
satya-bodapati, line 97: root will be the directory passed by the user? Can you please add a comment here?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  std::string fifo_path;
+  pthread_mutex_lock(&fifo_context->fifo_mutex);
+  auto fifo_it = fifo_context->FIFO_list.begin();
+  if (fifo_it == fifo_context->FIFO_list.end()) {
+    pthread_mutex_unlock(&fifo_context->fifo_mutex);
+    return NULL;
+  }
+  fifo_path = fifo_it->first;
+  fd = fifo_it->second;
+  fn_format(fullpath, fifo_path.c_str(), ctxt->root, "", MYF(MY_RELATIVE_PATH));
```
satya-bodapati: Why are we constructing fullpath again?  FIFO_list is filled with the fullpath and named pipe descriptor already here:

fifo_context->FIFO_list.insert({fullpath, fd});

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  fifo_context->FIFO_list.erase(fifo_it);
+  pthread_mutex_unlock(&fifo_context->fifo_mutex);
+
+  path_len = strlen(fullpath) + 1; /* terminating '\0' */
+
+  file = (ds_file_t *)my_malloc(
+      PSI_NOT_INSTRUMENTED,
+      sizeof(ds_file_t) + sizeof(ds_fifo_file_t) + path_len, MYF(MY_FAE));
```
satya-bodapati: why do we allocate the memory of three structs? file is of type of ds_file_t?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  fifo_file = (ds_fifo_file_t *)(file + 1);
```
satya-bodapati: file +1 ? I dont understand this.

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+
+  fifo_file->fd = fd;
+  fifo_file->fifo_context = fifo_context;
+  fifo_file->path = (char *)fifo_path.c_str();
+  file->path = (char *)fifo_file + sizeof(ds_fifo_file_t);
+  memcpy(file->path, fullpath, path_len);
+
+  file->ptr = fifo_file;
```
satya-bodapati, line 166: file->datasink is not filled here?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+
+  return file;
+}
+
+static int fifo_write(ds_file_t *file, const void *buf, size_t len) {
```
satya-bodapati, line 171: bool is better return type? We are not returning the number of bytes written.

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  if (!my_write(fd, static_cast<const uchar *>(buf), len,
+                MYF(MY_WME | MY_NABP))) {
+    posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED);
+    return 0;
+  }
+
+  return 1;
+}
+
+static int fifo_close(ds_file_t *file) {
```
satya-bodapati, line 183: void return type?

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+#include "datasink.h"
+#include "file_utils.h"
+#include "msg.h"
+
+typedef struct {
+  /* List of FIFO files to be used for stream */
+  std::map<std::string, File> FIFO_list;
+  /* Mutex protecting FIFO list */
+  pthread_mutex_t fifo_mutex;
+} ds_fifo_ctxt_t;
```
satya-bodapati, line 60: after reading a bit of the below code, it seems to me, this class should be FIFO manager.
ie. it should be responsible for FIFO objects. And some of the methods should
be  allocate_from_list() (or allocate()). and close()/return_to_list()),  populate_list();

storage/innobase/xtrabackup/src/ds_fifo.cc
```diff
@@ -0,0 +1,170 @@
...
+  pthread_mutex_unlock(&fifo_context->fifo_mutex);
+  my_free(file);
+
+  return 0;
+}
+
+static void fifo_deinit(ds_ctxt_t *ctxt) {
+  ds_fifo_ctxt_t *fifo_context = (ds_fifo_ctxt_t *)ctxt->ptr;
+
+  pthread_mutex_destroy(&fifo_context->fifo_mutex);
```
satya-bodapati: If FIFO data ctxt is a class, it would be responsible for destroying its internal mutex etc. Also, we could validate that all objects given to FIFO files are returned before deinit.

ie, before destroying the list size should be the same as the size at the time of initialization.

storage/innobase/xtrabackup/src/ds_xbstream.cc
```diff
@@ -75,28 +82,39 @@ static ssize_t my_xbstream_write_callback(xb_wstream_file_t *f
...
-  xbstream = xb_stream_write_new();
-  if (xbstream == NULL) {
-    msg("xb_stream_write_new() failed.\n");
-    goto err;
+  for (uint i = 0; i < xtrabackup_parallel_stream; i++) {
+    ds_stream_ctxt_t *stream_ctxt = new ds_stream_ctxt_t;
+    xb_wstream_t *xbstream;
+    if (pthread_mutex_init(&stream_ctxt->mutex, NULL)) {
+      msg("xbstream_init: pthread_mutex_init() failed for stream_ctxt.\n");
+      goto err;
```
satya-bodapati: Please check how we will handle opened objects so far if we fail in middle of the loop.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,59 +736,83 @@ static bool parse_args(int argc, char **argv) {
...
-  }
-
-  if (!exists) {
-    if (!store->create_container(container)) {
-      return false;
-    }
-  }
+typedef struct {
+  uint num;
+  std::atomic<bool> *has_errors;
```
satya-bodapati, line 742: do we need a pointer to std::atomic<bool>? can we simply use std::atomic<bool> (is there some initialization issue?)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -836,29 +888,98 @@ bool xbcloud_put(Object_store *store, const std::string &container,
...
+  } while (!cntx->has_errors->load());
 
   h.stop();
-  thread.join();
+  ev.join();
 
-  xb_stream_read_done(stream);
+end:
+  if (stream != nullptr) xb_stream_read_done(stream);
+  pthread_exit(NULL);
```
satya-bodapati: any significance of pthread_exit(NULL)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
satya-bodapati: whenever possible, lets move to std::threads

storage/innobase/xtrabackup/src/ds_fifo.cc
satya-bodapati: [optional suggestion]

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
satya-bodapati: [optional]

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
```diff
@@ -0,0 +1,186 @@
...
+  my_off_t last_chunk;
+  my_off_t next_chunk;
+  std::string name;
+};
+
+/** struct to hold state of current in progress files. This avoids multiple
+ * chunks from the same file to be downloaded in parallel */
+struct download_state {
+  std::mutex m;
+  /* files with active chunks been downloaded (waiting http client to complete
```
satya-bodapati, line 31: can there be multiple threads downloading multiple chunks in same file?

satya-bodapati (changes requested):
Xbstream format has a limitation in that it requires that each chunk of
a file is downloaded/extracted in order. For example, chunk 3 of the
file X cannot arrive before chunks 0, 1 and 2.

How do we solve this problem in this PR?

storage/innobase/xtrabackup/src/ds_fifo.cc
satya-bodapati: [optional]

storage/innobase/xtrabackup/src/ds_fifo.cc
satya-bodapati: I guess we are trying to minimize the number of malloc() calls?

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: xtrabackup will terminate in case of error here

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Done. Also changed the FIFO to 600

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: fixed

storage/innobase/xtrabackup/src/ds_fifo.cc
satya-bodapati: I guess we are trying to minimize the number of malloc() calls?

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.cc
satya-bodapati: [optional]

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: Fixed

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: We allocate the size of:
- ds_file_t = return object
- ds_fifo_file_t = the actually object meaningful for FIFO
- path_len = variable length size of file path

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: At file we allocate memory for ds_file_t / ds_fifo_file_t / path_len in a continuous way. Here we are setting fifo_file pointer to the correct memory address offset we allocated for it:

```
(gdb) p file
$27 = (ds_file_t *) 0x55555e5b0980
(gdb) p file+1
$28 = (ds_file_t *) 0x55555e5b0998
(gdb) p 0x55555e5b0998-0x55555e5b0980
$29 = 24
(gdb) p sizeof(ds_file_t)
$30 = 24
```

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: That is filled on the base ds_open:

```
/************************************************************************
Open a datasink file */
ds_file_t *ds_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *stat) {
  ds_file_t *file;

  file = ctxt->datasink->open(ctxt, path, stat);
  if (file != NULL) {
    file->datasink = ctxt->datasink;
  }

  return file;
}
```

` file = ctxt->datasink->open(ctxt, path, stat);` this will call `fifo_open`, in case we return a file that is not null, file will inherit the datasink from the context.

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: this has to respect the base datasink signature:

```
/************************************************************************
Write to a datasink file.
@return 0 on success, 1 on error. */
int ds_write(ds_file_t *file, const void *buf, size_t len) {
  return file->datasink->write(file, buf, len);
}
```

storage/innobase/xtrabackup/src/ds_fifo.cc
altmannmarcelo: same here:

```
/************************************************************************
Close a datasink file.
@return 0 on success, 1, on error. */
int ds_close(ds_file_t *file) { return file->datasink->close(file); }
```

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
altmannmarcelo: Not possible, that is ensured by download_state, we get a file from global list of files and that file remain bind to the thread that first assigned it.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: The problem comes when passing them to thread objects, for example at xbcloud_put

```
bool xbcloud_put(Object_store *store, const std::string &container,
                 const std::string &backup_name) {
  std::atomic<bool> has_errors{false};
. . .
  for (uint i = 0; i < (uint)opt_threads; i++) {
    data_threads[i].num = i;
    data_threads[i].has_errors = has_errors;
    data_threads[i].store = store;
    data_threads[i].container = &container;
    data_threads[i].buf_md5 = new Http_buffer();
    pthread_create_ret =
        pthread_create(&threads[i], nullptr, &put_func, data_threads + i);
    if (pthread_create_ret != 0) {
      msg_ts("%s: failed to create thread %d. Error: %d.\n", my_progname, i,
             pthread_create_ret);
      has_errors.store(true);
      goto cleanup;
    }
  }
  ```

This errors out because `    data_threads[i].has_errors = has_errors;` will try to use copy constructor which (1) is deleted by default (2) is not what we want, we want all threads to be using the same atomic address.

Attribute `data_threads[i].has_errors` will be accessed in the thread context at `put_func` function, and we want them all to share the same atomic address created at `xbcloud_put`, in case one thread has an error, it will immediately be visible to all threads and they will exit, thus I created it as a pointer.

Other option would be to make it global, but I would rather keep it in the xbcloud_put context and make it available via pointer within the upload/download threads.

File: storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc (hunk 32 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc]: removed

File: storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc (hunk 12 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc]: done

File: storage/innobase/xtrabackup/src/ds_xbstream.cc (hunk 30 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/ds_xbstream.cc]: remove as part of pthread refactor

File: storage/innobase/xtrabackup/src/ds_fifo.cc (hunk 28 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/ds_fifo.cc:60]: done

File: storage/innobase/xtrabackup/src/ds_fifo.cc (hunk 29 above)
Comment (altmannmarcelo) [reply storage/innobase/xtrabackup/src/ds_fifo.cc]: done

File: storage/innobase/xtrabackup/src/ds_fifo.cc [hunk 34]
```diff
@@ -35,6 +35,28 @@ typedef struct {
   std::unordered_map<std::string, File> FIFO_list;
   /* Mutex protecting FIFO list */
   std::mutex fifo_mutex;
+
+  /* Add a new pair of fullpath and fd to FIFO_list */
+  void populate_list(std::string fullpath, File fd) {
```
satya-bodapati, line 40: we now have methods, so we can make this as class, make members private and add a destructor that closes fd? Where do we return FIO_list objects back to map?

altmannmarcelo: They are returned at fifo_close:
```
static int fifo_close(ds_file_t *file) {
  . . .
  fifo_context->populate_list(((ds_fifo_file_t *)file)->path, fd);
```

altmannmarcelo: > and add a destructor that closes fd?

This is what fifo_close does. https://github.com/altmannmarcelo/percona-xtrabackup/blob/PXB-2514-8.0/storage/innobase/xtrabackup/src/ds_fifo.cc#L183 
The thing is that all datasink have a specific contract they have to implement which is defined at datasink.cc .

altmannmarcelo: I can convert this to a class but I don't see much benefit from deviant from original datasink standard

satya-bodapati: ok, np  it is not super important anyway.

storage/innobase/xtrabackup/src/ds_fifo.h
```diff
@@ -0,0 +1,75 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+*******************************************************/
+
+/* FIFO Datasink implementation
+
+at xtrabackup_init_datasinks we initialize ds_data as DS_TYPE_FIFO, and create a
```
satya-bodapati: formatting went wrong. Can you please fix it?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -294,6 +306,22 @@ static struct my_option my_long_options[] = {
      &opt_parallel, &opt_parallel, 0, GET_ULONG, REQUIRED_ARG, 1, 1, ULONG_MAX,
      0, 0, 0},
 
+    {"parallel-stream", OPT_THREADS, "Number of parallel stream threads.",
+     &opt_threads, &opt_threads, 0, GET_ULONG, REQUIRED_ARG, 1, 1, ULONG_MAX, 0,
+     0, 0},
```
satya-bodapati: I know that we there is default internal value of parallel value of 5 (1+4?), can we choose a reasonable parallel default when parallel-stream is say like 12?

Or is 5 reasonable enough? :thinking:

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
satya-bodapati: yeah, since it is across threads, it looks global is OK. but it is ok to pass pointers as well

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
satya-bodapati: We also need to document that nmber of streams passed at both ends should be same? Do we also have a test for this?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
```diff
@@ -0,0 +1,186 @@
...
+   * the download) */
+  std::set<std::string> in_progress_files;
+
+  /* list of files allocated for this thread */
+  std::unordered_map<std::string, file_metadata_t> file_list;
```
satya-bodapati, line 36: shall we call this as per_thread_list?

storage/innobase/xtrabackup/src/xbstream_read.cc
```diff
@@ -41,6 +43,26 @@ struct xb_rstream_struct {
...
+  if (fd < 0) return nullptr;
+
+  stream = (xb_rstream_t *)my_malloc(PSI_NOT_INSTRUMENTED, sizeof(xb_rstream_t),
+                                     MYF(MY_FAE));
+  stream->fd = fd;
+  stream->offset = 0;
+
+#ifdef __WIN__
+  setmode(stream->fd, _O_BINARY);
+#endif
```
satya-bodapati, line 61: do we support windows?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1368,6 +1375,23 @@ struct my_option xb_client_options[] = {
...
+     0, GET_INT, REQUIRED_ARG, 1, 1, INT_MAX, 0, 0, 0},
+
+    {"stream-dir", OPT_XTRA_STREAM_DIR,
+     "Directory to write Named Pipe. If ommited, we use --target-dir",
+     &xtrabackup_stream_dir, &xtrabackup_stream_dir, 0, GET_STR_ALLOC,
+     REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
+
+    {"stream-timeout", OPT_XTRA_STREAM_TIMEOUT,
+     "How many seconds to wait for other end to open the stream for reading. "
+     "Default 10 seconds",
```
satya-bodapati: 10 seconds seems a bit less. Can we increase the default to 30 seconds?

storage/innobase/xtrabackup/test/suites/xbcloud/fifo.sh
```diff
@@ -0,0 +1,59 @@
...
+insert_pid=$!
+vlog "Test 1 - Taking inc1 backup"
+take_backup_fifo_xbcloud ${topdir}/stream "--parallel=4 --parallel-stream=3 --incremental-basedir=${full_backup_dir} --extra-lsndir=${inc_backup_dir}" "--parallel=10 --parallel-stream=3 ${inc_backup_name}"
+wait ${insert_pid}
+vlog "Test 1 - Taking inc2 backup"
+take_backup_fifo_xbcloud ${topdir}/stream "--parallel=4 --parallel-stream=3 --incremental-basedir=${inc_backup_dir}" "--parallel=10 --parallel-stream=3 ${inc2_backup_name}"
+record_db_state test
+
+vlog "download & prepare"
+download_fifo_xbcloud ${topdir}/stream "--parallel=10 --parallel-stream=3 ${full_backup_name}" "-x -C ${topdir}/full --parallel-stream=3"
```
satya-bodapati: I want a negative test scenario that uses a different number of fifo-streams on both ends.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1368,6 +1375,23 @@ struct my_option xb_client_options[] = {
      (G_PTR *)&xtrabackup_parallel, (G_PTR *)&xtrabackup_parallel, 0, GET_INT,
      REQUIRED_ARG, 1, 1, INT_MAX, 0, 0, 0},
 
+    {"parallel-stream", OPT_XTRA_PARALLEL_STREAM,
```
satya-bodapati: a suggestion to use : --fifo-streams

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1368,6 +1375,23 @@ struct my_option xb_client_options[] = {
...
+     "Number of FIFO files to use for parallel datafiles stream. Setting this "
```
satya-bodapati, line 1379: If there are more fifo-stream at senders side and parallel is not set by user, choose at least a reasonable parallel stream value?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -294,6 +306,22 @@ static struct my_option my_long_options[] = {
...
+
+    {"stream-dir", OPT_STREAM_DIR,
```
satya-bodapati: should we rename to --fifo-stream-dir

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -960,136 +1079,182 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
...
+      continue;
+    }
 
-    bool empty() {
-      std::lock_guard<std::mutex> g(m);
-      return chunks.empty();
+    if (cntx.file_list->empty() && download_state->file_list_empty()) {
+      break;
     }
+    if (error->load() && !download_state->in_progress_files_empty()) {
```
satya-bodapati: if there is an error and we have some files in progress (to be downloaded),  shouldn't we abort? Can you please add a comment here? (why we continue instead of abort)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -960,136 +1079,182 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
...
-        ++it;
-      }
+    for (uint i = 0; i < opt_threads; i++) {
+      std::string path = "thread_" + std::to_string(i);
+      fn_format(fullpath, path.c_str(), opt_stream_dir, "",
+                MYF(MY_RELATIVE_PATH));
+      mkfifo(fullpath, 0666);
     }
+    msg_ts(
+        "Created %ld Named Pipes(FIFO). Waiting up to 10 seconds for xbstream "
```
satya-bodapati: hardcoded 10 seconds message here. we should the variable here

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -960,136 +1079,182 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
...
+        "to "
+        "open "
+        "the files for reading.\n",
+        opt_threads);
```
satya-bodapati: possible to merge these small strings int one?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,59 +736,82 @@ static bool parse_args(int argc, char **argv) {
   return (0);
 }
 
-bool xbcloud_put(Object_store *store, const std::string &container,
-                 const std::string &backup_name, bool upload_md5) {
-  bool exists;
+typedef struct {
+  uint num;
+  std::atomic<bool> *has_errors;
+  struct file_list_t *file_list;
```
satya-bodapati: would be nice if we can rename to global_file_list (because the file_list at different objects is confusing)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+  Get the next available file to be downloaded from global list. The file
+  will be removed from this list. The caller must allocate the file into
+  its own thread file list.
+
+  @param [in/out]  file  file available to download.
+
+  @return true if a file has been allocated, false otherwise */
+  bool next_file(file_metadata_t &file) {
```
satya-bodapati, line 158: next_file() in two different contexts is confusing. WHen I see "next_file()" I should be able to make out if it is from global/per_thread.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
```diff
@@ -0,0 +1,186 @@
...
+    }
+    return false;
+  }
+
+  /**
+  Adds a new file to global list of files or update the last_chunk if necessary.
+
+  @param [in]  filename   filename of the new file.
+  @param [in]  idx        chunk index of the file. */
+  void add(std::string filename, size_t idx) {
```
satya-bodapati, line 174: add_to_global

satya-bodapati: add_or_update() i somehow need to understand the reference to "global list" in a easy/non confusing way

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -175,14 +202,27 @@ int main(int argc, char **argv) {
...
 
+  if (opt_parallel > 1 && opt_parallel_stream == 1) {
+    msg("%s: option --parallel is deprecated and has no effect. Please use "
+        "--parallel-stream instead.\n",
+        my_progname);
+  }
+
+  if (opt_mode == RUN_MODE_EXTRACT && opt_parallel_stream > 1 &&
+      opt_stream_dir == nullptr) {
+    msg("%s: --parallel-stream requires --stream-dir parameter.\n",
```
satya-bodapati: If we decide to --fifo-streams on xtrabackup side, better to use the same name here.

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -441,36 +479,52 @@ static file_entry_t *file_entry_new(extract_ctxt_t *ctxt, const char *path,
...
   xb_rstream_chunk_t chunk;
   file_entry_t *entry;
   xb_rstream_result_t res;
-
-  extract_ctxt_t *ctxt = (extract_ctxt_t *)arg;
+  if (opt_parallel_stream > 1) {
+    char filename[FN_REFLEN];
+    snprintf(filename, sizeof(filename), "%s%s%lu", opt_stream_dir, "/thread_",
+             (ulong)ctxt.thread_id);
+    stream = xb_stream_read_new(filename, opt_stream_timeout);
```
satya-bodapati: I would instead use different file name than overloading function names.  xb_stream_read_fifo() and xb_stream_read_stdout()?

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -668,46 +722,34 @@ static int mode_extract(int n_threads, int argc __attribute__((unused)),
       ds_set_pipe(ds_decrypt_zstd_ctxt, ds_decompress_zstd_ctxt);
     }
   }
+  data_threads = (extract_ctxt_t *)my_malloc(
+      PSI_NOT_INSTRUMENTED, sizeof(extract_ctxt_t) * (n_threads + 1),
+      MYF(MY_FAE));
+  for (int i = 0; i < n_threads; i++) {
+    data_threads[i].thread_id = i;
```
satya-bodapati, line 734: We are now relaying this thread number to create a stream file. Please comment this somewhere.

satya-bodapati (changes requested):
few more minor comments

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbstream_read.cc
altmannmarcelo: No, this is kept as it was there before this PR

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
altmannmarcelo: Done by renaming the object

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
altmannmarcelo: Done by renaming the object

storage/innobase/xtrabackup/src/xbstream.cc

storage/innobase/xtrabackup/src/xbstream.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: We abort, but we ensure to process all callbacks from the event queue list. Since we have callbacks, we want to ensure that we won't try to access any memory object on the event queue that has been deleted by the main read thread.

Comment added.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

altmannmarcelo: https://jira.percona.com/browse/PXB-3062

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbstream.cc

storage/innobase/xtrabackup/test/suites/xbcloud/fifo.sh
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.h
altmannmarcelo: Done

altmannmarcelo:
Test Cases:

```
$ XBCLOUD_CREDENTIALS="--storage=s3 --s3-access-key=s3user --s3-secret-key=backup123 --s3-endpoint=http://localhost:9000 --s3-bucket=my-bucket" ./run.sh -d /work/ps/ins/8.0/ -r /dev/shm/ -s xbcloud -t t/xbstream_fifo.sh -f
Detecting server version...
Running against Percona Server 8.0.32-24-debug-asan (XtraDB 8.0.32-24)
Using 16 parallel workers
Running suite(s): xbcloud

storage_class                            w6	[skipped]   Storage Class is not supported on MinIO servers
ec2_instance_profile                     w3	[skipped]   This test requires EC2 Instance Profile set
redact_password                          w5	[passed]    58
backup_incomplete                        w1	[passed]    62
basic_operations                         w2	[passed]    82
fifo                                     w4	[passed]    88
xbstream_fifo                            w7	[passed]    89
Spent 382 of 111 seconds executing testcases

SUMMARY: 7 run, 5 successful, 2 skipped, 0 failed

Skipped tests:  storage_class ec2_instance_profile

```

storage/innobase/xtrabackup/src/ds_fifo.h
```diff
@@ -0,0 +1,70 @@
...
+ datasink = 0x55555d670cc0 <datasink_fifo>,
+ root = 0x55555e15ae10 "/tmp/stream",
+ ptr = 0x55555e15ae50,
+ pipe_ctxt = 0x666e632e706d75,
+ fs_support_punch_hole = false
+}
+
+fifo_init - responsible for creating the stream dir and all the fifo files (thread_XX).
+Once created, we open them for writing at the init function despite the fact that we have an open function, this is because with FIFO / STDOUT we do not actually close the FD when a copy thread finishes its works on a specific data file, so ds_open / ds_close operate differently, ds_open just returns a ds_fifo_file_t as file(ds_file_t)->ptr back to xbstream datasink
+xbstream_init will create a list of ds_stream_ctxt_t, those stream contexts will later receive a fifo FD assigned to it. this list is called ds_parallel_stream_ctxt_t.
```
satya-bodapati, line 46: rename to ds_fifo_ctx_t in this entire comment

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,59 +736,85 @@ static bool parse_args(int argc, char **argv) {
   return (0);
 }
 
-bool xbcloud_put(Object_store *store, const std::string &container,
-                 const std::string &backup_name, bool upload_md5) {
-  bool exists;
+typedef struct {
+  /* used to create FIFO file and print thd nr to log */
```
satya-bodapati: thread number

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,59 +736,85 @@ static bool parse_args(int argc, char **argv) {
...
+  uint thread_id;
+  std::atomic<bool> *has_errors;
+  struct global_list_t *global_list;
+  const std::string *container;
+  Object_store *store;
+} download_thread_ctxt_t;
+
+typedef struct {
+  /* used to create FIFO file and print thd nr to log */
```
satya-bodapati: s/nr/number

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -708,59 +736,85 @@ static bool parse_args(int argc, char **argv) {
...
-    return false;
-  }
-
-  struct file_entry_t {
-    my_off_t chunk_idx;
-    my_off_t offset;
-    std::string path;
-  };
+@return string containing the file name */
+std::string build_file_name(const std::string file_name, my_off_t idx) {
```
satya-bodapati: const std::string&

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -960,136 +1082,180 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
...
+  /* Create FIFO files if necessary */
 
-  while (true) {
-    if (chunks.empty() && download_state.empty()) {
-      break;
-    }
-    if (error && !download_state.empty()) {
-      continue;
+  if (opt_threads > 1) {
+    if (my_mkdir(opt_fifo_dir, 0777, MYF(0)) < 0 && my_errno() != EEXIST &&
```
satya-bodapati: 0777 ?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
```diff
@@ -0,0 +1,186 @@
+/******************************************************
+Copyright (c) 2014, 2022 Percona LLC and/or its affiliates.
```
satya-bodapati: we can just write  2023 for new files. Please check other new files

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -897,11 +904,11 @@ struct my_option xb_client_options[] = {
      0},
 
     {"stream", OPT_XTRA_STREAM,
-     "Stream all backup files to the standard output "
-     "in the specified format. Currently supported formats are 'tar' and "
-     "'xbstream'.",
+     "Stream all backup files using xbstream format. Files are streamed to "
```
satya-bodapati, line 907: right, value should be optional since we support only xbstream.
Am talking about REQUIRED_ARG to OPT_ARG

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4416,6 +4454,25 @@ void xtrabackup_backup_func(void) {
     exit(EXIT_FAILURE);
   }
 
+  if (xtrabackup_fifo_streams_set && xtrabackup_fifo_dir != NULL) {
+    xtrabackup_target_dir = xtrabackup_fifo_dir;
```
satya-bodapati, line 4458: what?? we change target_dir to fifo_dir?

satya-bodapati: shouldn't it be opposite? if fifo_streams is set and a directory is not passed, we should use target_dir as fifo_dir?

satya-bodapati (changes requested):
few more changes I noticed (one comment is very important to address)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
altmannmarcelo: Done - 0600

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: It is already OPT_ARG

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: That is what we do. `xtrabackup_fifo_streams_set = --fifo-streams > 1` `xtrabackup_fifo_dir = --fifo-dir missing`

Basically, if you are using `--fifo-streams=2` you can save the FIFO files into `--fifo-dir`, however, for convenience you don't need to remember yet another parameter, you can use `--fifo-streams=2 --target-dir=path` and fifo files will be written on path.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.h
altmannmarcelo: Done

storage/innobase/xtrabackup/src/ds_fifo.h
satya-bodapati: I see that it is used in xbstream.cc, so no need to rename

satya-bodapati (changes requested):
Can you please do before the final approval
1. squash
2. update title to " implement multiple FIFO streams"
3. fix references to new parameter names in commit and github description

satya-bodapati (approved):
LGTM!

altmannmarcelo:
CircleCI failure is due to ds_fifo.h comment on the start of the file. We deliberately did not use clang-format there for better readability

PR #1452, opened 2023-05-04
PXB-3038 - Merge MySQL 8.0.33

altmannmarcelo:
https://jira.percona.com/browse/PXB-3038

Merged 8.0.33 in PXB


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4953,11 +4934,6 @@ static void xtrabackup_stats_func(int argc, char **argv) {
   srv_page_size_shift = 14;
   srv_page_size = (1 << srv_page_size_shift);
 
-  if (srv_n_file_io_threads < 10) {
```
satya-bodapati: How do we enforce the minimum limit?

storage/innobase/srv/srv0start.cc
```diff
@@ -1569,10 +1569,16 @@ static dberr_t recreate_redo_files(lsn_t &flushed_lsn) {
   tables space in case the shutdown wasn't slow. In such case we should start
   from an lsn at least equal to checkpoint_lsn as pages in the tablespace will
   have lsns larger than flushed_lsn. */
+
+#ifdef XTRABACKUP
+  /* PXB recreats redo log file for rollbackup transactionts */
```
satya-bodapati: recreates redo log file for rollback of transactions

storage/innobase/srv/srv0start.cc
```diff
@@ -2221,10 +2221,20 @@ dberr_t srv_start(bool create_new_db IF_XB(, lsn_t to_lsn)) {
 
       ut_ad(buf_pool_pending_io_reads_count() == 0);
 
+#ifdef XTRABACKUP
+      /* we have to recreate always in PXB because the ib_redo0 file
```
satya-bodapati: its an existing mistake, can you add "we have to recreate redo log file always during PXB prepare because the ..."

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: We now rely on exclusively on innobase_read_io_threads & innobase_write_io_threads 

```
srv_n_read_io_threads = (ulint)innobase_read_io_threads;
srv_n_write_io_threads = (ulint)innobase_write_io_threads;
```

storage/innobase/srv/srv0start.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Then I guess, we should change the defaults of these variables to 10? 

   (G_PTR *)&innobase_read_io_threads, (G_PTR *)&innobase_read_io_threads, 0,
     GET_LONG, REQUIRED_ARG, 4, 1, 64, 0, 1, 0},
    {"innodb_write_io_threads", OPT_INNODB_WRITE_IO_THREADS,
     "Number of background write I/O threads in InnoDB.",
     (G_PTR *)&innobase_write_io_threads, (G_PTR *)&innobase_write_io_threads,
     0, GET_LONG, REQUIRED_ARG, 4, 1, 64, 0, 1, 0},

satya-bodapati:
Hi @altmannmarcelo can you please paste Jenkins results and do failure analysis?

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: default is currently 4
The if in question were setting srv_n_read_io_threads and srv_n_write_io_threads to 4 in case srv_n_file_io_threads was less than 10.
srv_n_file_io_threads was a sum of both srv_n_read_io_threads and srv_n_write_io_threads.
To be honest I don't know why this condition was in place, instead of removing, we could do something like:

```
if ( (srv_n_read_io_threads + srv_n_write_io_threads) < 10 ) {
    srv_n_read_io_threads = 4;
    srv_n_write_io_threads = 4;
}
```

what you think?

satya-bodapati: @altmannmarcelo IO I think, we could make the variable minimum values as 4. Then we can get rid of this condition.

altmannmarcelo: Let me test the condition where someone already has this set to 1 for example and we change the default. If we get and error or automatically adjust it to 4. In case we automatically adjust we can change the minimal, otherwise we should keep the condition.

satya-bodapati: Makes sense marce

altmannmarcelo:
```
Breakpoint 1, innodb_init_param () at /work/pxb/src/8.0/storage/innobase/xtrabackup/src/xtrabackup.cc:2211
2211	  srv_n_read_io_threads = (ulint)innobase_read_io_threads;
(gdb) p innobase_read_io_threads
$1 = 4
(gdb) show args
Argument list to give program being debugged when it is started is "--backup --stream=xbstream --innodb-read-io-threads=1 > /dev/null ".
(gdb) 
```
yes, backward compatibility is maintained and we change lower numbers to minimum values.

satya-bodapati: Did you test with explicit value as 1 in .cnf or command line argument?

altmannmarcelo: command line

altmannmarcelo: Argument list to give program being debugged when it is started is "--backup --stream=xbstream --innodb-read-io-threads=1 > /dev/null ".

altmannmarcelo: @satya-bodapati just tested with my.cnf and get the same result. The minimum is adjusted to 4 if I set it to 1.

altmannmarcelo:
Test results https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/316/#showFailuresLink

pxb-1679.sh - https://jira.percona.com/browse/PXB-2948
All others - https://jira.percona.com/browse/PXB-2865

satya-bodapati (approved):
LGTM!

PR #1453, opened 2023-05-12
Fixed PXB-3033 - Execute STOP SLAVE conditionally depending on --lock…

altmannmarcelo:
…-ddl

https://jira.percona.com/browse/PXB-3033

Problem:
After PS-4758 STOP SLAVE is not allowed if the same session has acquired LIFB/LTFB. As part of enabling lock-ddl by default we moved the place where we execute STOP SLAVE to the start of the backup. In some cases, users can ensure DDL do not happen and executing STOP SLAVE later in the backup process can help with minimizing replication lag.

Fix:
Execute STOP SLAVE conditionally depending on --lock-ddl: ON - We execute STOP SLAVE at the start of backup, before LIFB/LTFB. OFF - We execute STOP SLAVE after copying InnoDB tables.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1386,6 +1386,13 @@ static bool backup_rocksdb_checkpoint(Backup_context &context, bool final) {
 @return true if success. */
 bool backup_start(Backup_context &context) {
   if (!opt_no_lock) {
+    /* STOP SLAVE if lock-ddl=OFF */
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/320/#showFailuresLink

binlog_102 - false positive (this does not execute --safe-slave-backup)
pxb-1679.sh - https://jira.percona.com/browse/PXB-2948
All others - https://jira.percona.com/browse/PXB-2865

satya-bodapati (approved):

PR #1454, opened 2023-05-16
PXB-2955 : Implement dictionary cache

satya-bodapati:
Problem(s)
----------
1. xtrabackup --prepare requires very high ram (--use-memory doesn't help)
2. xtrabackup --prepare crashes or doesn't finish if backup dir has more than 300K tablespaces
3. xtrabackup --prepare takes long time to complete if backup dir has huge number of tablespaces (redo to be applied is small)

Before this PR(the old design)
------------------------------
xtrabackup --prepare, after applying redo logs, starts InnoDB engine to do "rollback" of uncompleted
transactions at the time of backup.

For this purpose, it has to prepare a dictionary. Rollback works by loading a table object. This is based
on "table_id" recorded in undo log record.

PXB doesn't use the same server APIs to boot DD engine. It is complicated and requires significant efforts
to load server component and initialize DD cache. PXB uses SDI (Serialized Dictionary Info) found on IBD tablespaces.

SDI is a JSON representation of the "table structure". Index information, index fields, columns, column types etc
are part of SDI.

it is possible for an user to see SDI using the below command

ibd2sdi t1.ibd > t1.sdi

PXB uses the same SDI present inside the tablespace. To build a dictionary cache (that is required for rollback). PXB
goes through all tablespaces (*.ibd files) in backup dir. For each IBD:

1. open IBD file
2. Parse SDI using SDI APIs
3. Uncompress the SDI (it is stored as zlib compressed stream)
4. Deserialize the SDI to Server table object (dd::Table)
5. Convert dd::Table to InnoDB table object (dict_table_t)

Note that we load every table in the above steps as "non-evictable". This is because InnoDB has master thread enforces
the table (dict_table_t) cache limit.

The reason for storing them as "non-evictable" is that if a table is evicted, and rollback asks to load a table using
"table_id", we don't know the "tablespace" that contains the table_id. The relation ship "table_id" to "space_id" doesnt
exist. Hence, we load the tables as "non-evictable"

This design (to not being able to evict tables) has a huge penalty. Since we have to load all tables from all IBDs,
the number of tables loaded into cache is huge. We load a table that is NOT required for rollback.

Fix (New Design):
-----------------

If we could somehow maintain the relation "table_id (key) -> space_id (value)", we can do "on-demand" loading of table

The "somehow" part:
-------------------
1. We scan mysql.indexes table to parse a DD table directly using InnoDB APIs
2. We parse the InnoDB record and extract "se_private_data"
3. A sample "se_private_data" looks like this: "id=158;root=4;space_id=6;table_id=1067;trx_id=1548;"
4. From se_private_data, extract "space_id" and "table_id". From this we know for table_id 1067, the space_id
   to look for SDi is 6.
8. For partitions we take a similar approach. we scan mysql.index_partitions and the build "table_id -> space_id" map

New Flow:
---------
1. Load all tables mysql.ibd, we need them to be able to do B-Tree scan of mysql.indexes, mysql.index_partitions, mysql.tables tables.
2. Read from mysql.indexes/index_partitions to build dictionary map (std::unordered_map) of  "table_id -> space_id"
3. On rollback, if server looks for a table_id, we first look into cache. If it doesn't exists, we load the table using the following steps.
   a. find the space_id for the given table_id
   b. Load a table_id from the above space_id (step a result)
   c. Load the table into cache
   d. Table is opened (table->n_ref_count is 1)

4. After the table object is used and undo apply is over, we close the table (table->n_ref_count is 0). Hence evictable

We save memory by
1. By not loading all tables from all tablespaces (using SDI) upfront
2. We load a table on-demand. i.e. "only" if there is rollback on the table (or on --export)

The partition conundrum:
------------------------
The above solution to use "table_id -> space_id" doesn't work with partitions. This is because Server stores SDI only in the first IBD.

Lets say, we create a partition table with 3 partitions. t1#p#p0.ibd, t1#p#p1.ibd and t1#p#p2.ibd. Server stores SDI only in t1#p#p0.ibd.
Other partitions t1#p#p1.ibd and t1#p#p2.ibd do not have the dd::Table SDI. Only dd::Tablespace SDI is present in p1 and p2.

This means, if ever there is rollback on p1 and p2, we cannot use the p1.ibd or p2.ibd. We have to look always into p0.ibd (or the first partition)

To solve this problem, we first have to identify all partition of a table. In the above example, we have to find the group is

t1#p#p0.ibd, t1#p#p1.ibd and t1#p#p2.ibd all three form one table with paritions. We could do a name scan but we chose a different way to figure out
the partitions. Again, that is by relying "partition index id"

Lets get into details. How do we find "all partitions" of table?
Answer: Scan mysql.index_partitions, we can get index_partition id and the
InnoDB space_id and also the table_id->space_id relation

table_id_space_map:
{table_id -> space_id}
1075 -> 13 (p0.ibd)
1076 -> 14 (p1.ibd)
1077 -> 15 (p2.ibd)}

part_id_spaces_map:
{partition_id -> space_id}
{296 -> 13
296 -> 14
296 -> 15 }

space_part_map
{space_id -> partition_id } {
13 : 296
14 : 296
15 : 296 }

Lets say, we are asked to load table for table_id 1077.
We look into table_id_space_map and figure out that the space_id
for 1077 is 15 (p2.ibd)

Next, check if this space belongs to a "partitioned table"
For this, we look into space_part_map. If an entry exists, it means
the space_id belongs to partitioned table.

We get the partition id. For space_id 15, the partition id is 296.

Given a partition_id 296, we can easily know all the partition IBs by using
part_id_spaces_map

For 296, we now know that the IBDs are 13, 14, 15.

We will start from lower bound(13) and start looking into all space_ids in this
group. (14, 15)

We will load SDI from space_id 13 (p0.ibd)

Other improvements:
-------------------
We now use multiple threads to load IBDs into cache. Previously, only the scan is parallel.
See Tablespace_files::open_ibds()

### DIFF
--- a/storage/innobase/dict/dict0dd.cc
+++ b/storage/innobase/dict/dict0dd.cc
@@ -73,8 +73,7 @@ Data dictionary interface */
 #include "query_options.h"
 #include "sql/create_field.h"
 #include "sql/mysqld.h"  // lower_case_file_system
-#include "sql_base.h"
-#include "sql_table.h"
+#include "storage/innobase/xtrabackup/src/xb_dict.h"
 #endif /* !UNIV_HOTBACKUP */
 
 const char *DD_instant_col_val_coder::encode(const byte *stream, size_t in_len,
@@ -1123,7 +1122,9 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
 
     table->version = version;
     dict_table_autoinc_lock(table);
-    dict_table_autoinc_initialize(table, autoinc + 1);
+    /* Table autoinc is later incremented (in this function) to next offset (+1)
+    after load from dynamic metadata */
+    dict_table_autoinc_initialize(table, autoinc);
     dict_table_autoinc_unlock(table);
     table->autoinc_persisted = autoinc;
   }
@@ -1189,8 +1190,20 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
 
   mutex_enter(&dict_sys->mutex);
 
-  dict_table_add_to_cache(table, false);
+  dict_table_add_to_cache(table, true);
   table->acquire();
+
+  /* Load autoinc from dynamic meta data */
+  if (dict_sys->dynamic_metadata != nullptr) {
+    dict_table_load_dynamic_metadata(table);
+  }
+
+  /* We do not do the table open via ha_innobase::open().
+  innobase_initialize_autoinc() will NOT be called and the
+  counter is NOT advanced. Hence, advance to next value here */
+  if (table->autoinc != ~0ULL) {
+    ++table->autoinc;
+  }
 
   mutex_exit(&dict_sys->mutex);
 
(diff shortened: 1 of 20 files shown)


satya-bodapati:
old dictionary PR that was closed automatically on 'sync fork'. #1435

altmannmarcelo:
@satya-bodapati  can you paste jenkins link ?

satya-bodapati:
@altmannmarcelo https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/323/#showFailuresLink

https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/284/

altmannmarcelo (approved):
LGTM - squash and merge

PR #1455, opened 2023-05-20
Set CRC as static explicitly

fundawang:
CRC is acturally a STATIC library, which is used to merge into other executables. Seting it as STATIC makes it more friendly with cmake options.


altmannmarcelo:
Hi @fundawang .

Thanks for your contribution. Can you clarify if you saw any warning that led you to work on this fix, or is it more to have better readability of the code and spot right away that we link it statically?

fundawang:
> Thanks for your contribution. Can you clarify if you saw any warning that led you to work on this fix, or is it more to have better readability of the code and spot right away that we link it statically?

Under most of distros, default cmake command produces shared libraries for `ADD_LIBRARY` (libcrc.so), which requires the library be installed into DESTDIR.

CRC is meant to be linked into other binaries, I don't think it worth to be built as a shared library.

PR #1465, opened 2023-06-04
PXB-3059 add a innodb_redo_log archive option to xtrabackup

wangbincmss:
Backup utilities that copy redo log records may sometimes fail to keep pace with redo log generation while a backup operation is in progress, resulting in lost redo log records due to those records being overwritten.
If the MySQL server innodb_redo_log_archive_dirs =NULL and the redo log file storage media operates at a faster speed than the backup storage media,then result in backup fail.
Add a option --redo_log_arch_dir=tmp_redo_dir:redo_dir can solve this backup fail,
If innodb_redo_log_archive_dirs =NULL ,then after backup innodb_redo_log_archive_dirs is always NULL not change user setting.

### DIFF
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -932,6 +932,7 @@ void Archived_Redo_Log_Monitor::thread_func() {
 
   stopped = false;
   ready = false;
+  redo_arch = false;
 
   auto mysql = xb_mysql_connect();
   if (mysql == nullptr) {
@@ -953,6 +954,20 @@ void Archived_Redo_Log_Monitor::thread_func() {
     std::string dir;
     std::string subdir;
   } archive;
+  std::ostringstream query;
+  read_mysql_variables(mysql, "SHOW VARIABLES", vars, true);
+
+  if (!(redo_log_archive_dirs == nullptr || *redo_log_archive_dirs == 0)) {
+    xb::info() << "innodb_redo_log_archive_dirs is now  set .";
+    redo_arch = true;
+  }
+
+  if (xtrabackup_redo_log_arch_dir) {
+    xb::info() << "xtrabackup redo_log_arch_dir is now  set .";
+    query << "SET GLOBAL innodb_redo_log_archive_dirs ="
+          << "\"" << xtrabackup_redo_log_arch_dir << "\"";
+    xb_mysql_query(mysql, query.str().c_str(), false);
+  }
 
   read_mysql_variables(mysql, "SHOW VARIABLES", vars, true);
 
@@ -1128,6 +1143,11 @@ void Archived_Redo_Log_Monitor::thread_func() {
   }
 
   xb_mysql_query(mysql, "SELECT innodb_redo_log_archive_stop()", false);
+  if (!redo_arch) {
+    xb::info() << "innodb_redo_log_archive_dirs is not set,set it NULL .";
+    xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",
+                   false, true);
+  }
   unlink(archive.filename.c_str());
   rmdir(archive.dir.c_str());
   mysql_close(mysql);
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -1030,7 +1030,8 @@ void Archived_Redo_Log_Monitor::thread_func() {
   if (res == nullptr) {
     xb::info() << "Redo Log Archiving is not used.";
     if (!innodb_redo_log_archive) {
-	xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",false, true);	    
+      xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",
+                     false, true);
     }
     mysql_close(mysql);
     my_thread_end();
@@ -1062,7 +1063,8 @@ void Archived_Redo_Log_Monitor::thread_func() {
     if (file < 0) {
       xb::error() << "cannot open " << SQUOTE(archive.filename.c_str());
       if (!innodb_redo_log_archive) {
-	 xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",false, true);
+        xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",
+                       false, true);
       }
       mysql_close(mysql);
       my_thread_end();
@@ -1100,9 +1102,11 @@ void Archived_Redo_Log_Monitor::thread_func() {
         if (n_read == MY_FILE_ERROR) {
           xb::error() << "cannot read from " << archive.filename.c_str();
           if (!innodb_redo_log_archive) {
-	      xb_mysql_query(mysql, "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",false, true);
-	  }
-	  mysql_close(mysql);
+            xb_mysql_query(mysql,
+                           "SET GLOBAL innodb_redo_log_archive_dirs = NULL;",
+                           false, true);
+          }
+          mysql_close(mysql);
           my_thread_end();
           return;
         }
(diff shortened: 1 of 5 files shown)


wangbincmss:
Add a option --redo_log_arch_dir

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -953,6 +954,20 @@ void Archived_Redo_Log_Monitor::thread_func() {
     std::string dir;
     std::string subdir;
   } archive;
+  std::ostringstream query;
```
altmannmarcelo: move this declaration close to utilization

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -953,6 +954,20 @@ void Archived_Redo_Log_Monitor::thread_func() {
...
+  read_mysql_variables(mysql, "SHOW VARIABLES", vars, true);
```
altmannmarcelo: is this necessary, having in mind we do the exact same query a few lines below ?

storage/innobase/xtrabackup/src/redo_log.h
```diff
@@ -271,6 +271,9 @@ class Archived_Redo_Log_Monitor {
   /** readiness flag. */
   std::atomic<bool> ready;
 
+  /** redo arch flag. */
```
altmannmarcelo: Elabore a bit more in the comment, also rename the variable name to something more useful. IIUC this flag controls if xtrabackup has set redo log arch

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -953,6 +954,20 @@ void Archived_Redo_Log_Monitor::thread_func() {
...
+
+  if (!(redo_log_archive_dirs == nullptr || *redo_log_archive_dirs == 0)) {
+    xb::info() << "innodb_redo_log_archive_dirs is now  set .";
+    redo_arch = true;
+  }
+
+  if (xtrabackup_redo_log_arch_dir) {
```
altmannmarcelo: Here we are setting redo archive unconditionally, what happens if someone already have it set? we will overwrite to --redo-log-arch-dir and later return it back to NULL. We should only set if it is not already set.

altmannmarcelo (changes requested):
Hi @wangbincmss .

Thanks a lot for your contribution. I have left a few comments to improve your PR. Apart from that I would like to ask you to enhance in two areas:
1) We try to avoid making xtrabackup change server configuration. When we do so, we need to ensure that in case of error we return the configuration back to the original value. Can you please add some error handling that if xtrabackup fails after we setting redo log archive we restore the configuration to its original value?

2) We have a test suite that ensures we are not introducing regression on other parts of the code, can you please create a test case to cover the funtionality we are adding here ? You can check https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/test/t/xb_log_archiving.sh and https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/test/inc/xb_log_archiving.sh

altmannmarcelo (changes requested):
@wangbincmss I think you re-requested the review by mistake, please have a look on above comments.

storage/innobase/xtrabackup/src/redo_log.cc
wangbincmss: Yes，need to check   SET GLOBAL innodb_redo_log_archive_dirs  takes effect or not

wangbincmss:
@altmannmarcelo hi ,    I make some changes
1: add some error handling that if xtrabackup fails,  the server option changes only when  redo log archive is null , so error handling : changing redo log archive  to original value (null)  if xtrabackup fails
2:  avoid making xtrabackup change server configuration.: the server option changes only when  redo log archive is null, and bakup finished then change  to original value (null)
3:     rename the variable  to innodb_redo_log_archive   and   add a bit   comments
4: add four test cases    covers   server  and xtrabackup redo log  redo  option is NULL and true value

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -953,17 +954,31 @@ void Archived_Redo_Log_Monitor::thread_func() {
...
     std::string subdir;
   } archive;
-
+  
   read_mysql_variables(mysql, "SHOW VARIABLES", vars, true);
 
-  if (redo_log_archive_dirs == nullptr || *redo_log_archive_dirs == 0) {
+  if (!(redo_log_archive_dirs == nullptr || *redo_log_archive_dirs == 0)) {
+    xb::info() << "innodb_redo_log_archive_dirs is already  set .";
+    innodb_redo_log_archive = true;
```
altmannmarcelo: what the code is doing and what the description of the variable says is contradicting.
You set this to true if the redo archive is already set on the server. Based on your comment this should be true when PXB changed redo archive dir from null to xtrabackup_redo_log_arch_dir.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1014,6 +1029,10 @@ void Archived_Redo_Log_Monitor::thread_func() {
 
   if (res == nullptr) {
     xb::info() << "Redo Log Archiving is not used.";
+    if (!innodb_redo_log_archive) {
```
altmannmarcelo: Since we are repeating the same code in multiple places, we might create a member function to execute the cleanup.

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1128,6 +1161,11 @@ void Archived_Redo_Log_Monitor::thread_func() {
   }
 
   xb_mysql_query(mysql, "SELECT innodb_redo_log_archive_stop()", false);
+  if (!innodb_redo_log_archive) {
+    xb::info() << "innodb_redo_log_archive_dirs is not set,set it NULL .";
```
altmannmarcelo: This message is measleading, if it is not set already (I read this as NULL) why we set it to NULL again ?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -849,6 +851,11 @@ struct my_option xb_client_options[] = {
      (G_PTR *)&xtrabackup_incremental_basedir,
      (G_PTR *)&xtrabackup_incremental_basedir, 0, GET_STR, REQUIRED_ARG, 0, 0,
      0, 0, 0, 0},
+    {"redo_log_arch_dir", OPT_INNODB_REDO_LOG_ARCHIVE_DIRS,
+     "redo log archive destination directory",
+     (G_PTR *)&xtrabackup_redo_log_arch_dir,
+     (G_PTR *)&xtrabackup_redo_log_arch_dir, 0, GET_STR, OPT_ARG, 0, 0, 0, 0, 0,
```
altmannmarcelo:
```suggestion
     (G_PTR *)&xtrabackup_redo_log_arch_dir, 0, GET_STR, REQUIRED_ARG, 0, 0, 0, 0, 0,
```

altmannmarcelo: This is not a flag, we need to require argument

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: I would rather re-work this whole block. Its a bit counter intuitive. Some thing like below will be better:

```
if (redo_log_archive_dirs == nullptr || *redo_log_archive_dirs == 0) {
  std::ostringstream query;
  if (xtrabackup_redo_log_arch_dir) {
    query << "SET GLOBAL innodb_redo_log_archive_dirs ="
          << "\"" << xtrabackup_redo_log_arch_dir << "\"";
    xb_mysql_query(mysql, query.str().c_str(), false, true); <-- we need to keep this to fail in case of error
    innodb_redo_log_archive = true;
  } else {
    xb::info() << "Redo Log Archiving is not set up.";
    free_mysql_variables(vars);
    mysql_close(mysql);
    my_thread_end();
    return;
  }
}
xb::info() << "xtrabackup redo_log_arch_dir is now  set .";
```

altmannmarcelo (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
wangbincmss: feels a little measleading ，the true meaning is  the server innodb_redo_log_archive_dirs is not set(NULL) and PXB innodb_redo_log_archive_dirs is not NULL .  in order to not change user option, so change  it to NULL at last.
 I will  correct  the mistake.

storage/innobase/xtrabackup/src/redo_log.cc
wangbincmss: I will create it

PR #1467, opened 2023-06-20
PXB-2514-8.0 - Postfix Fifo Stream

altmannmarcelo:
- Adjusted xtrabackup cleanup
 - Added MacOS support for kqueue (epoll equivalent).
 - Fixed missing mutex initialization.
 - Added Backup --parallel on xbstream. Which triggered below problem:

1 Write side Open operation happens on a single thread (xtrabackup)
2 for each fifo-streams; do
 2.1 open it on non blocking mode
 2.2 if the other side is open for read - set it back to blocking mode
 2.3 loop again to open next fifo stream
3 Read side open operation happens on a single thread (xbstream/xbcloud)
4 similar to write, for each fifo-streams; do
  4.1 open it on non blocking mode - this always succeed
  4.2 then we add the FD to epoll and wait for EPOLLIN. Once that event
is triggered we set it back to blocking mode.
  4.3 loop again to open next fifo stream

The problem is EPOLLIN event is only set when there is data to be read
in the FD. Essentially, this will happen when Write side opens all fifo
files and progress to writing data (redo follow thread and copy threads)
However, this will never happen because now read side is on a single
thread (while before we opened all fifo files) in parallel, now we open
one after the other, meaning the first fifo file will be blocked on 4.2
waiting for data to be available on the fifo file, and write will be
waiting on 2.2 for the second fifo file, because the read side has not
open the second file for read and is waiting data to be available on the
pipe to trigger EPOLLIN event.

Fix:
Added a controll character that the write side will send on the fifo
file triggering EPOLLIN event on the read side. This will allow the read
to open the next fifo file and so on. On the read side, we will read
this 1 byte character to reset the fifo (consume the available) data
and ensure we have a clear buffer for real data.

### DIFF
--- a/storage/innobase/xtrabackup/src/file_utils.cc
+++ b/storage/innobase/xtrabackup/src/file_utils.cc
@@ -20,9 +20,14 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 
 #include "file_utils.h"
 #include <mysql/service_mysql_alloc.h>
+#ifdef __APPLE__
+#include <sys/event.h>
+#else
 #include <sys/epoll.h>
+#endif
 #include <thread>
 #include "common.h"
+#include "ds_fifo.h"
 #include "msg.h"
 #include "my_dir.h"
 #include "my_io.h"
@@ -358,6 +363,15 @@ File open_fifo_for_write_with_timeout(const char *path, uint timeout) {
     return -1;
   }
 
+  /*
+   * Signals the read side by writing DS_FIFO_CONTROL_CHAR
+   * This triggers EVFILT_READ/EPOLLIN. This is necessary because
+   * the read side is also opened in non-blocking mode. We change
+   * back to blocking node once kqueue/epoll is notified. The way
+   * to notify that data is available to read is to write some data
+   * to it.
+   */
+  my_write(fd, (uchar *)DS_FIFO_CONTROL_CHAR, 1, MYF(0));
   return fd;
 }
 
@@ -365,7 +379,7 @@ File open_fifo_for_read_with_timeout(const char *path, uint timeout) {
   int fd;
   uint attempt = 0;
   do {
-    fd = open(path, O_RDONLY | O_NONBLOCK);
+    fd = my_open(path, O_RDONLY | O_NONBLOCK, MYF(0));
     if (fd < 0) {
       attempt++;
       std::this_thread::sleep_for(std::chrono::seconds(1));
@@ -377,6 +391,22 @@ File open_fifo_for_read_with_timeout(const char *path, uint timeout) {
   }
 
   /* File was open, lets check its open on the other side */
+#ifdef __APPLE__
+  struct timespec tm = {timeout, 0};
+  int kqueue_fd = kqueue();
+  if (kqueue_fd < 0) {
+    return -1;
+  }
+  struct kevent ev;
+  EV_SET(&ev, fd, EVFILT_READ, EV_ADD, 0, 0, NULL);
+  if (kevent(kqueue_fd, &ev, 1, NULL, 0, NULL) < 0) {
+    return -1;
+  }
+  if (kevent(kqueue_fd, NULL, 0, &ev, 1, &tm) <= 0) {
+    return -1;
+  }
+  close(kqueue_fd);
+#else
   int epoll_fd = epoll_create1(0);
   if (epoll_fd < 0) {
     return -1;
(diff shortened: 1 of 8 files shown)


altmannmarcelo:
Tests
___
```
 ╰▶ $ XBCLOUD_CREDENTIALS="--storage=s3 --s3-access-key=s3user --s3-secret-key=backup123 --s3-endpoint=http://localhost:9000 --s3-bucket=my-bucket" ./run.sh -d /work/ps/ins/8.0/ -r /dev/shm/ -t t/xbstream_fifo.sh -s xbcloud
Detecting server version...
This is ASAN build
Running against Percona Server 8.0.32-24-debug-asan (XtraDB 8.0.32-24)
Using 16 parallel workers
Running suite(s): xbcloud

storage_class                            w7	[skipped]   Storage Class is not supported on MinIO servers
ec2_instance_profile                     w4	[skipped]   This test requires EC2 Instance Profile set
redact_password                          w6	[passed]    59
backup_incomplete                        w2	[passed]    60
basic_operations                         w3	[passed]    85
fifo                                     w5	[passed]    89
xbstream_fifo                            w1	[passed]    97
Spent 394 of 118 seconds executing testcases

SUMMARY: 7 run, 5 successful, 2 skipped, 0 failed

Skipped tests:  storage_class ec2_instance_profile

```

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -376,23 +376,5 @@ File open_fifo_for_read_with_timeout(const char *path, uint timeout) {
     return -1;
   }
 
-  /* File was open, lets check its open on the other side */
```
satya-bodapati, line 379: where is this handled now? in my_open()?

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -120,10 +121,8 @@ static struct my_option my_long_options[] = {
...
      GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
     {"parallel", OPT_PARALLEL,
-     "Deprecated. Use --fifo-streams instead. Number of worker threads for "
-     "reading / writing.",
-     &opt_parallel, &opt_parallel, 0, GET_INT, REQUIRED_ARG, 1, 1, INT_MAX, 0,
-     0, 0},
+     "Number of worker threads for reading / writing.", &opt_parallel,
+     &opt_parallel, 0, GET_INT, REQUIRED_ARG, 1, 1, INT_MAX, 0, 0, 0},
     {"fifo-streams", 'f', "Number of worker threads for reading / writing.",
      &opt_fifo_streams, &opt_fifo_streams, 0, GET_INT, REQUIRED_ARG, 1, 1,
```
satya-bodapati: comment is same for both. Please fix

storage/innobase/xtrabackup/src/xbstream.cc
```diff
@@ -202,16 +202,20 @@ int main(int argc, char **argv) {
...
+    if (opt_fifo_dir == nullptr) {
+      msg("%s: --fifo-streams requires --fifo-dir parameter.\n", my_progname);
+      goto err;
+    }
+    /* adjust n_threads to at least fifo streams */
+    if (opt_parallel < opt_fifo_streams) {
+      msg_ts(
+          "%s: Adjusting number of threads from %d to %d to match number of "
+          "fifo "
+          "streams.\n",
```
satya-bodapati: can be combined?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/xbstream.cc

storage/innobase/xtrabackup/src/xbstream.cc

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: Fixed. Added support for MacOS Kqueue (mac's replacement of epoll)

storage/innobase/xtrabackup/src/file_utils.cc
```diff
@@ -376,5 +383,49 @@ File open_fifo_for_read_with_timeout(const char *path, uint timeout) {
...
+
+  if (fcntl(fd, F_SETFL, O_RDONLY) == -1) {
+    return -1;
+  }
+  char buf[1];
+  if (my_read(fd, (uchar *)buf, 1, MYF(0)) < 0) {
+    return -1;
+  }
+  if (memcmp(buf, DS_FIFO_CONTROL_CHAR, 1)) {
+    return -1;
```
satya-bodapati: can you please explain the synchronization part here as a code comment

satya-bodapati (commented):
looks good to me

storage/innobase/xtrabackup/src/file_utils.cc
altmannmarcelo: Done. I added a more extensive comment on commit too.

satya-bodapati (approved):

PR #1468, opened 2023-06-20
Fixed PXB-2811 - Fix compilation warnings 

altmannmarcelo:
-- CMakeLists Warning
CMake Warning at cmake/compile_flags.cmake:36 (MESSAGE):
  -DMYSQL_CLIENT should be in COMPILE_DEFINITIONS not COMPILE_FLAGS
Call Stack (most recent call first):
  storage/innobase/xtrabackup/src/CMakeLists.txt:73 (ADD_COMPILE_FLAGS)

Modified ADD_COMPILER_FLAGS to use ADD_COMPILE_DEFINITIONS (https://github.com/percona/percona-xtrabackup/commit/6198ce663e0aec822523f2bc2e46bfe5291888b3 / https://github.com/percona/percona-xtrabackup/commit/898f86afc0d67baf2370249215683ed599d74cbb)

-- libkmip Warning

Adjusted submodule pointer to pick up d87d916

-- Truncate string by strncopy / snprintf
Adjuster functions to safely consider the null terminator / avoid
buffer overflows.

-- sdi.cc

warning: unused variable 'mysqld_version_id' [-Wunused-variable]

left only the assert to be compiled when xtrabackup is defined

Fixed Issue with procps version 4.
Now we detect the version during cmake and use the correct library
when linking xbtrabackup

This also fixes PXB-3066 - Compilation issues on Debian 12

### DIFF
--- a/sql/dd/impl/sdi.cc
+++ b/sql/dd/impl/sdi.cc
@@ -375,24 +375,27 @@ Sdi_type serialize(const Tablespace &tablespace) {
    compatibility check: MYSQL_VERSION less than or equal, dd_version
    equal, and sdi_version equal.
 */
-bool CheckDefaultCompatibility(const RJ_Document &doc) {
+bool CheckDefaultCompatibility(const RJ_Document &doc __attribute__((unused))) {
   assert(doc.HasMember("mysqld_version_id"));
-
+#if defined(XTRABACKUP)
+  assert(doc["mysqld_version_id"].IsUint64());
+  assert(doc.HasMember("dd_version"));
+  assert(doc["dd_version"].IsUint());
+  assert(doc.HasMember("sdi_version"));
+  assert(doc["sdi_version"].IsUint64());
+#else
   const RJ_Value &mysqld_version_id = doc["mysqld_version_id"];
   assert(mysqld_version_id.IsUint64());
-#if !defined(XTRABACKUP)
   if (mysqld_version_id.GetUint64() > std::uint64_t(MYSQL_VERSION_ID)) {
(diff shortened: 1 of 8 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/328/

All failures are known (double --prepare issue or pxb-1679.sh test issue)

sql/dd/impl/sdi.cc
```diff
@@ -375,37 +375,38 @@ Sdi_type serialize(const Tablespace &tablespace) {
    compatibility check: MYSQL_VERSION less than or equal, dd_version
    equal, and sdi_version equal.
 */
-bool CheckDefaultCompatibility(const RJ_Document &doc) {
+bool CheckDefaultCompatibility(const RJ_Document &doc __attribute__((unused))) {
   assert(doc.HasMember("mysqld_version_id"));
```
satya-bodapati, line 379: warning: unused variable 'mysqld_version_id' [-Wunused-variable]
but this assert is using it? :thinking:

satya-bodapati (approved):

satya-bodapati (approved):

sql/dd/impl/sdi.cc
altmannmarcelo: yes, I think the assertion goes to a macro and the compiler does not recognize it

satya-bodapati: I see

chenrui333:
FYI, this PR actually breaks the build on macOS, https://github.com/Homebrew/homebrew-core/pull/137048

PR #1472, opened 2023-07-07
add loongarch support

qiangxuhui:
Add support for double-conversion according to [google-conversion:utils.h](https://github.com/google/double-conversion/blob/master/double-conversion/utils.h#L142)


altmannmarcelo:
Hi @qiangxuhui .

First, thanks for your PR and suggestions.
However, we do not make custom modifications to third-party tools; we rely on merging them from upstream (oracle MySQL).
I suggest you propose the same PR there, and we will get it into xtrabackup when they release it.

Thanks.

PR #1473, opened 2023-07-10
PXB-3075 Xtrabackup 8.0.33 cannot prepare backups from 8.0.29 and previous versions

satya-bodapati:
Revert "Bug#34881268 : Recovery should abort if logged index version isn't matching"
    
    This reverts commit 39702c432e60639deefd269a6431fc10c2951bcb.
    
    Unlike a server which doesn't accept recovery from older versions of
    redo log, xtrabackup should be able to prepare older backup using newer
    xtrabackup version

    Add testcase. We now verify we can prepare a backup dir taken on 8.0.29.
    The backup dir has redo log with instant columns information.


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/287/

PR #1474, opened 2023-07-18
PXB-3089 - Merge 8.0.34 - Q3 2023

altmannmarcelo:
https://jira.percona.com/browse/PXB-3089


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/335/#showFailuresLink

storage/innobase/log/log0recv.cc
```diff
@@ -4120,18 +4105,10 @@ Parses and hashes the log records if new data found.
                                         an mtr which we can ignore, as it is
                                         already applied to tablespace files)
                                         until which all redo log has been
-<<<<<<< HEAD
                                         scanned
 @param[in,out]  to_lsn                  LSN to stop recovery at */
-static dberr_t recv_recovery_begin(log_t &log, const lsn_t checkpoint_lsn,
+static void recv_recovery_begin(log_t &log, const lsn_t checkpoint_lsn,
```
satya-bodapati, line 4110: one doubt, why do we change to void? is it because we ignore the return value of this function in caller anyway?

altmannmarcelo: correct, check https://github.com/altmannmarcelo/percona-xtrabackup/commit/5d2f068908c53676bb2319b11d5378c85ae08735#diff-26f5d4b9822d8a85edd9ea715cde19e90347775be561b8919601cfcb3df8cb49L3864

satya-bodapati: Hmm, but the immediate caller is using the returned value.
``` 
 if (err != DB_SUCCESS) {
    return err;
  }
```

satya-bodapati: ah, this is done by upstream?

satya-bodapati: Indeed it is by upstream

satya-bodapati (approved):
LGTM!

PR #1476, opened 2023-07-25
Implement PXB-3091 - Remove Server Version Checks

altmannmarcelo:
https://jira.percona.com/browse/PXB-3091

Removed server version check. Going forward, PXB 8.0.X will support all versions from the same minor version (8.0).
Backing up any major versions will be blocked.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-multijob/290/

All test failures are known.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1128,7 +1128,7 @@ struct my_option xb_client_options[] = {
 
     {"no-server-version-check", OPT_NO_SERVER_VERSION_CHECK,
      "This option allows backup to proceed when the server version is greater "
-     "(newer) than the PXB supported version",
+     "(newer) than the PXB supported version. (This option is decprecated).",
```
satya-bodapati: deprecated

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -370,7 +368,7 @@ static bool check_server_version(unsigned long version_number,
   }
 
   version_supported =
-      version_supported || (version_number > 80000 && version_number < 90000);
+      version_supported || (version_number > 80000 && version_number <= 80099);
```
satya-bodapati, line 371: what about blocking higher versions like 8.1+ and 9.X and 10.X and so on? Separate PR?

altmannmarcelo: This is what this check does

80099 = [8].[00].[99] - Essentially we allow up to version 8.0.99.
8.1.0 is 80100

satya-bodapati: Ah, right. We don't need another major version check.

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: fixed and force-pushed

satya-bodapati (approved):

PR #1478, opened 2023-07-31
Implemented PXB-3059 - Add an option to set innodb_redo_log archive.

altmannmarcelo:
https://jira.percona.com/browse/PXB-3059

Added an option to set innodb_redo_log_archive_dirs. xtrabackup only sets the option if the server innodb_redo_log_archive_dirs is currently set to NULL. We always restore it back to NULL in the end of the backup.

Thanks wangbincmss for the patch.

Closes #1465


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/342/#showFailuresLink

PR #1481, opened 2023-08-07
Fixed PXB-3029 - Xtrabackup keeps locking table even with the use of …

altmannmarcelo:
…--tables-exclude and –lock-ddl-per-table.

https://jira.percona.com/browse/PXB-2860

Problem:
When running lock-ddl-per-table, xtrabackup places an MDL on all known tables at the start of the backup.  This is to prevent DDL from happening until we close the connection to the server.  However, if user has opted for a selective backup, we are not taking it into consideration and still locking all tables.

Fix:
Adjust mdl_lock_tables to only lock tables that are being backed up. Also added a check to skip Percona Server compression dictionary tables. Those tables show up on the list but they do not accept queries.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -2176,47 +2176,47 @@ void mdl_unlock_all() {
     mysql_close(mdl_con);
   }
 }
-bool is_fts_index(const std::string &tablespace) {
+bool is_fts_index(const std::string &table_name) {
   const char *pattern =
       "^(FTS|fts)_[0-9a-fA-f]{16}_(([0-9a-fA-]{16}_(INDEX|index)_[1-6])|"
       "DELETED_CACHE|deleted_cache|DELETED|deleted|CONFIG|config|BEING_DELETED|"
       "being_deleted|BEING_DELETED_CACHE|beign_deleted_cache)$";
   const char *error_context = "is_fts_index";
-  return check_regexp_table_name(tablespace, error_context, pattern);
+  return check_regexp_table_name(table_name, error_context, pattern);
 }
 
-bool is_tmp_table(const std::string &tablespace) {
+bool is_tmp_table(const std::string &table_name) {
   const char *pattern = "^#sql";
   const char *error_context = "is_tmp_table";
-  return check_regexp_table_name(tablespace, error_context, pattern);
+  return check_regexp_table_name(table_name, error_context, pattern);
 }
 
-bool is_access_blocked(const std::string &tablespace) {
+bool is_access_blocked(const std::string &table_name) {
   const char *pattern = "compression_dictionary";
   const char *error_context = "is_blocked_table";
-  return check_regexp_table_name(tablespace, error_context, pattern);
+  return check_regexp_table_name(table_name, error_context, pattern);
 }
 
-bool check_regexp_table_name(std::string tablespace, const char *error_context,
+bool check_regexp_table_name(std::string table_name, const char *error_context,
                              const char *pattern) {
   bool result = false;
-  get_table_name_from_tablespace(tablespace);
+  get_table_name_from_fq(table_name);
   xb_regex_t preg;
   size_t nmatch = 1;
   xb_regmatch_t pmatch[1];
   compile_regex(pattern, error_context, &preg);
 
-  if (xb_regexec(&preg, tablespace.c_str(), nmatch, pmatch, 0) != REG_NOMATCH) {
+  if (xb_regexec(&preg, table_name.c_str(), nmatch, pmatch, 0) != REG_NOMATCH) {
     result = true;
   }
   xb_regfree(&preg);
   return result;
 }
-void get_table_name_from_tablespace(std::string &tablespace) {
-  std::size_t pos = tablespace.find("`.`");  //`db`.`table` separator
+void get_table_name_from_fq(std::string &fq_table_name) {
+  std::size_t pos = fq_table_name.find("`.`");  //`db`.`table` separator
   if (pos != std::string::npos) {
-    tablespace = tablespace.substr(pos + 3);
-    tablespace.erase(tablespace.size() - 1);  // remove leading `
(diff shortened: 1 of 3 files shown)


storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2137,13 +2137,21 @@ void mdl_lock_tables() {
       if (fsp_is_ibd_tablespace(atoi(row[1]))) {
         char full_table_name[MAX_FULL_NAME_LEN + 1];
         innobase_format_name(full_table_name, sizeof(full_table_name), row[0]);
+        if (check_if_skip_table(row[0]) &&
+            !fsp_is_dd_tablespace(atoi(row[1]))) {
```
satya-bodapati, line 2141: we do not check compression dictionary tables here?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2137,13 +2137,21 @@ void mdl_lock_tables() {
...
+          continue;
+        }
         if (is_fts_index(full_table_name)) {
           // We will eventually get to the row to lock the main table
           xb::info() << full_table_name << " is a Full-Text Index. Skipping";
           continue;
-        } else if (is_tmp_table(full_table_name)) {
+        } else if (is_tmp_table(full_table_name) ||
+                   (is_access_blocked(full_table_name) &&
+                    fsp_is_dd_tablespace(atoi(row[1])))) {
```
satya-bodapati, line 2151: any table in mysql schema to be skipped is invalid?

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -2183,6 +2191,12 @@ bool is_tmp_table(const std::string &tablespace) {
   return check_regexp_table_name(tablespace, error_context, pattern);
 }
 
+bool is_access_blocked(const std::string &tablespace) {
```
satya-bodapati: rename to is_compression_dict_table()

Should the parameter be named as table or tablespaces?

storage/innobase/xtrabackup/src/backup_mysql.h
```diff
@@ -245,6 +245,11 @@ bool is_fts_index(const std::string &tablespace);
 @return true if tablespace is temporary table. */
 bool is_tmp_table(const std::string &tablespace);
 
+/** Identifies if table is blocked by select - PS compression dictionary.
```
satya-bodapati: what is this 'blocking by select'?

storage/innobase/xtrabackup/src/backup_mysql.h
```diff
@@ -245,6 +245,11 @@ bool is_fts_index(const std::string &tablespace);
...
+@param[in]	tablespace		tablespace
+@return true if tablespace is blocked. */
```
satya-bodapati: am bit confused if we are usin tablespace or table name here?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: We always copy mysql.ibd, so even if the tables are from mysql.ibd will be excluded by `check_if_skip_table` we want to make sure we run MDL on mysql.ibd tables.

Regarding compression dictionary, they will be excluded in the if statement that was `is_access_blocked`

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo:
```
(is_access_blocked(full_table_name) &&
                    fsp_is_dd_tablespace(atoi(row[1])))
```


This ensure that if the table is from dd and is named compression_dictionary* we will not run a SELECT on it because PS does not allow selecting on those tables. This avoids PXB doing invalid queries.

storage/innobase/xtrabackup/src/backup_mysql.cc
altmannmarcelo: we pass `db`.`tbl`  format. If changing here, we should also change on `is_fts_index` and `is_tmp_table`

storage/innobase/xtrabackup/src/backup_mysql.h
altmannmarcelo:
```
mysql> SELECT * FROM mysql.compression_dictionary;
ERROR 3554 (HY000): Access to system table 'mysql.compression_dictionary' is rejected.
mysql>
```

storage/innobase/xtrabackup/src/backup_mysql.h
altmannmarcelo: name

storage/innobase/xtrabackup/src/backup_mysql.h
satya-bodapati: would be nice to add that SELECT from mysql.compression_* tables is not allowed by Server.

storage/innobase/xtrabackup/src/backup_mysql.cc
satya-bodapati: ok, not a must. but if you don't think its much, please go ahead

storage/innobase/xtrabackup/src/backup_mysql.cc
satya-bodapati: Got it. I guess it is un-necessary to do SELECT on mysql.* tables because DDLs on DD tables or to say system tables is not allowed by users. but no harm either. Ok keep to the current behaviour as is

storage/innobase/xtrabackup/src/backup_mysql.cc
satya-bodapati: ok, understood taht this avoid access denied errors from those compression tables.

satya-bodapati (approved):
Ok to merge after Jenkins run is clean.

PR #1489, opened 2023-08-29
Fast-Forward

altmannmarcelo:
This is a one-off merge to bring the trunk in sync with 8.0 changes after trunk branch creation.


satya-bodapati (approved):

.github/workflows/checks.yml
```diff
@@ -0,0 +1,45 @@
...
+  bidi_scan_job:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: '32'
+    - name: Run BiDi Scan
+      run: |
+        git fetch origin 8.0
```
satya-bodapati: @altmannmarcelo these should change? and all references to 8.0 in this file?

satya-bodapati (changes requested):

.github/workflows/checks.yml
altmannmarcelo: It does, this and clang-format, thankfully we adopted trunk and not 8.1, 8.2, 8.3 ...

satya-bodapati (approved):

PR #1495, opened 2023-09-13
PXB-3145 [8.0]: Remove a diff with upstream in "mysql-test/" directory

inikep:
1. Make a diff with upstream smaller to reduce a number of merge conflicts.

2. It will also help to reduce number of modifications added in a linear-history branch.


altmannmarcelo (approved):
LGTM

PR #1496, opened 2023-09-18
Fixed test cases and bootstrap

altmannmarcelo:
Adjusted keyring to always pass plugin dir.
Bootstrap to download correct glibc version

### DIFF
--- a/storage/innobase/xtrabackup/test/bootstrap.sh
+++ b/storage/innobase/xtrabackup/test/bootstrap.sh
@@ -22,13 +22,23 @@ function ssl_version() {
         100|101) ;;
         102) unset sslv; sslv="102.$OS" ;;
         *)
-            if ! test -r "${1}"; then
-                >&2 echo "tarball for your openssl version (${sslv}) is not available"
-                exit 1
-            fi
+            >&2 echo "tarball for your openssl version (${sslv}) is not available"
+            exit 1
             ;;
     esac
     echo ${sslv}
+}
+
+function glibc_version() {
+    glibc=$(ldd --version | head -1 | awk '{print $NF}')
+    case ${glibc} in
+        2.12|2.17|2.27|2.28|2.31|2.34|2.35) ;;
+        *)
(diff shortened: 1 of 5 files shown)


storage/innobase/xtrabackup/test/bootstrap.sh
```diff
@@ -31,6 +31,20 @@ function ssl_version() {
     echo ${sslv}
 }
 
+function glibc_version() {
+    glibc=$(ldd --version | head -1 | awk '{print $NF}')
+    case ${glibc} in
+        2.12|2.17|2.27|2.28|2.31|2.34|2.35) ;;
```
satya-bodapati, line 35: how do we know this is valid list of glibc versions?

storage/innobase/xtrabackup/test/bootstrap.sh
```diff
@@ -31,6 +31,20 @@ function ssl_version() {
...
+        *)
+            if ! test -r "${1}"; then
```
satya-bodapati: where do we get ${1} from?

storage/innobase/xtrabackup/test/bootstrap.sh
altmannmarcelo: This is the list from which we produce binaries from.

storage/innobase/xtrabackup/test/bootstrap.sh
altmannmarcelo: I think this all originated from https://github.com/percona/percona-xtrabackup/commit/6d0b9ea4172cb4b4c738838591444c2206dd858e , and I think this commit got the `if ! test -r "${1}"; then` from the main loop. In the SSL and GLIBC context this is not required. I will remove

satya-bodapati (approved):
please test and merge

altmannmarcelo:
@satya-bodapati all tests are now passing.

https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/367/#showFailuresLink

With this I discovered https://jira.percona.com/browse/PXB-3173 :tada:

PR #1498, opened 2023-10-02
8.0 Fix PXB-3147: Xtrabackup failed to execute query 'DO innodb_redo_log_…

altmannmarcelo:
…consumer_register("PXB");'

https://jira.percona.com/browse/PXB-3147

Problem:
Xtrabackup failed to execute query
'DO innodb_redo_log_consumer_register("PXB");' with double quotes were interpreted as identifiers.

Fix:
Force empty sql_mode for redo log consumer connection.

### DIFF
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -1185,6 +1185,10 @@ bool Redo_Log_Data_Manager::init() {
         xtrabackup_register_redo_log_consumer = false;
         return (false);
       }
+      /* Consumer queries might not work on some sql_mode.
+       * Forcing it to be empty.
+       */
+      xb_mysql_query(redo_log_consumer_cnx, "SET sql_mode=''", false, true);
       redo_log_consumer.init(redo_log_consumer_cnx);
       redo_log_consumer_can_advance.store(true);
     }
--- a/storage/innobase/xtrabackup/test/t/redo_log_consumer.sh
+++ b/storage/innobase/xtrabackup/test/t/redo_log_consumer.sh
@@ -72,3 +72,13 @@ kill -SIGCONT $xb_pid
 run_cmd wait $job_pid
 
 xtrabackup --prepare --target-dir=$topdir/backup
+
+# PXB-3147 - register redo log consumer fails with ANSI_QUOTES
+stop_server
+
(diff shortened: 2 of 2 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/361/#showFailuresLink - only the usual failures

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1185,6 +1185,7 @@ bool Redo_Log_Data_Manager::init() {
         xtrabackup_register_redo_log_consumer = false;
         return (false);
       }
+      xb_mysql_query(redo_log_consumer_cnx, "SET sql_mode=''", false, true);
```
satya-bodapati, line 1191: what about not usin double quotes?

altmannmarcelo: it's either that or enforce sql_mode. I thought about enforcing SQL mode so we do not get surprises on any other variation, or mix of sql_modes that are not compatible with the consumer.

satya-bodapati: Then we better do it for other connections from PXB too ?

altmannmarcelo: I would only change the sql_mode everywhere with more detailed research on all the queries on those connections. Consumer only executes three queries: register, advance, and deregister consumer.

satya-bodapati: ok, sounds resonable. Please add a comment in source, why we do this.

satya-bodapati: btw, will you merge the fix to 8.1?

altmannmarcelo: yeah, now we are back to old times. GCA between 8.0 and trunk and merge it into two branches

altmannmarcelo: btw, this branch was already created based on the GCA point between 8.0 and trunk

satya-bodapati: right, back to merge times.. tedious work..

satya-bodapati (approved):
comment style is new. If possible, please change it to existing style

PR #1501, opened 2023-10-17
Fix PXB-3170 - Test failure in keyring_pxb_2275.sh

altmannmarcelo:
https://jira.percona.com/browse/PXB-3170

Test was failing because we were generating an amount of data that was in a range that could cause the server to delete the required redo log. Also the test was failing due to missing redo log file, but we were not checking the error log for that which was a false positive.

Fix:
Adjusted the test to have a slightly higher redo log capacity. Added a check for the error log to make sure we fail due to missing keyring key.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/365/

satya-bodapati (approved):

PR #1503, opened 2023-10-18
Fix PXB-3003 - Race Condition in log_files_find_and_analyze

altmannmarcelo:
https://jira.percona.com/browse/PXB-3003

Problem:
log_files_find_and_analyze has a potential race condition when we list the available files and when we try to open them. In between those two operations, the log files might be deleted by the server causing the function to return an error.

Fix:
Adjust the function to tolerate missing files and return the list of valid files. Later we check if we found at least one file and if any of the files has the desired LSN.

Adjusted error message to info to make it more user-friendly, since we no longer error out in case of missing files.

### DIFF
--- a/storage/innobase/log/log0files_finder.cc
+++ b/storage/innobase/log/log0files_finder.cc
@@ -249,6 +249,9 @@ Log_files_find_result log_files_find_and_analyze(
     Log_files_dict &files, Log_format &format, std::string &creator_name,
     Log_flags &log_flags, Log_uuid &log_uuid) {
   ut::vector<Log_file_id_and_size> file_sizes;
+#ifdef XTRABACKUP
+  std::set<Log_file_id> missing_files;
+#endif
 
   switch (log_collect_existing_files(files.ctx(), read_only, file_sizes)) {
     case DB_NOT_FOUND:
@@ -267,7 +270,10 @@ Log_files_find_result log_files_find_and_analyze(
   ut::vector<Log_file_id_and_header> file_headers;
 #ifdef XTRABACKUP
   auto last_file_id = file_sizes.at(file_sizes.size() - 1).m_id;
+  debug_sync_point("log_files_find_and_analyze");
+
 #endif
+
   for (const auto &file : file_sizes) {
     Log_file_header file_header;
 
@@ -281,7 +287,12 @@ Log_files_find_result log_files_find_and_analyze(
     /* The file can be opened - checked by
     the log_collect_existing_files() */
     if (!file_handle.is_open()) {
+#ifdef XTRABACKUP
+      missing_files.insert(file.m_id);
+      continue;
+#else
       return Log_files_find_result::SYSTEM_ERROR;
+#endif /* XTRABACKUP */
     }
 
     if (log_file_header_read(file_handle, file_header) != DB_SUCCESS) {
@@ -302,6 +313,18 @@ Log_files_find_result log_files_find_and_analyze(
 
     file_headers.emplace_back(file.m_id, file_header);
   }
+
+#ifdef XTRABACKUP
+  if (missing_files.size() == file_sizes.size()) {
+    /* All files are missing. */
+    return Log_files_find_result::FOUND_NO_FILES;
+  }
+  file_sizes.erase(std::remove_if(file_sizes.begin(), file_sizes.end(),
+                                  [&missing_files](const auto &file) {
+                                    return missing_files.count(file.m_id) > 0;
+                                  }),
+                   file_sizes.end());
+#endif /* XTRABACKUP */
 
   /* Read properties global to the whole set of redo log files:
     - format,
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -46,6 +46,7 @@ static bool archive_first_block_zero = false;
 std::atomic<bool> Redo_Log_Reader::m_error;
 lsn_t Redo_Log_Reader::checkpoint_lsn_start;
 os_offset_t Redo_Log_Reader::checkpoint_offset_start;
+IF_DEBUG(bool force_reopen = false;);
 
 Redo_Log_Reader::Redo_Log_Reader() {
   log_hdr_buf.alloc_withkey(UT_NEW_THIS_FILE_PSI_KEY,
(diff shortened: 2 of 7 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/364/#showFailuresLink only the usual ones and #1501

storage/innobase/log/log0files_finder.cc
```diff
@@ -281,7 +288,12 @@ Log_files_find_result log_files_find_and_analyze(
     /* The file can be opened - checked by
     the log_collect_existing_files() */
     if (!file_handle.is_open()) {
+#ifdef XTRABACKUP
+      missing_files.insert(file_index++);
```
satya-bodapati: Is it possible to use file.m_id instead of new file_index?

storage/innobase/log/log0files_finder.cc
```diff
@@ -298,11 +310,29 @@ Log_files_find_result log_files_find_and_analyze(
...
+  /*
+   * Remove missing files from the list.
+   * Do it backwards to avoid reindexing. Eg.:
+   * files.erase(files.begin() + 0); - pos 1 will become pos 0
+   */
+  std::set<size_t>::iterator index = missing_files.end();
+  while (index != missing_files.begin()) {
+    --index;
+    file_sizes.erase(file_sizes.begin() + *index);
+  }
```
satya-bodapati: then we can use std::remove_if() with matches to saved file.m_ids ?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -153,7 +153,12 @@ static bool reopen_log_files(lsn_t desired_lsn) {
     return false;
   }
 
-  ut_a(res == Log_files_find_result::FOUND_VALID_FILES);
+  if (res != Log_files_find_result::FOUND_VALID_FILES) {
```
satya-bodapati, line 156: should we suggest our new redo log archiving option? (introduced in 8.0.34)?

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -46,6 +46,7 @@ static bool archive_first_block_zero = false;
 std::atomic<bool> Redo_Log_Reader::m_error;
 lsn_t Redo_Log_Reader::checkpoint_lsn_start;
 os_offset_t Redo_Log_Reader::checkpoint_offset_start;
+IF_DEBUG(bool force_reopen = false;);
```
satya-bodapati, line 49: better to move this near to the first usage (unless there is goto)

storage/innobase/xtrabackup/src/redo_log.cc
```diff
@@ -1289,6 +1305,11 @@ bool Redo_Log_Data_Manager::start() {
 
   debug_sync_point("xtrabackup_pause_after_redo_catchup");
 
+  DBUG_EXECUTE_IF("xtrabackup_reopen_files_after_catchup",
```
satya-bodapati, line 1310: It looks like the lack of control of debug sync for background threads is making our life difficult to control the threads at specific points. (FYI, you can close the comment after reading :))

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: This needs to be a global variable (or a member variable of this class). We change it to true at `start` and read it at `read_log_seg_8030`. So I think the declaration should stay here

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: You mean changing the wording at `Consider using Redo Log Archiving` ?

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: ok, didn't realize it is global. It can stay there

storage/innobase/xtrabackup/src/redo_log.cc
satya-bodapati: --redo-log-arch-dir=name

storage/innobase/log/log0files_finder.cc
altmannmarcelo: We are using an iterator here. If we try to remove this in place the for loop will jump one file. If this is a matter of keeping the missing_files set and just using the m_id, we should check if the files will not get reordered after a removal. I will check

satya-bodapati: Right, the final order is also important. The invariants before and after removal should remain same.

storage/innobase/xtrabackup/src/redo_log.cc
altmannmarcelo: Done

storage/innobase/log/log0files_finder.cc
altmannmarcelo: Done

storage/innobase/log/log0files_finder.cc
altmannmarcelo: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -8023,9 +8023,7 @@ int main(int argc, char **argv) {
   }
 
 #ifndef __WIN__
-  if (xtrabackup_debug_sync) {
```
satya-bodapati, line 8026: if condition intentionally removed?

storage/innobase/log/log0files_finder.cc
satya-bodapati: Thanks for implementing. It looks good now

storage/innobase/xtrabackup/src/xtrabackup.cc
altmannmarcelo: yes, we set debug sync via debug_execute_if (--debug) parameter. We need to handle the cont signal.

satya-bodapati: Ok, got it :+1:

satya-bodapati (approved):
LGTM!

PR #1515, opened 2023-10-26
Trunk Fix PXB-3003 - Race Condition in log_files_find_and_analyze

altmannmarcelo:
https://jira.percona.com/browse/PXB-3003

Problem:
log_files_find_and_analyze has a potential race condition when we list the available files and when we try to open them. In between those two operations, the log files might be deleted by the server causing the function to return an error.

Fix:
Adjust the function to tolerate missing files and return the list of valid files. Later we check if we found at least one file and if any of the files has the desired LSN.

Adjusted error message to info to make it more user-friendly, since we no longer error out in case of missing files.

### DIFF
--- a/storage/innobase/include/xb0xb.h
+++ b/storage/innobase/include/xb0xb.h
@@ -101,4 +101,10 @@ const std::string KEYRING_NOT_LOADED =
     "--keyring_file_data. If keyring component is used, check if "
     "--component-keyring-config points to valid configuration";
 
+/** pause xtrabackup and wait for resume.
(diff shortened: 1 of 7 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.1-test-param/27/#showFailuresLink

xbcloud tests were executed, but there was no s3 storage attached. I need to check Jenkins scripts - not related to this change.
Other failures are known.

satya-bodapati (approved):

PR #1516, opened 2023-10-26
 trunk Fix PXB-3141 - fix folder permission on fifo get

altmannmarcelo:
https://jira.percona.com/browse/PXB-3141

Problem
When creating a folder without executing permission, a user gets limited action on that folder. For a folder, the x privilege means the user can cd into the folder.
Currently, if we are doing a xbcloud get and the folder does not exists, xbcloud will create it as 600. The same problem does no happens with xtrabackup as the thread stream-dir as target-dir, and that is created as 777.

Fix
Adjusted folder creation on ds_fifo.cc and xbcloud to use 777 as permission. The same mask is used on xtrabackup.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.1-test-param/28/

failures:
https://jira.percona.com/browse/PXB-3189 - not related.

satya-bodapati (approved):

PR #1520, opened 2023-11-01
Fix PXB-3173 - Partial download functionality regression

altmannmarcelo:

Problem:
As part of FIFO redesign work, the xbcloud code was refactored to use multiple threads and a global list of files. This list was not taken into consideration that the user might want just a subset of files to be downloaded.
The test case was missing to enforce it should fail in case diff command return non zero return code.

Fix:
Re-implemented partial download functionality as a new function. Renamed the original variable to be more descriptive. Adjusted test to exit in case diff yields non-zero return code.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -140,7 +140,8 @@ static std::map<std::string, std::string> extra_http_headers;
 
 const char *s3_api_version_names[] = {"AUTO", "2", "4", NullS};
 
-static std::set<std::string> file_list;
+/* list of partial files to be downloaded or delete*/
+static std::set<std::string> partial_file_list;
 
 TYPELIB storage_typelib = {array_elements(storage_names) - 1, "", storage_names,
                            nullptr};
@@ -736,7 +737,7 @@ static bool parse_args(int argc, char **argv) {
   }
 
   while (argc > 0) {
-    file_list.insert(*argv);
+    partial_file_list.insert(*argv);
     --argc;
     ++argv;
   }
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/372/ -
Only error https://jira.percona.com/browse/PXB-3189

storage/innobase/xtrabackup/test/suites/xbcloud/basic_operations.sh
```diff
@@ -67,7 +67,7 @@ xbstream -xv -C $topdir/partial < $topdir/partial/partial.xbs \
 
 sort -o $topdir/partial/partial.list $topdir/partial/partial.list
 
-diff -u $topdir/partial/partial.list - <<EOF
+run_cmd diff -u $topdir/partial/partial.list - <<EOF
```
satya-bodapati, line 70: I think because of set -e mode, anything that returns nonzero should cause script exit. Can you please check why this doesn't happen?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -1058,8 +1077,7 @@ bool xbcloud_delete(Object_store *store, const std::string &container,
     if (!chunk_name_to_file_name(obj, file_name, idx)) {
       continue;
     }
-    if (!file_list.empty() &&
-        file_list.count(file_name.substr(backup_name.length() + 1)) < 1) {
+    if (skip_file(file_name, backup_name)) {
```
satya-bodapati, line 1080: This looks like refactor? :thinking: where exactly is the fix?

altmannmarcelo: This is indeed a refactor. Since the same logic is used on xbcloud_delete and xbcloud_download I created a function with this logic.

> where exactly is the fix?

It is at xbcloud_download by calling this function, which was not done before.

storage/innobase/xtrabackup/test/suites/xbcloud/basic_operations.sh
altmannmarcelo: That is correct, currently in the test framework we do not exit with error in case of any command return non-zero.
xtrabackup is a function that calls run_cmd xtrabackup. So this test was wrongly assuming diff will fail the test in case of any diff, while it should be doing it via run_cmd.

satya-bodapati: set -e is broken in pxb testframework. So everything has to be run_cmd :(

satya-bodapati (approved):

PR #1521, opened 2023-11-07
Trunk Fix PXB-3173 - Partial download functionality regression

altmannmarcelo:

Problem:
As part of FIFO redesign work, the xbcloud code was refactored to use multiple threads and a global list of files. This list was not taken into consideration that the user might want just a subset of files to be downloaded.
The test case was missing to enforce it should fail in case diff command return non zero return code.

Fix:
Re-implemented partial download functionality as a new function. Renamed the original variable to be more descriptive. Adjusted test to exit in case diff yields non-zero return code.


altmannmarcelo:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.1-test-param/34/

only test failure https://jira.percona.com/browse/PXB-3189

satya-bodapati (approved):

PR #1530, opened 2023-12-12
PXB-3034 Reduce the time the instance remains under lock

altmannmarcelo:
https://jira.percona.com/browse/PXB-3034


satya-bodapati:
This is outdated PR

PR #1531, opened 2023-12-15
PXB-3210: PXB fails to build on macOS since 8.0.33-28 due to FIND_PRO…

aybek:
…CPS()

https://jira.percona.com/browse/PXB-3210

When predicting required memory for prepare (PXB-2710) `sys/sysctl` is used by macOS and `proc/sysinfo` by other systems for memory calculation.
As part of the PXB-2993 fix to make the memory calculation logic compatible with procps-4 procps.cmake file was introduced to detect the version during cmake,
but cmake macro is not accounting for macOS and failing to find procps on macOS.
Now we make sure that cmake does not search for procps on macOS.


altmannmarcelo (changes requested):
Hi @aybek , it's not clear from the commit message what is done in oracle.
I assume procps is used in some feature, and procps is not available on apple. I would like to have a clear understanding as part of the commit message what is used instead and why this is not necessary on apple.

aybek:
> Hi @aybek , it's not clear from the commit message what is done in oracle. I assume procps is used in some feature, and procps is not available on apple. I would like to have a clear understanding as part of the commit message what is used instead and why this is not necessary on apple.

commit message updated, please take a look thanks

altmannmarcelo (approved):
LGTM

PR #1532, opened 2023-12-21
8.0 Fix PXB-3210: PXB fails to build on macOS since 8.0.33-28 due to FIND_PRO…

aybek:
…CPS()

https://jira.percona.com/browse/PXB-3210

When predicting required memory for prepare (PXB-2710) `sys/sysctl` is used by macOS and `proc/sysinfo` by other systems for memory calculation. As part of the PXB-2993 fix to make the memory calculation logic compatible with procps-4 procps.cmake file was introduced to detect the version during cmake, but cmake macro is not accounting for macOS and and failing to find procps on macOS. Now we make sure that cmake does not search for procps on macOS.


storage/innobase/xtrabackup/src/CMakeLists.txt
```diff
@@ -21,7 +21,10 @@ OPTION(WITH_VERSION_CHECK "Build with version check" ON)
 INCLUDE(${MYSQL_CMAKE_SCRIPT_DIR}/compile_flags.cmake)
 
 FIND_GCRYPT()
-FIND_PROCPS()
+
+IF(NOT APPLE)
+  FIND_PROCPS()
+ENDIF()
```
satya-bodapati, line 27: what is necessary for apple, do we need to add apple specific checks?

aybek: @satya-bodapati 
Actually I tried to explain this in commit message, probably I need to update it :)
`Procps` is used for calculation of memory required for `prepare` (PXB-2710) where  apple uses `sys/sysctl` so nothing else is necessary for apple.
It is just the check to not search for procps in apple was missing in cmake file

satya-bodapati (approved):
LGTM!

PR #1534, opened 2023-12-21
8.0 Fix PXB-2797: Schema mismatch when importing table with full-text index f…

aybek:
…rom xtrabackup backup

https://jira.percona.com/browse/PXB-2797

When importing a single table (IMPORT TABLESPACE) from a backup made using xtrabackup and the table contains a full-text index the import process will error out with: ERROR 1808 (HY000) at line 132: Schema mismatch (Index xxxxxx field xxxxxx is ascending which does not match metadata file which is descending)

The problem occurs because FTS tables are created internally by InnoDB Storage engine and they do not choose any ordering flag. Default is UNDEF. This patch solves the issue by treating UNDEF order as ASC when creating dict_table_t on --prepare step.

### DIFF
--- a/storage/innobase/dict/dict0dd.cc
+++ b/storage/innobase/dict/dict0dd.cc
@@ -438,6 +438,7 @@ ulint get_innobase_type_from_dd(const dd::Column *col, ulint &unsigned_type) {
   return MYSQL_TYPE_LONG;
 }
 
+#ifdef XTRABACKUP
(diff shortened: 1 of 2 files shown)


aybek:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/330/

satya-bodapati (approved):
LGTM!

PR #1538, opened 2024-01-17
Fix xbcloud out of range memory read

imrivera:
Fix a memory out of bounds read when `chunk_name.size() < 21`

```diff
- if (chunk_name.size() < 22 && chunk_name[chunk_name.size() - 21] != '.') {
+ if (chunk_name.size() < 22 || chunk_name[chunk_name.size() - 21] != '.') {
```

| chunk_name.size() |  Before (&&) | After (\|\|) |
| ------------------------- | ---------- | ------ |
| < 21                        |   - 1st condition is true <br> - 2nd condition **use negative index**           |   - 1st condition is true <br> - 2nd condition not evaluated      |
| == 21                     |   - 1st condition is true <br> - 2nd condition using index 0, ok but useless if we want names longer than 21 chars | - 1st condition is true <br> - 2nd condition not evaluated |
| >= 22                    |  - 1st condition is false <br> - 2nd condition is not evaluated | - 1st condition is false <br> - 2nd condition is evaluated with an index always > 0 |

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -1038,7 +1038,7 @@ cleanup:
 @return true in case of success or false otherwise */
 bool chunk_name_to_file_name(const std::string &chunk_name,
                              std::string &file_name, my_off_t &idx) {
-  if (chunk_name.size() < 22 && chunk_name[chunk_name.size() - 21] != '.') {
+  if (chunk_name.size() < 22 || chunk_name[chunk_name.size() - 21] != '.') {
     /* chunk name is invalid */
     return false;
   }


satya-bodapati:
Thank you for the contribution @imrivera ! Fix looks good. I will verify with ASAN/vaglrind tests and merge it to the next upcoming releases.

satya-bodapati:
Hi @imrivera, I just want to know if you reproduced this bug/behavior somehow. While I agree this is a bug from reading the code, I want to reproduce this bug. Since the default suffix name '.0000000000..' is 21 characters long, it should be impossible to get a chunk name less than this length.

Please let me know if you have a way to reproduce or how you found out this bug?

imrivera:
> Hi @imrivera, I just want to know if you reproduced this bug/behavior somehow. While I agree this is a bug from reading the code, I want to reproduce this bug. Since the default suffix name '.0000000000..' is 21 characters long, it should be impossible to get a chunk name less than this length.
> 
> Please let me know if you have a way to reproduce or how you found out this bug?

I found it just reading the code. I was trying to figure out some things about the naming scheme.

imrivera:
@satya-bodapati any news on merging this PR?

PR #1544, opened 2024-03-19
PKG-20 Create arm based packages for PXB 2.4

surbhat1595:


rfay:
Yay!

cmake/boost.cmake
```diff
@@ -41,7 +41,7 @@
 SET(BOOST_PACKAGE_NAME "boost_1_59_0")
 SET(BOOST_TARBALL "${BOOST_PACKAGE_NAME}.tar.gz")
 SET(BOOST_DOWNLOAD_URL
-  "http://jenkins.percona.com/downloads/boost/${BOOST_TARBALL}"
+  "https://downloads.percona.com/downloads/TESTING/issue-CUSTO83/boost/${BOOST_TARBALL}"
```
EvgeniyPatlan: it is better to move boost and mecab to separate directory. @adivinho can help you

EvgeniyPatlan (changes requested):

cmake/boost.cmake
surbhat1595: Yes, @adivinho moved it to a different location.

EvgeniyPatlan (approved):

PR #1549, opened 2024-04-11
Backup History: save to remote server for backup in RO replica

fmbiete:
Using the _--history_ option registers the backup metadata in the local database.

When using replication (classic, ReplicaSet, Cluster...), the replica is in RO (or SRO). If we want to execute the backups in a replica (to reduce impact in primary) and save the backup metadata we run into issues. This create an errant gtid (replica in RO) or fails (replica in SRO).

This adds the option to save the backup metadata in a remote server. In the case of a replication setup, we would run xtrabackup in the replica, but save the backup metadata in the primary.

## New xtrabackup options
- history-host: (string) the name of the MySQL server where we will save the backup metadata. Default to _host_
- history-port: (uint) the port of the MySQL server where we will save the backup metadata. Default to _port_
- history-user: (string) the name of the MySQL user that we will use to save the backup metadata. Default to _user_
- history-password: (string) the password for the MySQL user that we will use to save the backup metadata

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -192,6 +192,50 @@ MYSQL *xb_mysql_connect() {
   return (connection);
 }
 
+MYSQL *xb_mysql_history_connect() {
+  MYSQL *connection = mysql_init(NULL);
+  char mysql_port_str[std::numeric_limits<int>::digits10 + 3];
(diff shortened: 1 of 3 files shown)


satya-bodapati:
Thank you for your contribution, @fmbiete! I have created https://perconadev.atlassian.net/browse/PXB-3327 to track this pull request.

I will review this soon(at max a week). If not, please remind me once.

PR #1559, opened 2024-05-23
PXB-3283: Column %s precise type mismatch error while importing a tablespace (8.0)

venkatesh-prasad-v:
https://perconadev.atlassian.net/browse/PXB-3283

Problem & Analysis
==================
MySQL always stores temporal fields in my_charset_numeric charset.

When the collation of the table is changed using ALTER TABLE CONVERT TO CHARACTER SET, the collation of the temporal field was also updated in SDI.

When PXB prepares the backup (with --export) having such table, the .cfg file created by PXB sets the prtype based on the collation that is present in the SDI.

While importing the tablespace, the server fails with error

    "Schema mismatch (Column <col_name> precise type mismatch.)"

since InnoDB detects different prtype for the temporal column.

However, this problem is not seen when we import .cfg file generated by FLUSH TABLE FOR EXPORT since FLUSH TABLE FOR EXPORT always uses the charset of the table->field (my_charset_numeric) and sets prtype using my_charset_numeric as the collation.

Solution
========
The problem of server allowing ALTER TABLE CONVERT TO CHARACTER SET command to alter the character set temporal feilds has been fixed in Percona Server by PS-9129. But in order to remain consistent with the server behavior, PXB now forces all temporal fields to use my_charset_numeric as the charset until MySQL server provides support for a separate charset/collation for temporal fields.

Testing Done
========
Jenkins: https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/334/console
NOTE: Testing branch is based on 8.0, but this PR branch is based on the gca commit c9c863c


storage/innobase/include/data0type.h
```diff
@@ -315,6 +320,16 @@ ulint dtype_get_at_most_n_mbchars(
  @return true if string type */
 bool dtype_is_string_type(
     ulint mtype); /*!< in: InnoDB main data type code: DATA_CHAR, ... */
+
+#ifdef XTRABACKUP
+/** Checks if a data main type is a real temporal type.
+ @return true if real temporal type */
```
satya-bodapati: @doxygen style comments please. What is real and not real? can u explain in comment?

storage/innobase/include/data0type.h
```diff
@@ -315,6 +320,16 @@ ulint dtype_get_at_most_n_mbchars(
...
+bool dtype_is_real_temporal_type(
+    dd::enum_column_types
+        type); /*!< in: InnoDB main data type code: TINY, SHORT, ... */
```
satya-bodapati: remove the in:.. and make param[in]

storage/innobase/xtrabackup/test/t/xb_export.sh
```diff
@@ -64,12 +64,28 @@ function insert_data() {
 }
 
 function create_tables() {
-  ### check if we can import all the columns type in mysql
```
satya-bodapati, line 67: we already have datetime here, why didn't this test fail?
I guess, that could be because, we only care about the number of .cfg files created but didn't do actual import? :thinking:

storage/innobase/dict/dict0dd.cc
```diff
@@ -813,7 +813,23 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
...
+
+        It is for the same reason, the .cfg file exported during FLUSH TABLE
+        FOR EXPORT always sets prtype using my_charset_numeric as the
+        collation.
+
+        In order to remain consistent with server's behavior, we force temporal
+        datatypes to use my_charset_numeric until MySQL server provides support
+        for a separate charset/collation for temporal fields.
+
+        TODO: Check if this needs to be done only for Percona variants?
```
satya-bodapati: reomve the TODO comment

storage/innobase/dict/dict0dd.cc
```diff
@@ -813,7 +813,23 @@ dict_table_t *dd_table_create_on_dd_obj(const dd::Table *dd_table,
       charset_no = my_charset_bin.number;
     else if (dd_col->type() == dd::enum_column_types::JSON)
       charset_no = my_charset_utf8mb4_bin.number;
-    else if (dtype_is_string_type(mtype)) {
+    else if (dtype_is_real_temporal_type(dd_col->type())) {
+      /*
+        MySQL always stores real temporal fields in my_charset_numeric charset.
```
satya-bodapati, line 818: I think this should be mentioned in is_real_temporal_type function comment.

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/t/xb_export.sh
venkatesh-prasad-v: It didnot fail because there was no scenario for testing temporal columns with different collations.

storage/innobase/include/data0type.h
venkatesh-prasad-v: I copied the format from the previous function declaration. I'll remove it.

storage/innobase/xtrabackup/test/t/xb_export.sh
satya-bodapati: I see, if we have  explicit collation at create table column, it would fail too?

venkatesh-prasad-v: No. Explicit collation given at the column declaration are ignored by server for temporal columns.

satya-bodapati: ah, so ALTER is compulsory?

venkatesh-prasad-v: Yes. Only through ALTER TABLE COVERT TO CHARACTER SET server allowed collation to change. However we are disallowing in server it as part of PS-9219. So MySQL won't be able to change the charsets of temporal types.

storage/innobase/dict/dict0dd.cc
venkatesh-prasad-v: IMO this comment may not be appropriate for a function comment since the function's responsibility is only to check if the data type is temporal or not. So I prefer to have it here where we take further decision.

storage/innobase/include/data0type.h
venkatesh-prasad-v: From what I have seen and studied, `real type` has different meanings in different contexts. It can mean

- a data type allowing fractional numbers for precision
- an underlying datatype used internally. For example binlog stores  CHAR, VARCHAR, SET and ENUM as MYSQL_TYPE_STRING. For such types, `field->type()` will be MYSQL_TYPE_STRING and `field->real_type()` will return the actual underlying data type.

However in the context of collation, temporal real types are the ones that have fractional numbers for precision as they need to be stored and displayed as FIXBINARY. So, it is safe to assume that `real temporal type` are the types which allow with fractional seconds.

storage/innobase/include/data0type.h
venkatesh-prasad-v: Done

satya-bodapati (approved):
Thanks for fixing PXB bug @venkatesh-prasad-v 
I am sure now I can assign few pxb bugs to you :)!
Good work

satya-bodapati:
@venkatesh-prasad-v you might have to use GCA and merge. Or we might face some issues because PXB 8.0 is still on 8.0.35

venkatesh-prasad-v:
@satya-bodapati This branch is based on gca commit c9c863c2973.

PR #1580, opened 2024-07-11
PXB-3240 MySQL 8.4 Merge 

aybek:


storage/innobase/log/log0recv.cc
```diff
@@ -3950,8 +3921,8 @@ bool meb_scan_log_recs(
         }
 #endif /* UNIV_DEBUG && HAVE_ASAN */
         more_data =
-            recv_sys_add_to_parsing_buf(log_block, scanned_lsn) || more_data;
->>>>>>> mysql-8.4.0
+            recv_sys_add_to_parsing_buf(log_block, scanned_lsn, data_len) ||
```
satya-bodapati: Do u know the purpose of data_len? Should we also pass to recv_sys_parse_byte_by_byte()?

storage/innobase/log/log0recv.cc
```diff
@@ -238,29 +238,6 @@ static bool recv_writer_is_active() {
...
-@param[in,out]  log             redo log
-@param[in,out]  buf             buffer where to read
-@param[in]      start_lsn       read area start
-@param[in]      end_lsn         read area end
-@return lsn up to which data was available on disk (ideally end_lsn)
-or zero in case of error */
-static lsn_t recv_read_log_seg(log_t &log, byte *buf, lsn_t start_lsn,
-                               lsn_t end_lsn);
-
->>>>>>> mysql-8.4.0
```
satya-bodapati: recv_read_log_seg is competely removed by upstream? or removed by us?

storage/innobase/log/log0recv.cc
```diff
@@ -2682,8 +2659,8 @@ static pxb_space_page *recv_get_page_map(space_id_t space_id) {
  * later when we run the --prepare phase.
  */
 static void recv_calculate_hash_heap(mlog_id_t type, space_id_t space_id,
-                                     page_no_t page_no, byte *body,
-                                     byte *rec_end, lsn_t start_lsn) {
```
satya-bodapati: I agree with constness, are they coming from upstream or u have added them?

sql/derror.cc
```diff
@@ -237,44 +237,10 @@ bool MY_LOCALE_ERRMSGS::read_texts() {
...
-      Trying pre-5.5 semantics of the --language parameter.
-      It included the language-specific part, e.g.:
-
-      --language=/path/to/english/
-    */
-    if ((file = mysql_file_open(
-             key_file_ERRMSG,
-             fn_format(name, ERRMSG_FILE, lc_messages_dir, "", 4), O_RDONLY,
-             MYF(0))) < 0) {
 #ifndef XTRABACKUP
```
satya-bodapati: Can you please check what happens if u remove this ifndef? is pxb affected? do we get this ER_ERRMSG_CANT_FIND_FILE printed?

storage/innobase/fil/fil0fil.cc
```diff
@@ -10671,22 +10654,12 @@ const byte *fil_tablespace_redo_rename(const byte *ptr, const byte *end,
     return nullptr;
   }
 
-<<<<<<< HEAD
-  ptr += to_len;
-  Fil_path::normalize(to_name);
-
```
satya-bodapati: we have special code with PXB. if parse_only, return ptr.

In the old code, there is ptr is incrementd by to_len. With the new code, there is no to_len. Is the ptr moved else where?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2037,10 +2036,6 @@ static bool innodb_init_param(void) {
     }
   }
 
-  os_innodb_umask = (ulint)0664;
-
-  os_file_set_umask(my_umask);
```
satya-bodapati: where is this set before? Also, can you please check if this os_file-set_umask() is removed from innodb_init_param in InnoDB sources?

satya-bodapati (changes requested):
Good work Aibek. I have left a few comments/questions.

We should also remove keyring plugin variants of vault and other plugins.
Also, modify vault plugin tests to component variants.

You may need PS 8.4 release to test but we can use release PS 8.4 branch.

storage/innobase/log/log0recv.cc
aybek: they are coming from upstream

storage/innobase/fil/fil0fil.cc
aybek: yes this logic is moved to parse_path_from_redo()

storage/innobase/log/log0recv.cc
satya-bodapati: Okay, this is also present in earlier versions (PXB 8.3/8.2, etc.), but to me, it doesn't make much sense.
Caller also passes log_block_get_data_len(block) as data_len.

There is no need to do an investigation now.

Closing the conversation for now.

sql/derror.cc
aybek: yes we get this message printed if we remove ifndef

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: if you mean ha_innodb.cc file os_file_set_umask() was not removed but has been surrounded by ifndef _WIN32;
can't find usage of os_file_set_umask anywhere else before

https://github.com/aybek/percona-xtrabackup/blob/pxb-3240/storage/innobase/handler/ha_innodb.cc#L5509

satya-bodapati: No, I mean the code in xtrabackup.cc
innob_init_params is duplicated in xtrabackup.cc and the os_file_set_umask() is removed this function in xtrbackup.cc 

By removing this mask, file permission in backup directory is changed from 0664 to 0614.

But if you bring back the set_umask(), prepare will fail with assertion because we reboot (and so reinit) innodb multiple times.

We have to allow the double init incase of pxb. May be introduce a de-init function of umask() and it allows the reinit.

storage/innobase/log/log0recv.cc
satya-bodapati: this is a misleading diff. recv_read_log_seg() is present in both. 
It is just that we make it non-static. the reason PXB redo log parser calls this function to parse redo log.

Closing this conversation as resolved

sql/derror.cc
satya-bodapati: thanks, next time I would prefer a more detailed reply. the stack trace is
(gdb) bt
#0  MY_LOCALE_ERRMSGS::read_texts (this=0x5555610ce680 <global_errmsgs>) at /home/satya/WORK/pxb-8.4/sql/derror.cc:241
#1  0x000055555911f0a8 in setup_error_messages () at /home/satya/WORK/pxb-8.4/storage/innobase/xtrabackup/src/xtrabackup.cc:7364
#2  0x0000555559144659 in main (argc=4, argv=0x7fffffffde18) at /home/satya/WORK/pxb-8.4/storage/innobase/xtrabackup/src/xtrabackup.cc:7415

and the error message printed is:
2024-07-26T14:14:07.764069+01:00 0 [ERROR] [MY-010338] [Server] Can't find error-message file 'english/errmsg.sys'. Check error-message file location and 'lc-messages-dir' configuration directive.

closing this conversation as marked

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: added the umask resetting option for xtrabackup in the latest commit, please take a look

storage/innobase/os/os0file.cc
```diff
@@ -7884,11 +7885,15 @@ void os_aio_print_pending_io(FILE *file) { AIO::print_to_file(file); }
...
-  static bool was_already_set{false};
-  ut_a(!was_already_set);
-  was_already_set = true;
+  ut_a(!umask_was_already_set);
+  umask_was_already_set = true;
   os_innodb_umask = umask;
 }
+
+#ifdef XTRABACKUP
+void os_file_allow_reset_umask() { umask_was_already_set = false; }
```
satya-bodapati: and set the os_innodb_umask to default / some _MAX value in this reset.

satya-bodapati (changes requested):

storage/innobase/os/os0file.cc
aybek: thanks Satya, done in the latest commit

satya-bodapati:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.1-test-param/94/testReport/

satya-bodapati:
on OEL9 platforms, perl mysql DBD library missing. Remaining failures are known failures.

satya-bodapati (approved):
LGTM! :clap:

PR #1603, opened 2024-08-27
PXB-3269 Reduce the time the Server is locked by xtrabackup

aybek:


storage/innobase/fil/fil0fil.cc
```diff
@@ -1645,6 +1650,8 @@ class Fil_system {
   @return DB_SUCCESS if the open was successful */
   [[nodiscard]] dberr_t open_for_recovery(space_id_t space_id);
 
+  [[nodiscard]] dberr_t open_for_reduced(const std::string &path);
```
satya-bodapati, line 1664: Add function comment in doxygen style

storage/innobase/fil/fil0fil.cc
```diff
@@ -3371,6 +3398,10 @@ static bool fil_space_free(space_id_t space_id, bool x_latched) {
   return true;
 }
 
+space_id_t fil_get_tablespace_id(const std::string &filename) {
+  return Fil_system::get_tablespace_id(filename);
+}
```
satya-bodapati, line 3478: entire function under ifdef XTRABACKUP?

storage/innobase/fil/fil0fil.cc
```diff
@@ -10183,6 +10208,22 @@ dberr_t fil_tablespace_open_for_recovery(space_id_t space_id) {
...
+dberr_t Fil_system::open_for_reduced(const std::string &path) {
+  space_id_t space_id = get_tablespace_id(path);
+  fil_space_t *space;
+
+  auto status = ibd_open_for_recovery(space_id, path, space);
+
+  if (status == FIL_LOAD_DBWLR_CORRUPTION) {
+    return DB_CORRUPTION;
+  }
+  return DB_SUCCESS;
```
satya-bodapati: status is a enum and other states are not handled, please handle all of them and convert to DB_ * error codes.

storage/innobase/include/fil0fil.h
```diff
@@ -2247,16 +2243,19 @@ void fil_set_scan_dir(const std::string &directory, bool is_undo_dir = false);
...
-dberr_t fil_scan_for_tablespaces(bool populate_fil_cache);
+dberr_t fil_scan_for_tablespaces(bool populate_fil_cache, bool only_undo);
 
 /** Open the tablespace and also get the tablespace filenames, space_id must
 already be known.
 @param[in]  space_id  Tablespace ID to lookup
 @return DB_SUCCESS if open was successful */
 [[nodiscard]] dberr_t fil_tablespace_open_for_recovery(space_id_t space_id);
 
+dberr_t fil_open_for_reduced(const std::string &path);
```
satya-bodapati, line 2268: Add function comment? Please add some comment this is a wrapper to fil_tablespace_open_for_recovery?

storage/innobase/fil/fil0fil.cc
```diff
@@ -10183,6 +10208,22 @@ dberr_t fil_tablespace_open_for_recovery(space_id_t space_id) {
   return fil_system->open_for_recovery(space_id);
 }
 
+dberr_t Fil_system::open_for_reduced(const std::string &path) {
+  space_id_t space_id = get_tablespace_id(path);
+  fil_space_t *space;
+
+  auto status = ibd_open_for_recovery(space_id, path, space);
```
satya-bodapati: once we get space_id, can we just call fil_tablespace_open_for_recovery(space_id) ?

storage/innobase/fil/fil0fil.cc
satya-bodapati: if we use fil_tablespace_open_for_recovery(), then this comment might be irrelevant?

storage/innobase/fil/fil0fil.cc
```diff
@@ -10505,11 +10546,16 @@ static void fil_make_abs_file_path(const char *file_name, ulint flags,
...
 @param[in]      parse_only      Don't apply, parse only
+@param[in]      start_lsn       LSN of the redo record
 @return pointer to next redo log record
 @retval nullptr if this log record was truncated */
-const byte *fil_tablespace_redo_create(const byte *ptr, const byte *end,
-                                       const page_id_t &page_id,
-                                       ulint parsed_bytes, bool parse_only) {
+const byte *fil_tablespace_redo_create(
+    const byte *ptr, const byte *end, const page_id_t &page_id,
+    ulint parsed_bytes, bool parse_only IF_XB(, lsn_t start_lsn)) {
```
satya-bodapati: lets rename this to record_lsn instead of start_lsn.  This function and other fil_tablespace_redo_* functions.

storage/innobase/fil/fil0fil.cc
```diff
@@ -11543,15 +11644,37 @@ void Tablespace_dirs::open_ibd(const Const_iter &start, const Const_iter &end,
...
+        /* table was renamed during scan. Since original table was loaded to
+        cache and rename ddl will be handled by ddl_tracker we skip loading
+        duplicate tablespace */
+        continue;
+      } else
+        /* PXB-2275 - Allow DB_INVALID_ENCRYPTION_META as we will test it in
+        the end of the backup */
+        if (err != DB_SUCCESS && err != DB_INVALID_ENCRYPTION_META) {
+          result = false;
+        }
```
satya-bodapati: can you please put braces aroudn the the last else {}

storage/innobase/fsp/fsp0file.cc
```diff
@@ -122,9 +122,11 @@ dberr_t Datafile::open_read_only(bool strict) {
...
 
   set_open_flags(OS_FILE_OPEN);
-  m_handle = os_file_create_simple_no_error_handling(
-      innodb_data_file_key, m_filepath, m_open_flags, OS_FILE_READ_ONLY, true,
-      &success);
+
+  m_handle = os_file_create(
+      innodb_data_file_key, m_filepath,
+      OS_FILE_OPEN | OS_FILE_ON_ERROR_SILENT | OS_FILE_ON_ERROR_NO_EXIT,
+      OS_FILE_NORMAL, OS_DATA_FILE, OS_FILE_READ_ONLY, &success);
```
satya-bodapati: I think, I mistakenly removed upstream code. Better to put xtrbackup code under #ifdef XTRABACKUP.

THe reason for this change is 
/* File discovery under xtrabackup should tolerate disappearing IBD files on server and we don't want to log warnings or errors. The DDL  tracker will track these deletions */

storage/innobase/include/fil0fil.h
```diff
@@ -2111,27 +2110,25 @@ inline fil_space_t *fil_space_get_sys_space() {
 @param[in]      page_id         Tablespace Id and first page in file
 @param[in]      parsed_bytes    Number of bytes parsed so far
 @param[in]      parse_only      Don't apply, parse only
+@param[in]      start_lsn       LSN of the redo record
```
satya-bodapati: as noted earlier, lets rename this to record_lsn

storage/innobase/include/fil0fil.h
```diff
@@ -2375,4 +2380,11 @@ inline bool fil_node_t::is_offset_valid(os_offset_t byte_offset) const {
   return byte_offset < max_offset;
 }
 
+/** Frees a space object from the tablespace memory cache.
+Closes a tablespaces' files but does not delete them.
```
satya-bodapati: remove the extra '

storage/innobase/log/log0recv.cc
```diff
@@ -1688,24 +1689,32 @@ static const byte *recv_parse_or_apply_log_rec_body(
       return fil_tablespace_redo_delete(
           ptr, end_ptr, page_id_t(space_id, page_no), parsed_bytes,
           recv_sys->bytes_to_ignore_before_checkpoint !=
-              0 IF_XB(|| recv_sys->recovered_lsn + parsed_bytes <
-                             backup_redo_log_flushed_lsn));
+              0 IF_XB(||
+                      recv_sys->recovered_lsn + parsed_bytes <
+                          backup_redo_log_flushed_lsn ||
+                      opt_lock_ddl == LOCK_DDL_REDUCED),
+          start_lsn);
```
satya-bodapati, line 1696: Lets not change the start_lsn to record_lsn here and the callers of this function. These are coming from upstream.
Only fil_tablespace_redo_*() or below is sufficient for us

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -143,6 +143,16 @@ static int original_innodb_buffer_pool_dump_pct;
...
 
+/* This could be either LOCK TABLES FOR BACKUP or LOCK INSTANCE */
+static xtrabackup::utils::time backup_lock_start_time =
+    xtrabackup::utils::INVALID_TIME;
+static xtrabackup::utils::time backup_lock_end_time =
+    xtrabackup::utils::INVALID_TIME;
+
+static xtrabackup::utils::time ftwrl_start_time =
+    xtrabackup::utils::INVALID_TIME;
+static xtrabackup::utils::time ftwrl_end_time = xtrabackup::utils::INVALID_TIME;
```
satya-bodapati, line 155: under ifdef XTRABACUP

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+#endif  // UNIV_DEBUG
+      break;
+  }
+}
+
+void ddl_tracker_t::add_table_from_ibd_scan(const space_id_t space_id,
+                                            std::string name) {
+  Fil_path::normalize(name);
+  if (Fil_path::has_prefix(name, Fil_path::DOT_SLASH)) {
+    name.erase(0, 2);
```
satya-bodapati: instead of 2, strlen(Fil_path::DOT_SLASH) or DOT_SLASH.length()?

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+  page. Can happen only on general tablespaces and mysql.ibd */
+  std::unordered_map<space_id_t, std::string> corrupted_tablespaces;
+  /** Multiple copy threads can add entries to corrupted_tablespaces and
+  recopy_tables concurrently */
+  std::mutex m_ddl_tracker_mutex;
+
+  std::tuple<filevec, filevec> handle_undo_ddls();
+
+ public:
+  void add_undo_tablespace(const space_id_t space_id, std::string name);
```
satya-bodapati, line 93: We have not clearly explained how undo tablespaces are tracked. We should add detailed function comment here.

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+void ddl_tracker_t::add_corrupted_tablespace(const space_id_t space_id,
+                                             const std::string &path) {
+  std::lock_guard<std::mutex> lock(m_ddl_tracker_mutex);
+
+  corrupted_tablespaces[space_id] = path;
+}
+
+void ddl_tracker_t::add_to_recopy_tables(space_id_t space_id, lsn_t start_lsn,
+                                         const std::string operation) {
+  // There should never be MLOG_INDEX_LOAD. However MLOG_WRITE_STRING on
```
satya-bodapati: There should never be MLOG_INDEX_LOAD on undo tablespaces.

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  datafiles_iter_t *it;
+  uint num;
+  uint *count;
+  ib_mutex_t *count_mutex;
+  bool *error;
+  std::thread::id id;
+  space_id_to_name_t *new_tables;
+} copy_thread_ctxt_t;
+
+static void data_copy_thread_func(copy_thread_ctxt_t *ctxt) {
```
satya-bodapati: I think these are copy of original copy thread code. Lets rename to copy_for_reduced().

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+
+  /*
+    Initialize mysys thread-specific memory so we can
+    use mysys functions in this thread.
+  */
+  my_thread_init();
+
+  /* create THD to get thread number in the error log */
+  THD *thd = create_thd(false, false, true, 0, 0);
+  debug_sync_point("data_copy_thread_func");
```
satya-bodapati: debug_sync_point here should be renamed and the corresponding usage in testcase should be renamed too

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+
+  mutex_enter(ctxt->count_mutex);
+  (*ctxt->count)--;
+  mutex_exit(ctxt->count_mutex);
+
+  destroy_thd(thd);
+  my_thread_end();
+}
+
+/**
```
satya-bodapati, line 380: Function comment first line missing.
Please give an example in the comment

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  undo_spaces_init();
+  xb_scan_for_tablespaces(true);
+
+  // Generates the after_lock_undo list
+  srv_undo_tablespaces_init(false, true);
+
+  std::ostringstream str;
+  str << "Before UNDO files" << to_string(before_lock_undo) << "\n"
+      << "After UNDO files" << to_string(after_lock_undo) << "\n";
+  fprintf(stderr, "%s\n", str.str().c_str());
```
satya-bodapati, line 476: Add a comment here that xb::info() is not used intentionally, as there is a loss of newlines characters from the stream/string.

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  std::ostringstream str;
+  str << "Tables_copied: " << to_string(tables_in_backup) << "\n"
+      << "New_tables: " << to_string(new_tables) << "\n"
+      << "Renames: " << to_string(renames) << "\n"
+      << "Drops: " << to_string(drops) << "\n"
+      << "Recopy_tables: " << to_string(recopy_tables) << "\n"
+      << "Corrupted_tablespaces: " << to_string(corrupted_tablespaces) << "\n"
+      << "Missing tables: " << to_string(missing_tables) << "\n"
+      << "Renamed_during_scan: " << to_string(renamed_during_scan) << "\n";
+  fprintf(stderr, "%s\n", str.str().c_str());
```
satya-bodapati, line 526: same as above. comment on why fprintf() instead of xb::info().

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+                                         &flush_lsn);
+
+      if (err != DB_SUCCESS) {
+        xb::error() << "could not reopen system tablespace";
+        ut_ad(0);
+        return;
+      }
+
+    } else {
+      std::tie(err, std::ignore) = fil_open_for_xtrabackup(
```
satya-bodapati, line 691: what to do if we cannot open a tablespace after taking the lock ?
We should abort the backup here. May an exit() would do 
[ I prefer to return to the callers way up and cleanup all subsystem but that is big refactoring change. Lets ignore for now]

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+          table->second, table->second.substr(0, table->second.length() - 4));
```
satya-bodapati: why 4?

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  // tables in system tablespace need not be tracked. They can be
+  // created via redo log
+  if (fsp_is_system_tablespace(space_id)) {
+    return;
+  }
+
+  std::string new_space_name = name;
+  Fil_path::normalize(new_space_name);
+  if (Fil_path::has_prefix(new_space_name, Fil_path::DOT_SLASH)) {
+    new_space_name.erase(0, 2);
```
satya-bodapati: dont use 2, strlen() or length

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  // DROP table in system tablespace is drop indexes. There is no file removal
+  // this is also tracked/redone from redo
+  if (fsp_is_system_tablespace(space_id)) {
+    return;
+  }
+
+  std::string new_space_name{name};
+  Fil_path::normalize(new_space_name);
+  if (Fil_path::has_prefix(new_space_name, Fil_path::DOT_SLASH)) {
+    new_space_name.erase(0, 2);
```
satya-bodapati: here and everywhere, about 2

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  /* Wait for threads to exit */
+  while (1) {
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    mutex_enter(&count_mutex);
+    if (count == 0) {
+      mutex_exit(&count_mutex);
+      break;
+    }
+    mutex_exit(&count_mutex);
+  }
```
satya-bodapati: Print a msg after sleep? Waiting for the recopy of tables to be finished. Expecting count be 0, current count is ?

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+
+  if (data_copying_error) {
```
satya-bodapati, line 763: Add xb::error() here before exit(). That there is problem with recopy tablespaces under lock (reduced mode)

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  meta_map.insert({space_id, meta_path});
+}
+
+std::tuple<bool, std::string> is_in_meta_map(space_id_t space_id) {
+  auto it = meta_map.find(space_id);
+  bool exists = it != meta_map.end();
+  return {exists, exists ? it->second : ""};
+}
+
+/** Read file content into STL string */
```
satya-bodapati: modernize to c++17 (chatgpt version. Please use xb::info() instead cerr etc.)

```
#include <fstream>
#include <string>
#include <filesystem>
#include <stdexcept>

static std::string read_file_as_string(const std::string& file) {
    try {
        // Check if the file exists using std::filesystem
        if (!std::filesystem::exists(file)) {
            return "";
        }

        std::ifstream f(file, std::ios::in | std::ios::binary);
        if (!f) {
            return "";
        }

        // Read entire file into a string
        std::string content((std::istreambuf_iterator<char>(f)),
                            std::istreambuf_iterator<char>());

        return content;
    } catch (const std::filesystem::filesystem_error& e) {
        // Handle filesystem-related exceptions
        // (e.g., file doesn't exist, permissions issues)
        std::cerr << "Filesystem error: " << e.what() << '\n';
        return "";
    } catch (const std::ifstream::failure& e) {
        // Handle ifstream-related exceptions
        // (e.g., file couldn't be opened, read errors)
        std::cerr << "File read error: " << e.what() << '\n';
        return "";
    } catch (const std::exception& e) {
        // Catch all other standard exceptions
        std::cerr << "Error: " << e.what() << '\n';
        return "";
    }
}

```

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  std::string dest_dir = to.substr(0, to.find_last_of("/"));
+  /* create target dir if not exist */
+  if (!my_stat(dest_dir.c_str(), &stat_info, MYF(0)) &&
+      (my_mkdir(dest_dir.c_str(), 0750, MYF(0)) < 0)) {
+    xb::error() << "cannot mkdir: " << my_errno() << " " << dest_dir;
+    exit(EXIT_FAILURE);
+  }
+  rename_file(from, to);
+}
+
```
satya-bodapati, line 884: as noted in slack DM:

```
if source file doesn't exist but destination file exists with the desired name -> not an error
if source file doesn't exists but destination file DO NOT exist with desired name -> error
if source file exists, and desitnation file exists with desired name -> error (to be validated)
if source file exists, and desitnation file doesn't exist with desired name -> not an error
```

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  return (true);
+}
+
+/**
+ * Handle DDL for renamed files
+ * example input: test/10.ren file with content = test/new_name.ibd ;
+ *        result: tablespace with space_id=10 will be renamed to
+ * test/new_name.ibd
+ * @return true on success
+ */
```
satya-bodapati: parameter comment missing.

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+bool prepare_handle_ren_files(
+    const datadir_entry_t &entry, /*!<in: datadir entry */
+    void * /*data*/) {
```
satya-bodapati: remove trailing comments, this is old style.. most likely this is copy paste

Applies to other places with parameter comments like this.

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+
+  // delete the .ren file, we don't need it anymore
+  os_file_delete(0, ren_path.c_str());
+  return true;
+}
+
+/**
+ * Handle .corrupt files. These files should be removed before we do *.ibd scan
+ * @return true on success
+ * */
```
satya-bodapati: parameter comment fix

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+    std::string meta_file = source_path + EXT_META;
+    xb::info() << "prepare_handle_corrupt_files: deleting  " << meta_file;
+    os_file_delete_if_exists_func(meta_file.c_str(), nullptr);
+  } else {
+    xb::info() << "prepare_handle_corrupt_files: deleting  " << source_path;
+    os_file_delete_if_exists_func(source_path.c_str(), nullptr);
+  }
+
+  // delete the .corrupt file, we don't need it anymore
+  os_file_delete_if_exists_func(corrupt_path.c_str(), nullptr);
```
satya-bodapati: what about return types of os_file_delte_if_exists..()

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+  name_to_space_id_t before_lock_undo;
+  name_to_space_id_t after_lock_undo;
+  /** Tablespaces involved in encryption or bulk index load.*/
+  std::unordered_set<space_id_t> recopy_tables;
+  /** Drop operations found in redo log. */
+  space_id_to_name_t drops;
+  /* For DDL operation found in redo log,  */
+  std::unordered_map<space_id_t, std::pair<std::string, std::string>> renames;
+  /** Tables that have been deleted between discovery and file open */
+  std::unordered_set<std::string> missing_tables;
```
satya-bodapati: instead of missing_tables, please use missing_afer_discovery.

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+  /** Function responsible to generate files based on DDL operations */
+  void handle_ddl_operations();
+  /** Note that a table has been deleted between disovery and file open
+  @param[in]  path  missing table name with path. */
+  void add_missing_table(std::string path);
+  /** Check if table is in missing list
+  @param[in]  name  tablespace name */
+  bool is_missing_table(const std::string &name);
+  std::string convert_file_name(space_id_t space_id, std::string file_name,
+                                std::string ext);
```
satya-bodapati: missing function comment

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+ * This map is later used to delete the .delta and .meta file for a dropped
+ * tablespace (ie. when processing the .del entries in reduced lock)
+ * @return true on success
+ */
+bool xtrabackup_scan_meta(const datadir_entry_t &entry, /*!<in: datadir entry */
+                          void * /*data*/);
+
+bool prepare_handle_new_files(
+    const datadir_entry_t &entry, /*!<in: datadir entry */
+    void *arg __attribute__((unused)));
```
satya-bodapati, line 223: missing function comment

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Handle DDL for deleted files
+ * example input: test/10.del file
+ *        result: tablespace with space_id=10 will be deleted
+ * @return true on success
+ */
+bool prepare_handle_del_files(
+    const datadir_entry_t &entry, /*!<in: datadir entry */
+    void *arg __attribute__((unused)));
+
```
satya-bodapati: improper function comments, parameter comments.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1558,6 +1576,18 @@ Disable with --skip-innodb-checksums.",
      "Debug sync point. This is only used by the xtrabackup test suite",
      (G_PTR *)&xtrabackup_debug_sync, (G_PTR *)&xtrabackup_debug_sync, 0,
      GET_STR, REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
+
+#ifdef UNIV_DEBUG
```
satya-bodapati: lets use ifndef NDEBUG (same as below)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2514,6 +2636,22 @@ static bool xtrabackup_read_info(char *filename) {
   } else if (xb_server_version < 80029) {
     cfg_version = IB_EXPORT_CFG_VERSION_V6;
   }
+  /* skip start_time, end_time, lock_time, binlog_pos, innodb_from_lsn,
+   * innodb_to_lsn, partial, incremental, format, compressed, encrypt */
+  for (int i = 0; i < 12; i++) {
```
satya-bodapati, line 2654: I wish a better way of skipping parameter without the hardcoded 12. Not very important. Just a note.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2752,6 +2890,30 @@ static bool xtrabackup_write_info(const char *filepath) {
...
+  if (type == "ON") {
+    return LOCK_DDL_ON;
+  } else if (type == "OFF") {
+    return LOCK_DDL_OFF;
+  } else if (type == "REDUCED") {
+    return LOCK_DDL_REDUCED;
+  } else {
+    ut_error;
+  }
+}
```
satya-bodapati: what if there is TRUE and FALSE. This is allowed too. Please update this.
What about case comparison?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -2998,6 +3160,11 @@ const char *xb_get_copy_action(const char *dflt) {
 /* TODO: We may tune the behavior (e.g. by fil_aio)*/
 
 static bool xtrabackup_copy_datafile(fil_node_t *node, uint thread_n) {
+  return xtrabackup_copy_datafile(node, thread_n, nullptr);
```
satya-bodapati: lets call the overloaded function by idfferent name. may be add _func or _low().

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4676,7 +4875,7 @@ static bool get_meta_path(
 {
   size_t len = strlen(delta_path);
 
-  if (len <= 6 || strcmp(delta_path + len - 6, ".delta")) {
+  if (len <= 6 || strcmp(delta_path + len - 6, EXT_DELTA.c_str())) {
```
satya-bodapati: possible to avoid 6 ? strlen(EXT_DELTA). Simlar comment to other EXT variants introduced in this code.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -5049,7 +5248,7 @@ static pfs_os_file_t xb_delta_open_matching_space(
       ut_a(res);
 
       xb::info() << "Renaming " << dest_space_name << " to " << tmpname
-                 << ".ibd";
+                 << EXT_IBU;
```
satya-bodapati: This should be EXT_IBD. Refactoring/cleanup mistake

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6498,6 +6698,78 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
 
   Tablespace_map::instance().deserialize("./");
 
+  if (opt_lock_ddl == LOCK_DDL_REDUCED) {
+    if (!xb_process_datadir(
+            xtrabackup_incremental_dir ? xtrabackup_incremental_dir : ".",
+            ".corrupt", prepare_handle_corrupt_files, NULL)) {
```
satya-bodapati: new extention for ".CORRUPT". May be introduced a short name ".CRPT" ?

storage/innobase/xtrabackup/test/bootstrap.sh
```diff
@@ -111,7 +111,7 @@ main () {
 
     case "${TYPE}" in
         innodb80)
-            url="https://dev.mysql.com/get/Downloads/MySQL-8.1"
+            url="https://dev.mysql.com/get/Downloads/MySQL-8.2"
```
satya-bodapati: 8.4

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
```diff
@@ -80,7 +80,7 @@ run_cmd_expect_failure $XB_BIN $XB_ARGS --backup --slave-info --no-lock \
   --target-dir=$topdir/backup
 
 vlog "Full backup of the slave server"
-xtrabackup --backup --lock-ddl=false --target-dir=$topdir/backup --slave-info --safe-slave-backup 2>&1 | tee $topdir/pxb.log
+xtrabackup --backup --lock-ddl=OFF --target-dir=$topdir/backup --slave-info --safe-slave-backup 2>&1 | tee $topdir/pxb.log
```
satya-bodapati: Lets revert these changes from =OFF to =false. It is a valid value.
May be leave some tests with the OFF.

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
```diff
@@ -196,7 +202,7 @@ start_server_with_id $slave3_id
 setup_slave $slave3_id $master_id
 
 vlog "Full backup of the GTID slave server"
-xtrabackup --backup --lock-ddl=false --slave-info --target-dir=$topdir/backup
+xtrabackup --backup --lock-ddl=OFF --slave-info --target-dir=$topdir/backup
```
satya-bodapati: Lets keep OFF here

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
```diff
@@ -165,7 +171,7 @@ setup_slave GTID $slave2_id $master_id
 mysql -e "SET GLOBAL general_log=1; SET GLOBAL log_output='TABLE';"
 
 vlog "Full backup of the GTID with AUTO_POSITION slave server"
-xtrabackup --backup --lock-ddl=false --slave-info --target-dir=$topdir/backup
+xtrabackup --backup --lock-ddl=OFF --slave-info --target-dir=$topdir/backup
```
satya-bodapati, line 174: Lets kep OFF here. (Everwhere else, we can use false)

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+  }
+  size_t len = fread(content, 1, FN_REFLEN, f);
+  fclose(f);
+  return std::string(content, len);
+}
+
+static void delete_force(const std::string &path) {
+  if (access(path.c_str(), R_OK) == 0) {
+    if (my_delete(path.c_str(), MYF(MY_WME))) {
+      exit(EXIT_FAILURE);
```
satya-bodapati: I think MYF_WME ensure we log the reason for failure on stderr. But lets be safe and write a xb:error() cannot delete.. and error code from strerr(errno)?

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1041 @@
...
+    backup_file_printf(
+        convert_file_name(elem.second, elem.first, EXT_DEL).c_str(), "%s", "");
+  }
+
+  // Add new undo files to be recopied
+  for (auto &elem : new_undo_files) {
+    new_tables[elem.second] = elem.first;
+  }
+
+  if (new_tables.empty()) return;
```
satya-bodapati: Lets indicate that there are no files are being copied by reduced lock. xb::info() msg please

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
aybek: AB

storage/innobase/fil/fil0fil.cc
aybek: AB

storage/innobase/include/fil0fil.h
aybek: AB

storage/innobase/fil/fil0fil.cc
aybek: AB

storage/innobase/fil/fil0fil.cc
aybek: AB

storage/innobase/include/fil0fil.h
aybek: AB

storage/innobase/include/fil0fil.h
aybek: AB

storage/innobase/log/log0recv.cc
aybek: AB

storage/innobase/xtrabackup/src/backup_mysql.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.h
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/test/bootstrap.sh
aybek: AB

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
aybek: AB

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: AB
I think its better to leave if comparisons because switch type is string

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/fil/fil0fil.cc
aybek: if we do that Fil_system::open_for_recovery() will fail with assertion "ut_ad(recv_recovery_is_on() || Log_DDL::is_in_recovery());"

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.h
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.h
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.h
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: I have added rename error handling logic in prepare_handle_ren_files() func. 
Here https://github.com/percona/percona-xtrabackup/pull/1603/files/65300ae9ee89262ea5de2e2f2ebff8737bc74f03#diff-0647c6584a748ffd0ebb9e7b4957e7ddbdcdfca51932a0c38c5960e5d7ec001eR825

storage/innobase/fsp/fsp0file.cc
aybek: Do you mean we should readd some code that was mistakenly removed and put it under #ifdef XTRABACKUP ? 
Not sure which part of the code you mean

storage/innobase/xtrabackup/test/suites/binlog/ib_slave_info.sh
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: AB

storage/innobase/xtrabackup/src/ddl_tracker.h
aybek: I think you can explain this better than me 😃

storage/innobase/fil/fil0fil.cc
aybek: AB

storage/innobase/fil/fil0fil.cc
satya-bodapati: The assertion should be handled, or lock_ddl == reduced

storage/innobase/xtrabackup/src/ddl_tracker.h
satya-bodapati: Yes, I will address this.

storage/innobase/fil/fil0fil.cc
satya-bodapati: Looks like some other problems exist. Dropping this comment

storage/innobase/fsp/fsp0file.cc
satya-bodapati: I just meant, using 
#ifdef XTRBACKUP
  xb_variant
#else
 upstream_variant
#endif

storage/innobase/fil/fil0fil.cc
```diff
@@ -519,8 +519,14 @@ class Tablespace_dirs {
   void set_scan_dirs(const std::string &directories);
 
   /** Discover tablespaces by reading the header from .ibd files.
+  @param[in] populate_fil_cache   if true, tabelspace are loaded to cache
+  @param[in] is_prep_handled_ddls if true, xtrabackup uses
+                                  ibd_open_for_recovery instead of
+                                  fil_open_for_xtrabackup()
```
satya-bodapati: remove

storage/innobase/fil/fil0fil.cc
```diff
@@ -1645,6 +1650,11 @@ class Fil_system {
   @return DB_SUCCESS if the open was successful */
   [[nodiscard]] dberr_t open_for_recovery(space_id_t space_id);
 
+  /** Open a tablespace when LOCK_DDL_REDUCED is on.
+  @param[in]  path    File path
```
satya-bodapati: Add comment, why we cannot use fil_open_for_xtrabackup() here

storage/innobase/fil/fil0fil.cc
```diff
@@ -1700,9 +1710,16 @@ class Fil_system {
...
 
   /** Scan the directories to build the tablespace ID to file name
-  mapping table. */
-  dberr_t scan(bool populate_fil_cache) {
-    return (m_dirs.scan(populate_fil_cache));
+  mapping table
+  @param[in] populate_fil_cache   if true, tabelspace are loaded to cache
+  @param[in] is_prep_handled_ddls if true, xtrabackup uses
+                                  ibd_open_for_recovery instead of
+                                  fil_open_for_xtrabackup()
```
satya-bodapati: remove

storage/innobase/fil/fil0fil.cc
```diff
@@ -2704,7 +2720,14 @@ dberr_t Fil_shard::get_file_size(fil_node_t *file, bool read_only_mode) {
     }
   }
 
+  /* space_id mismatch can happen for truncation of tablespaces. Regular and
+   * Undo */
```
satya-bodapati: formatting

storage/innobase/fil/fil0fil.cc
```diff
@@ -11463,6 +11565,14 @@ std::tuple<dberr_t, space_id_t> fil_open_for_xtrabackup(
   err = file.validate_first_page(SPACE_UNKNOWN, &flush_lsn, false);
 
   space_id_t space_id = file.space_id();
+
+  /* Note that table has been renamed during scan, we will skip
+  opening renamed table since original table was loaded to cache
+  and copy/rename will be handled by ddl_tracker */
+  if (ddl_tracker && err == DB_TABLESPACE_EXISTS) {
+    ddl_tracker->add_renamed_table(space_id, path);
```
satya-bodapati: may be rename to add_rename_ibd_scan()

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,210 @@
...
+  @return DB_SUCCESS for success, others for errors */
+  dberr_t handle_ddl_operations();
+
+  /** Note that a table has been deleted between disovery and file open
+  @param[in]  path  missing table name with path. */
+  void add_missing_table(std::string path);
+
+  /** Check if table is in missing list
+  @param[in]  name  tablespace name */
+  bool is_missing_table(const std::string &name);
```
satya-bodapati: is_missing_after_discovery

storage/innobase/xtrabackup/src/ddl_tracker.h
```diff
@@ -0,0 +1,210 @@
...
+  void backup_file_op(uint32_t space_id, mlog_id_t type, const byte *buf,
+                      ulint len, lsn_t start_lsn);
+
+  /** Handle DDL operations that happenned in reduced lock mode
+  @return DB_SUCCESS for success, others for errors */
+  dberr_t handle_ddl_operations();
+
+  /** Note that a table has been deleted between disovery and file open
+  @param[in]  path  missing table name with path. */
+  void add_missing_table(std::string path);
```
satya-bodapati: add_missing_after_discovery()

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1108 @@
...
+#include "srv0start.h"
+#include "utils.h"
+#include "xb0xb.h"       // check_if_skip_table
+#include "xtrabackup.h"  // datafiles_iter_t
+
+extern bool xb_read_delta_metadata(const char *filepath, xb_delta_info_t *info);
+
+void ddl_tracker_t::backup_file_op(uint32_t space_id, mlog_id_t type,
+                                   const byte *buf, ulint len,
+                                   lsn_t start_lsn) {
```
satya-bodapati: rename to record_lsn, here and everywhere

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -0,0 +1,1108 @@
...
+    os_thread_create(PFS_NOT_INSTRUMENTED, i, copy_for_reduced,
+                     data_threads + i)
+        .start();
+  }
+
+  /* Wait for threads to exit */
+  while (1) {
+    xb::info() << "Sleeping for 1 second, waiting for the recopy of tables to "
+                  "be finished."
+               << "Expecting count be 0, current count is " << count;
```
satya-bodapati: space between finished. and Expecting

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
```diff
@@ -1700,9 +1710,13 @@ class Fil_system {
   }
 
   /** Scan the directories to build the tablespace ID to file name
-  mapping table. */
-  dberr_t scan(bool populate_fil_cache) {
-    return (m_dirs.scan(populate_fil_cache));
+  mapping table
+  @param[in] populate_fil_cache   if true, tabelspace are loaded to cache
```
satya-bodapati: tablespace*

storage/innobase/fil/fil0fil.cc
```diff
@@ -11543,15 +11662,38 @@ void Tablespace_dirs::open_ibd(const Const_iter &start, const Const_iter &end,
...
-    is unused here and we get unused variable error during compilation */
-    dberr_t err;
-    std::tie(err, std::ignore) = fil_open_for_xtrabackup(
-        phy_filename, filename.substr(0, filename.length() - 4));
-    /* PXB-2275 - Allow DB_INVALID_ENCRYPTION_META as we will test it in the
-    end of the backup */
-    if (err != DB_SUCCESS && err != DB_INVALID_ENCRYPTION_META) {
-      result = false;
+    /* when LOCK_DDL_REDUCED while processing ddl files on prepare phase
+    we should load data files without first page validation */
```
satya-bodapati: /* With lock-ddl=reduced, prepare on such backup directories, we open tablespaces before recovery starts. This is required to process to space_id.del or .ren files. Hence, we have to
tolerate encryption errors during first-page validation. Hence, we use a separate opening function of reduced mode */

storage/innobase/os/os0enc.cc
```diff
@@ -664,7 +664,9 @@ bool Encryption::decode_encryption_info(space_id_t space_id,
     encryption info maybe hasn't written into datafile when
     the table is newly created. For clone encryption information
     should have been already correct. */
-    return (recv_recovery_is_on() ? true : false);
+    bool prep_lock_ddl_reduced =
```
satya-bodapati, line 671: /* With reduced lock ddl, we open just for space_id and filename mapping. Tolerate missing encryption information */

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3182,6 +3360,11 @@ static void data_copy_thread_func(data_thread_ctxt_t *ctxt) {
     if (xtrabackup_copy_datafile(node, num)) {
       xb::error() << "failed to copy datafile " << node->name;
       *(ctxt->error) = true;
+    } else {
+      if (ddl_tracker && fsp_is_undo_tablespace(node->space->id)) {
+        ddl_tracker->add_undo_tablespace(node->space->id,
+                                         node->space->files.front().name);
+      }
```
satya-bodapati: Add a comment that, instead of tracking at the beginning, we will do the tracking here after xtrabackup copies of the undo files.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7445,6 +7719,12 @@ int main(int argc, char **argv) {
     exit(EXIT_FAILURE);
   }
 
+  if (opt_page_tracking && opt_lock_ddl == LOCK_DDL_REDUCED) {
+    xb::error() << "--page-tracking and --lock-ddl=REDUCED cannot be enabled "
+                   "together.";
```
satya-bodapati: fix formatting

storage/innobase/xtrabackup/src/xtrabackup.h
```diff
@@ -237,15 +240,22 @@ enum binlog_info_enum {
 void xtrabackup_io_throttling(void);
 bool xb_write_delta_metadata(const char *filename, const xb_delta_info_t *info);
 
+std::string ddl_lock_type_to_str(lock_ddl_type_t type);
+
+lock_ddl_type_t ddl_lock_type_from_str(std::string type);
```
satya-bodapati: function comments

storage/innobase/xtrabackup/src/xtrabackup.h
```diff
@@ -237,15 +240,22 @@ enum binlog_info_enum {
...
+
 datafiles_iter_t *datafiles_iter_new(
     const std::shared_ptr<const xb::backup::dd_space_ids>);
 fil_node_t *datafiles_iter_next(datafiles_iter_t *it);
 void datafiles_iter_free(datafiles_iter_t *it);
 
+bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
+                                   const char *dest_name);
+
```
satya-bodapati, line 264: function comment. We are exposing this function now.

storage/innobase/xtrabackup/src/xtrabackup.h
```diff
@@ -332,4 +342,6 @@ bool xb_process_datadir(const char *path,   /*!<in: datadir path */
 @param[in,out]	buf		log header buffer
 @param[in]	lsn		lsn to update */
 void update_log_temp_checkpoint(byte *buf, lsn_t lsn);
+
+void xb_scan_for_tablespaces(bool only_undo);
```
satya-bodapati, line 359: function comment. we are exposing this now

storage/innobase/xtrabackup/test/bootstrap.sh
```diff
@@ -111,7 +111,7 @@ main () {
 
     case "${TYPE}" in
         innodb80)
-            url="https://dev.mysql.com/get/Downloads/MySQL-8.1"
+            url="https://dev.mysql.com/get/Downloads/MySQL-8.4"
```
satya-bodapati, line 114: lets fix the PS 8.4 URL too

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -306,9 +306,16 @@ static void copy_for_reduced(copy_thread_ctxt_t *ctxt) {
     if (ctxt->new_tables->find(node->space->id) == ctxt->new_tables->end()) {
       continue;
     }
-    std::string dest_name = node->name;
-    dest_name.append(EXT_NEW);
-    if (xtrabackup_copy_datafile_func(node, num, dest_name.c_str())) {
+    char dest_name[FN_REFLEN];
+    if (fsp_is_file_per_table(node->space->id, node->space->flags)) {
+      snprintf(dest_name, sizeof(dest_name), "%s%s%s",
+               fil_path_to_space_name(node->name), EXT_IBD.c_str(),
```
satya-bodapati: Let's use a single place to convert a source filename to a destination filename.


Please move logic fil_path_to_space_name().

May be also also pass node->space->id or the entire node struct. use the space_id to determine if it is undo or not.

mysql> CREATE UNDO TABLESPACE UNDO_25 ADD DATAFILE 'blah_25.ibu';
Query OK, 0 rows affected (0.09 sec)

mysql> CREATE TABLESPACE ts1;
Query OK, 0 rows affected (0.01 sec)

datadir contents:
ls -l var/mysqld.1/data/
total 140612
-rw-rw---- 1 satya satya       56 Oct 10 11:02  auto.cnf
-rw-rw---- 1 satya satya   198343 Oct 10 11:03  binlog.000001
-rw-rw---- 1 satya satya       16 Oct 10 11:02  binlog.index
-rw-rw---- 1 satya satya 16777216 Oct 10 11:04  blah_25.ibu
-rw-rw---- 1 satya satya 16777216 Oct 10 11:03  bloh_25.ibu
-rw-rw---- 1 satya satya   114688 Oct 10 11:03  f3914a36-86ee-11ef-9c27-047bcbcb6b7e.ibd

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ddl_tracker.cc
aybek: done

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -288,6 +288,36 @@ typedef struct {
...
+ * This function extracts the database name and space name from the
+ * tablepsace path. If the space is not file-per-table tablespace, it
+ * returns only the space name with appended extension. For other cases,
+ * it constructs a path in the format "database_name/space_name"
+ * with an appended extension.
+ *
+ * @param node File node of a tablespace
+ * @return A string representing the destination name
+ */
+std::string get_dest_name(fil_node_t *node) {
```
satya-bodapati: we can make parameter const

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -288,6 +288,36 @@ typedef struct {
...
+    return space_name + EXT_NEW;
+  }
+
+  // Extract the last part of db_name after the last separator
+  last_separator_pos = db_name.find_last_of(Fil_path::SEPARATOR);
+  if (last_separator_pos != std::string::npos) {
+    db_name = db_name.substr(last_separator_pos + 1);
+  }
+
+  return db_name + '/' + space_name + EXT_NEW;
```
satya-bodapati: Use Fil_path::SEPARATOR instead of '/'

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -306,16 +336,12 @@ static void copy_for_reduced(copy_thread_ctxt_t *ctxt) {
...
-      snprintf(dest_name, sizeof(dest_name), "%s%s",
-               Fil_path::get_basename(node->name).c_str(), EXT_NEW.c_str());
-    }
-    if (xtrabackup_copy_datafile_func(node, num, dest_name)) {
+    // Get destination name for the datafile
+    std::string dest_name = get_dest_name(node);
+    char dest_name_buffer[FN_REFLEN];
+    snprintf(dest_name_buffer, FN_REFLEN, "%s", dest_name.c_str());
+
+    if (xtrabackup_copy_datafile_func(node, num, dest_name_buffer)) {
```
satya-bodapati: simply use dest_name.c_str()

satya-bodapati (changes requested):

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -288,6 +288,36 @@ typedef struct {
...
+ * @return A string representing the destination name
+ */
+std::string get_dest_name(fil_node_t *node) {
+  std::string path{node->name};
+  auto last_separator_pos = path.find_last_of(Fil_path::SEPARATOR);
+
+  std::string db_name = path.substr(0, last_separator_pos);
+  std::string space_name = path.substr(last_separator_pos + 1);
+
+  if (!fsp_is_file_per_table(node->space->id, node->space->flags)) {
```
satya-bodapati: add debug assertion here that space_name shouldn't be empty/(not size 0)

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -288,6 +288,36 @@ typedef struct {
...
+    return space_name + EXT_NEW;
+  }
+
+  // Extract the last part of db_name after the last separator
+  last_separator_pos = db_name.find_last_of(Fil_path::SEPARATOR);
+  if (last_separator_pos != std::string::npos) {
+    db_name = db_name.substr(last_separator_pos + 1);
+  }
+
```
satya-bodapati, line 338: also the same for db_name.

add debug assertion here that db_name shouldn't be empty/(not size 0)

satya-bodapati (changes requested):

satya-bodapati:
A newer squashed version will be submitted.

PR #1611, opened 2024-09-17
Fix authentication for Azure API in xbcloud

imrivera:
There was a problem when an Azure key contained a null byte, the result when decoding base64 was trimmed and part of the key was lost.

Also if the decoded binary key started or ended with white space, some bytes would be discarded.

This PR fixes both problems

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
@@ -558,15 +558,15 @@ TEST(azure_signer, basicDNS) {
   req.append_payload("test", 4);
   Azure_signer signer(
       "my-storage-account",
-      "zUfvsKXc6+2RMJCwvnElnG/"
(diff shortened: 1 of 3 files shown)


storage/innobase/xtrabackup/src/xbcloud/xbcloud-t.cc
```diff
@@ -558,15 +558,15 @@ TEST(azure_signer, basicDNS) {
   req.append_payload("test", 4);
   Azure_signer signer(
       "my-storage-account",
-      "zUfvsKXc6+2RMJCwvnElnG/"
-      "Kk7wxQ8V4TPXIuZ53qFwJNtpLUEjYdBe9iGTkMgwUGFHVfFgn2qkgoqDP/b3OAg==",
+      "zUfvsKXcAO2RMJCwvnElnG/"
```
imrivera, line 561: Now this key contains a NULL byte when decoded, and it ends in `\r` to check that the request signing is correctly done

PR #1641, opened 2025-03-11
(trunk) PXB-3279 Remove requirement to have open file handles equal to files …

aybek:
…in datadir


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3210,6 +3212,9 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
   res = xb_fil_cur_open(&cursor, read_filter, node, thread_n);
   if (res == XB_FIL_CUR_SKIP) {
     goto skip;
+  } else if (res == XB_FIL_CUR_MISSING) {
+    ddl_tracker->add_missing_after_discovery(cursor.space_id);
+    goto error;
```
satya-bodapati: shouldn't it be continue here? if a file is missing, we should continue

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3248,9 +3253,10 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
   action = xb_get_copy_action();
 
   if (xtrabackup_stream) {
-    xb::info() << action << " " << node_path;
+    xb::info() << action << " " << node->space->id << " " << node_path;
   } else {
-    xb::info() << action << " " << node_path << " to " << dstfile->path;
+    xb::info() << action << " " << node->space->id << " " << node_path << " to "
```
satya-bodapati: May be it would be better something like this: action <filename> with space_id  10;

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3273,10 +3279,11 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
...
   /* close */
   if (xtrabackup_stream) {
-    xb::info() << "Done: " << action << " " << node_path;
+    xb::info() << "Done: " << action << " " << node->space->id << " "
+               << node_path;
   } else {
-    xb::info() << "Done: " << action << " " << node_path << " to "
-               << dstfile->path;
+    xb::info() << "Done: " << action << " " << node->space->id << " "
+               << node_path << " to " << dstfile->path;
```
satya-bodapati: same as above

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3366,15 +3373,29 @@ static void data_copy_thread_func(data_thread_ctxt_t *ctxt) {
...
     } else {
-      /* With reduced lock mode, instead of tracking undo from the startup scan,
-      we track undo tablespaces after we copy them */
-      if (ddl_tracker && fsp_is_undo_tablespace(node->space->id)) {
-        ddl_tracker->add_undo_tablespace(node->space->id,
-                                         node->space->files.front().name);
+      // failure
+      if (ddl_tracker != nullptr && opt_lock_ddl == LOCK_DDL_REDUCED) {
+        // ingore failure and continue, we should ideally track these..
+        continue;
```
satya-bodapati: Please treat this as a failure and abort backup. I think XB_FIL_CUR_MISSING is designed to handle all known issues with copying files during reduced lock.

storage/innobase/xtrabackup/test/suites/reducedlock/limit_file_handles.sh
```diff
@@ -0,0 +1,74 @@
...
+# PXB-3279: Remove requirement to have open file handles equal to files in datadir
+###############################################################################
+
+. inc/common.sh
+
+require_debug_pxb_version
+require_pro_pxb_version
+
+open_file_limit=$(ulimit -n)
+ulimit -n 1024
```
satya-bodapati, line 11: Did you run the test on Jenkins? The user might not have permission to execute this.

storage/innobase/xtrabackup/test/suites/reducedlock/limit_file_handles.sh
```diff
@@ -0,0 +1,74 @@
...
+
+for i in {1..5000} ; do
+    $MYSQL $MYSQL_ARGS -Ns -e "CREATE TABLE test.tb_${i} (id INT PRIMARY KEY AUTO_INCREMENT); INSERT INTO test.tb_${i} VALUES (1)" test
+done;
+
+innodb_wait_for_flush_all
+
+xtrabackup --backup --target-dir=$topdir/backup \
+  --debug-sync="xtrabackup_suspend_between_file_discovery_and_open" --lock-ddl=REDUCED \
+  2> >( tee $topdir/backup.log)&
```
satya-bodapati, line 29: We should also test dropping and renaming files, after discovery & loading to cache. Can you please add this test scenarios?

satya-bodapati (changes requested):

storage/innobase/xtrabackup/test/suites/reducedlock/limit_file_handles.sh
aybek: do you mean after sync point `xtrabackup_load_tablespaces_pause` ?
I think this scenario is already covered, for instance in
/mnt/nvme0_xfs/dev_aibekb/pxb/backup_pxb3279/storage/innobase/xtrabackup/test/suites/reducedlock/basic_operation.sh

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -145,16 +145,14 @@ void ddl_tracker_t::add_to_recopy_tables(space_id_t space_id, lsn_t record_lsn,
...
-  Fil_path::normalize(path);
-  if (Fil_path::has_prefix(path, Fil_path::DOT_SLASH)) {
-    path.erase(0, strlen(Fil_path::DOT_SLASH));
-  }
-
   std::lock_guard<std::mutex> lock(m_ddl_tracker_mutex);
-  missing_after_discovery.insert(path);
+  missing_after_discovery.insert(space_id);
+
+  xb::info() << " missing space ID: " << space_id;
```
satya-bodapati: Let's add the same prefix DDL tracking: missing after discovery space_id :

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -593,6 +589,8 @@ dberr_t ddl_tracker_t::handle_ddl_operations() {
       continue;
     }
 
+    ulint flags = tables_copied_no_lock[space_id].second;
```
satya-bodapati: Lets not use the [] notation anymore. Please use the find() method.

storage/innobase/xtrabackup/src/fil_cur.cc
```diff
@@ -151,7 +151,7 @@ xb_fil_cur_result_t xb_fil_cur_open(
 
       xb::error() << "cannot open tablespace " << cursor->abs_path;
 
-      return (XB_FIL_CUR_ERROR);
+      return (XB_FIL_CUR_MISSING);
```
satya-bodapati: Should be we keep XB_FIL_CUR_ERROR here? because this path, the caller doesn't need to close the file.

May be better to simplify. Let the caller always call the close file function? I think it is OK. because we have stored the state FIL_CLOSED or OPENED)

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3210,6 +3212,9 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
   res = xb_fil_cur_open(&cursor, read_filter, node, thread_n);
   if (res == XB_FIL_CUR_SKIP) {
     goto skip;
+  } else if (res == XB_FIL_CUR_MISSING) {
+    ddl_tracker->add_missing_after_discovery(cursor.space_id);
+    goto skip;
```
satya-bodapati, line 3217: do we call this function under the second copy phase? (ie under the lock)?

After taking the lock, any error is a serious error, and we should abort the backup.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3248,9 +3253,11 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
   action = xb_get_copy_action();
 
   if (xtrabackup_stream) {
-    xb::info() << action << " " << node_path;
+    xb::info() << action << " with space_id " << node->space->id << " "
```
satya-bodapati: Please change to " file with space_id "

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3248,9 +3253,11 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
...
+               << node_path;
   } else {
-    xb::info() << action << " " << node_path << " to " << dstfile->path;
+    xb::info() << action << " with space_id " << node->space->id << " "
```
satya-bodapati: file with space_id

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3273,10 +3280,11 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
 
   /* close */
   if (xtrabackup_stream) {
-    xb::info() << "Done: " << action << " " << node_path;
+    xb::info() << "Done: " << action << " " << node->space->id << " "
```
satya-bodapati: fix here to; "file with space_id "

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3273,10 +3280,11 @@ bool xtrabackup_copy_datafile_func(fil_node_t *node, uint thread_n,
...
+               << node_path;
   } else {
-    xb::info() << "Done: " << action << " " << node_path << " to "
-               << dstfile->path;
+    xb::info() << "Done: " << action << " with space_id " << node->space->id
```
satya-bodapati: file with space_id

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -3366,16 +3374,25 @@ static void data_copy_thread_func(data_thread_ctxt_t *ctxt) {
...
+        /* With reduced lock mode, instead of tracking undo from the startup
+          scan, we track undo tablespaces after we copy them */
+        if (fsp_is_undo_tablespace(node->space->id)) {
+          ddl_tracker->add_undo_tablespace(node->space->id,
+                                           node->space->files.front().name);
+        } else if (fsp_is_ibd_tablespace(node->space->id)) {
+          ddl_tracker->add_table_from_ibd_scan(node->space->id,
+                                               node->space->files.front().name,
+                                               node->space->flags);
+        }
```
satya-bodapati: I would an else branch. this will be unhandled path? On release build, print message about space->id, name and return error.

On debug build, let it crash ut_ad(0);

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4083,80 +4100,6 @@ static uint xb_set_max_open_files(
 #endif
 }
 
-/* true on success, false on failure */
-bool xb_check_and_set_open_files_limit(size_t num_files) {
```
satya-bodapati, line 4087: can you make this as separate commit, remove "commit id"? use `git revert`

storage/innobase/xtrabackup/test/suites/reducedlock/limit_file_handles.sh
satya-bodapati: For now keep it, we will see when jenkins works.

storage/innobase/xtrabackup/test/suites/reducedlock/limit_file_handles.sh
satya-bodapati: Can you please ensure that the file opening fails from your new code? XB_FIL_CUR_MISSING is returned.

Also do we have testcase for space_id mismatch? This is to ensure the space_id mismatch path is activated/used.

storage/innobase/xtrabackup/test/suites/reducedlock/rename_problem.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+#  stop_server
+#
+#  rm -rf $mysql_datadir/*
+#  xtrabackup --copy-back --target-dir=$topdir/backup_enc_general_tablespace --xtrabackup-plugin-dir=${plugin_dir}
+#
+#  start_server
+#  verify_db_state test
+#  stop_server
+#  rm -rf $mysql_datadir $topdir/backup_with_enc_general_tablespace.log $topdir/backup_enc_general_tablespace $topdir/backup_inc
```
satya-bodapati: I think these parts should be activated. Wihtout the patch, the prepare would crash. But since we have a fix, we have to restore, start_server, etc.

satya-bodapati (changes requested):

satya-bodapati:
@aybek can you also please rebase the patch to 8.4 and also create a trunk version of it?

storage/innobase/xtrabackup/src/fil_cur.cc
aybek: Do you think we need to keep the XB_FIL_CUR_ERROR here ?

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: Do you mean we need to crash in `else case` (not fsp_is_undo_tablespace and not fsp_is_ibd_tablespace) ?

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: I will check this

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: yes, we should hanlde all cases So am talking about the else branch
if () {
} else if () {
} else {
--> unexpected path, so crash here on debug builds and  print warning for releae builds
}

aybek: left this as is, in else case we have all other tablespaces like system/tmp etc

storage/innobase/xtrabackup/src/xtrabackup.cc
aybek: commited separately but without revert, all reducedlock feature commits were squashed during merge

storage/innobase/xtrabackup/src/ddl_tracker.cc
```diff
@@ -548,6 +556,12 @@ dberr_t ddl_tracker_t::handle_ddl_operations() {
     }
   }
 
+  for (auto &table : missing_after_discovery) {
```
satya-bodapati: @aybek Do you know why need this? We will not add any table which is missing to tables_copied_no_lock. isn't it? :thinking:

satya-bodapati: Fixed by the new commit

satya-bodapati:
#1664 handles this

PR #1650, opened 2025-04-10
PXB-3498: [xbcloud] Fix HTTP request signing when retrying

imrivera:
Using `xbcloud`, when there is a problem downloading a file and it's going to be retried, the new request should be signed again for S3 and Azure. But the parameters used were only valid for S3.

That means that if you're downloading from Azure, once a file fails to download, the retry is going to miscalculate the authorization token, that would receive a `403 Unauthorized` and the whole downloading process stop.

This PR fixes that problem. The approach is encapsulating the signing process and the related data in a `std::function` in the `Http_request` object. This way, you just have to call `req->sign()` to resign the request, you don't need to know what data it needs, etc.

**This bug was found in a production service**. The fix was also tested in the same service with successful result.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/http.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/http.cc
@@ -634,8 +634,7 @@ void Http_client::callback(CLIENT *client, std::string container,
            delay, name.c_str(), count);
     std::this_thread::sleep_for(std::chrono::milliseconds(delay));
     resp->reset_body();
-    client->signer->sign_request(client->hostname(container), container, *req,
-                                 time(0));
+    req->sign();
     http_client->make_async_request(
         *req, *resp, h,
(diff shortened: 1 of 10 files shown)


storage/innobase/xtrabackup/src/xbcloud/http.cc
```diff
@@ -634,8 +634,7 @@ void Http_client::callback(CLIENT *client, std::string container,
            delay, name.c_str(), count);
     std::this_thread::sleep_for(std::chrono::milliseconds(delay));
     resp->reset_body();
-    client->signer->sign_request(client->hostname(container), container, *req,
```
imrivera, line 637: This is the main problem. These parameters are correct for S3, but not for Azure (for Azure they should be `container` and `name`)

satya-bodapati:
Thank you for the contribution: I've created JIRA for this issue: https://perconadev.atlassian.net/browse/PXB-3498

PR #1651, opened 2025-04-16
(8.4) PXB-3279 Remove requirement to have open file handles equal to files

aybek:


satya-bodapati:
This is tracked as #1663

PR #1712, opened 2025-12-12
[8.4] PXB-3624: Fix HTTP connection uninitialization order

jakub-nowakowski-percona:
https://perconadev.atlassian.net/browse/PXB-3624

Problem
=======

Wrong uninitialization order resulted in libcurl connections using libev event loops that have already been destroyed.

We discovered that libcurl versions affected by CVE-2025-0665 cause a crash in xbcloud.

Fix
===

Uninitialization order of libcurl multi handles and libev event loops has been fixed.

Support for affected versions of libcurl has been dropped, and a warning is now displayed when xbcloud is run with an unsupported version of libcurl.


jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/132/

satya-bodapati (approved):

PR #1716, opened 2025-12-19
[8.4] PXB-3458 : Ignore invalid innodb_undo_directory setting in --copy-back mode

jakub-nowakowski-percona:
https://perconadev.atlassian.net/browse/PXB-3458

Problem
-------
When backup-my.cnf is used in --copy-back mode, it sets innodb_undo_directory setting to "./". This doesn't make sense in this mode, and results in a backup restoration error.

Fix
---
We ignore innodb_undo_directory setting if it's set to "./" in --copy-back mode.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -82,7 +82,6 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <version_check_pl.h>
 #endif
 
-
 /** Possible values for system variable "innodb_checksum_algorithm". */
 extern const char *innodb_checksum_algorithm_names[];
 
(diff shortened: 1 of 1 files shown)


jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/135/

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -2290,7 +2290,10 @@ bool copy_back(int argc, char **argv) {
 
   /* copy undo tablespaces */
   if (srv_undo_tablespaces > 0) {
-    dst_dir = (srv_undo_dir && *srv_undo_dir) ? srv_undo_dir : mysql_data_home;
+    dst_dir = mysql_data_home;
+    if (srv_undo_dir && *srv_undo_dir && strcmp(srv_undo_dir, "./") != 0) {
```
satya-bodapati, line 2298: same as 8.0, a code comment would be nice to have.

satya-bodapati (approved):

PR #1717, opened 2025-12-19
Using HeadBucket for backups is dangerous in multi-tenant S3 buckets

### LINKED ISSUE #2251: Using HeadBucket for backups is dangerous in multi-tenant S3 buckets
### Proposal

This operator currently uses HeadBucket in order to check whether the provided S3 permissions are adequate for PITR/backup storage (which is not an extensive check since uploading objects will require more permissions).

The problem with this is that in IAM, there is literally no way to let a user call HeadBucket while also preventing them from listing objects inside that bucket. This is because HeadBucket requires the `s3:ListBucket` permission.

There are zero air-tight IAM workarounds that would prevent this permission from also allowing users to list objects inside the bucket. Here are some examples:
1. Grant s3:ListBucket conditionally on the "" prefix - does not allow HeadBucket because HeadBucket does not pass a prefix parameter, and so it is denied.
2. Grant s3:ListBucket conditionally on a specific directory - same as above.
3. Grant s3:ListBucket conditionally on `"Null": {"s3:prefix": "true"}` - this actually still allows listing all objects in the bucket, just not specifying a prefix.
4. Grant s3:ListBucket conditionally on everything except `*/*` - can still specify a prefix without a slash and list objects.

There has been extensive discussion around this issue as it relates to the CNPG plugin:
https://github.com/EnterpriseDB/barman/issues/929
https://github.com/EnterpriseDB/barman/issues/1101
https://github.com/hashicorp/terraform-provider-aws/issues/17195

The initial suggestion was to use ListObjectsV2 with a prefix rather than HeadBucket - this works in terms of security, but it is more expensive in terms of AWS API calls, and so was rejected.

The final solution as it relates to CNPG was to completely remove cloud connectivity + bucket existence checks before running backups. I propose the same for this operator.

### Use-Case

The use case is for multi-tenant S3 buckets. Using HeadBucket in this scenario is a non-starter because it would allow users to see backups of other users.

### Is this a feature you are interested in implementing yourself?

Maybe

### Anything else?

_No response_

DimitriosLisenko:
### Proposal

This project uses HeadBucket to determine the signature version to use, as well as whether the bucket exists before creating backups. The problem with this is that in IAM, there is literally no way to let a user call HeadBucket while also preventing them from listing objects inside that bucket. This is because HeadBucket requires the `s3:ListBucket` permission.

There are zero air-tight IAM workarounds that would prevent this permission from also allowing users to list objects inside the bucket. Here are some examples:
1. Grant s3:ListBucket conditionally on the "" prefix - does not allow HeadBucket because HeadBucket does not pass a prefix parameter, and so it is denied.
2. Grant s3:ListBucket conditionally on a specific directory - same as above.
3. Grant s3:ListBucket conditionally on `"Null": {"s3:prefix": "true"}` - this actually still allows listing all objects in the bucket, just not specifying a prefix.
4. Grant s3:ListBucket conditionally on everything except `*/*` - can still specify a prefix without a slash and list objects.

There has been extensive discussion around this issue as it relates to the CNPG plugin:
https://github.com/EnterpriseDB/barman/issues/929
https://github.com/EnterpriseDB/barman/issues/1101
https://github.com/hashicorp/terraform-provider-aws/issues/17195

The initial suggestion was to use ListObjectsV2 with a prefix rather than HeadBucket - this works in terms of security, but it is more expensive in terms of AWS API calls, and so was rejected.

The final solution as it relates to CNPG was to completely remove cloud connectivity + bucket existence checks before running backups. I propose the same here, which would require to remove the signature checks and require that it is provider; or instead to use HeadObject, which can be scoped to a directory.

### Use-Case

The use case is for multi-tenant S3 buckets. Using HeadBucket in this scenario is a non-starter because it would allow users to see backups of other users.

### Anything else?

This is a requirement to resolve https://github.com/percona/percona-xtradb-cluster-operator/issues/2251, because that operator calls HeadBucket in itself, as well as by virtue of calling xbcloud.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/s3.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/s3.cc
@@ -431,7 +431,7 @@ bool S3_client::create_bucket(const std::string &name) {
   return true;
 }
 
-bool S3_client::probe_api_version_and_lookup(const std::string &bucket) {
+bool S3_client::probe_api_version_and_lookup(const std::string &bucket, const std::string &backup) {
   for (auto lookup : {LOOKUP_DNS, LOOKUP_PATH}) {
     if (bucket_lookup != LOOKUP_AUTO && bucket_lookup != lookup) {
       continue;
@@ -455,7 +455,7 @@ bool S3_client::probe_api_version_and_lookup(const std::string &bucket) {
       api_version = version;
 
       bool exists;
-      if (bucket_exists(bucket.c_str(), exists)) {
+      if (object_exists(bucket.c_str(), backup.c_str(), exists)) {
         msg_ts("%s: Successfully connected.\n", my_progname);
         return true;
       }
@@ -479,6 +479,29 @@ bool S3_client::bucket_exists(const std::string &name, bool &exists) {
   Http_request req(Http_request::HEAD, protocol, hostname(name),
                    bucketname(name) + "/");
   signer->sign_request(hostname(name), name, req, time(0));
+
+  Http_response resp;
+  if (!http_client->make_request(req, resp)) {
+    return false;
+  }
+
+  if (resp.ok()) {
+    exists = true;
+    return true;
+  }
+
+  if (resp.http_code() == 404) {
+    exists = false;
+    return true;
+  }
+
+  return false;
+}
+
+bool S3_client::object_exists(const std::string &bucket, const std::string &name, bool &exists) {
+  Http_request req(Http_request::HEAD, protocol, hostname(bucket),
+                   bucketname(bucket) + "/" + name);
+  signer->sign_request(hostname(bucket), bucket, req, time(0));
 
   Http_response resp;
   if (!http_client->make_request(req, resp)) {
--- a/storage/innobase/xtrabackup/src/xbcloud/s3.h
+++ b/storage/innobase/xtrabackup/src/xbcloud/s3.h
@@ -227,13 +227,15 @@ class S3_client {
 
   void set_api_version(s3_api_version_t version) { api_version = version; }
 
-  bool probe_api_version_and_lookup(const std::string &bucket);
+  bool probe_api_version_and_lookup(const std::string &bucket, const std::string &backup);
 
   bool delete_object(const std::string &bucket, const std::string &name);
 
   bool create_bucket(const std::string &name);
 
   bool bucket_exists(const std::string &name, bool &exists);
+
+  bool object_exists(const std::string &bucket, const std::string &name, bool &exists);
 
   bool upload_object(const std::string &bucket, const std::string &name,
                      const Http_buffer &contents);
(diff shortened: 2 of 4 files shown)


DimitriosLisenko:
This PR is required for HeadBucket to be removed from the backup flow, and also for the operator to stop using HeadBucket. The operator work is currently at https://github.com/percona/percona-xtradb-cluster-operator/pull/2326 - are the same people able to review this item as well?

egegunes:
@satya-bodapati could you please take a look when you have the chance?

satya-bodapati:
Hi @DimitriosLisenko,

I tested this against a few IAM shapes against MinIO before pulling it into our 8.4 backport. Full reproduction and tests are on [PXB-3654](https://perconadev.atlassian.net/browse/PXB-3654?focusedCommentId=463356).

The first one is the push-only uploader: a user with `s3:ListBucket` and `s3:PutObject` only. This works today. After the patch, the new `HeadObject` probe asks for `s3:GetObject` on `bucket/<backup_name>`, gets `403 AccessDenied`, and xbcloud bails out with `Probe failed` before any chunk is uploaded.

The delete-only retention case is the same story. `s3:ListBucket` and `s3:DeleteObject` are enough today, `xbcloud delete` works. After the patch, no `GetObject` means the probe fails the same way.

The prefix-scoped tenant is more subtle. Their IAM grants `Get/Put/Delete` on `bucket/tenant1/*` only, and they happen to take backups named `tenant1`. The new probe sends `HeadObject` to `bucket/tenant1`, which does not match the IAM resource `bucket/tenant1/*` (the `*` requires at least one character after the slash), so the probe fails even though every actual chunk would land at `bucket/tenant1/<chunk>` and would succeed.

To be fair: the case your PR is built around, the multi-tenant tenant with prefix-conditioned `s3:ListBucket`, does start passing after the patch. The bug you set out to fix is real. It is just that the chosen fix trades one broken IAM pattern for three others.

I have a candidate fix in commit #4 of [PXB-3654-xbcloud-pr1717-regression-tests](https://github.com/satya-bodapati/percona-xtrabackup/tree/PXB-3654-xbcloud-pr1717-regression-tests) that keeps your `xbcloud_put` change (the bucket pre-check removal, which is the part that actually unblocks the multi-tenant case end-to-end) but reverts the probe to `HeadBucket` and makes it tolerant of `403 AccessDenied`. The reasoning is that the probe loop's only useful job is auto-detecting signature version and bucket lookup style; once the endpoint accepts our signature, the probe is done, and whether the caller has `ListBucket` should not be its business. With that change all four IAM shapes pass, including yours. The 403-tolerance bit is a heuristic and I am not 100% sure at the moment.

A couple of things I have not tested. GCS goes through the same `S3_object_store` code path so the fix should apply, but I did not point xbcloud at a real GCS bucket. Azure uses its own probe so neither your patch nor my other patch touches it. Your `xbcloud_put` change does affect Azure and Swift though, since the auto-create is in the common path; users on those backends who relied on xbcloud creating the container for them will need to pre-create it. Worth a changelog mention if it is not in there already.

[PXB-3654]: https://perconadev.atlassian.net/browse/PXB-3654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

satya-bodapati:
Another issue with removing the probe: https://perconadev.atlassian.net/browse/PXB-3654?focusedCommentId=463526

hulyav:
@DimitriosLisenko when do you think you could be addressing @satya-bodapati's review comments?

PR #1722, opened 2026-01-19
PXB-3543 [8.4] : Xtrabackup is extremely slow when performing incremenal backup for 1 Million Tables

satya-bodapati:


satya-bodapati:
trunk: #1723

jakub-nowakowski-percona (approved):

PR #1726, opened 2026-01-27
PXB-3643 [8.0] : Improve xbcloud delete to support Hierarchical Namespaces

satya-bodapati:
https://perconadev.atlassian.net/browse/PXB-3643

Problem:
--------
On Azure platform, buckets with HNS (Hierarchical Namespaces) enabled, xbcloud delete operation failed.

Analysis:
---------
When HNS is enabled on Azure, it is no longer a flatnamespace, meanining just deleting objects as files is not sufficient. Deleting a dir before deleting all the files in directory caused delete failure.

Refactor xbcloud delete to work with HNS-style directories by listing files and directories separately. Introduce a partitioned listing API with a default implementation and add Azure-specific directory detection using ResourceType. Refactor Azure listing parsing logic, and introduce a separate listin logic only for delete.

With this new listing logic, xbcloud will receive information about files and directories. xbcloud will delete all files first and then delete the directories.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/object_store.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/object_store.cc
@@ -1,0 +1,41 @@
+/******************************************************
+Copyright (c) 2026 Percona LLC and/or its affiliates.
+
+Object Store interface.
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; version 2 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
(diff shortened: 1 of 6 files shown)


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/387/

satya-bodapati:
8.4 version: #1727

storage/innobase/xtrabackup/src/xbcloud/object_store.cc
```diff
@@ -0,0 +1,45 @@
...
+ * @param directory Directory prefix to list.
+ * @param files Output list of file objects.
+ * @param dirs Output list of directory objects (unused here).
+ * @return true on success, false on error.
+ */
+bool Object_store::list_objects_files_and_dirs(const std::string &container,
+                                               const std::string &directory,
+                                               std::vector<std::string> &files,
+                                               std::vector<std::string> &dirs) {
+  if (!list_objects_in_directory(container, directory, files)) {
```
jakub-nowakowski-percona: Maybe this body could be just like this?
```
return list_objects_in_directory(container, directory, files);
```

jakub-nowakowski-percona (approved):

storage/innobase/xtrabackup/src/xbcloud/object_store.cc
satya-bodapati: Yep, make sense @jakub-nowakowski-percona  I will fix it.

PR #1729, opened 2026-01-28
Optimize recovery performance for compressed tables

zhaoxinyu:
https://perconadev.atlassian.net/browse/PXB-3658

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_local.cc
+++ b/storage/innobase/xtrabackup/src/ds_local.cc
@@ -186,13 +186,7 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
       return 1;
     }
 
-#ifdef HAVE_FALLOC_PUNCH_HOLE_AND_KEEP_SIZE
-    if (punch_hole_supported) {
-      fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, seek,
-                sparse_map[i].skip);
-    }
-#endif
-
+    posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED);
     ptr += sparse_map[i].len;
   }
   /* to track if last page is sparse */


zhaoxinyu:
@satya-bodapati , Hi, Satya, is this PR ok to be merged?

satya-bodapati:
> @satya-bodapati , Hi, Satya, is this PR ok to be merged?

While I agree that it is a problem, I think this fadvise(0,0) is also problematic. It means remove entire file from page cache and un-necessarily do fadvise of entire file for every write operation.

Just removing punch_hole is fine. We also need testcases to verify correctness. I will have to rework this a bit.

satya-bodapati:
#1739 is created with additional test scenarios and better commit message.

satya-bodapati:
Thank you @zhaoxinyu  for creating the bug and suggesting a patch. Lets track the development from #1739

zhaoxinyu:
> > @satya-bodapati , Hi, Satya, is this PR ok to be merged?
> 
> While I agree that it is a problem, I think this fadvise(0,0) is also problematic. It means remove entire file from page cache and un-necessarily do fadvise of entire file for every write operation.
> 
> Just removing punch_hole is fine. We also need testcases to verify correctness. I will have to rework this a bit.

@satya-bodapati  Or can we simply fadvise the necessary data segment in `local_write_sparse`? At present, we have fadvise(0, 0) in `local_write`. I think it is useful for streaming large files, as fadvise can actively write back dirty pages in the specified region, lowering the risk of I/O thrashing when the system is under high memory pressure.

BTW, in our tests, there is not much difference between fadvise(0,0) and fadvise(offset, len).

satya-bodapati:
> > > @satya-bodapati , Hi, Satya, is this PR ok to be merged?
> > 
> > 
> > While I agree that it is a problem, I think this fadvise(0,0) is also problematic. It means remove entire file from page cache and un-necessarily do fadvise of entire file for every write operation.
> > Just removing punch_hole is fine. We also need testcases to verify correctness. I will have to rework this a bit.
> 
> @satya-bodapati Or can we simply fadvise the necessary data segment in `local_write_sparse`? At present, we have fadvise(0, 0) in `local_write`. I think it is useful for streaming large files, as fadvise can actively write back dirty pages in the specified region, lowering the risk of I/O thrashing when the system is under high memory pressure.
> 
> BTW, in our tests, there is not much difference between fadvise(0,0) and fadvise(offset, len).

Ok. added the fadvise().

PR #1733, opened 2026-04-14
PXB-3328 [8.4] : Add --check-tables option to validate InnoDB indexes during prepare

satya-bodapati:
## JIRA
https://perconadev.atlassian.net/browse/PXB-3328

## Problem
After `xtrabackup --prepare` completes redo apply, there is no built-in way to verify that the B-tree indexes are structurally correct. Users must restore and run `CHECK TABLE` on a live server to find corruption. This delays discovery and complicates backup validation workflows.

## Solution
Add `--check-tables` to `xtrabackup --prepare` which validates every committed InnoDB index after recovery. This is a prepare-phase-only option: it is rejected with `--backup`, `--copy-back`, `--move-back`, and other non-prepare modes. It can be combined with `--export` to validate indexes and produce export artifacts in a single prepare invocation. Works with both `--apply-log-only` and final prepare since validation is strictly read-only.

### 1. Tablespace filtering
All ibd tablespaces are validated via a single `fsp_is_ibd_tablespace()` check, covering both file-per-table and general (shared) tablespaces including partitions. System (`ibdata`), undo, and temporary tablespaces are excluded.

### 2. Dictionary loading
For each qualifying tablespace, load `dict_table_t` objects via `dict_load_from_spaces_sdi()`, similar to what `--export` already does.

### 3. Parallel B-tree validation
Spawn `--parallel` worker threads that call `btr_validate_index()` on every committed index. All corruptions are reported (no fail-fast). Per-index and per-table names are printed for each failure.

Index type handling:
- **B-tree** (regular, unique, prefix, functional, multi-valued): full `btr_validate_level()` per level
- **Spatial (R-tree)**: `btr_validate_spatial_index()` with relaxed ordering
- **Full-text (DICT_FTS)**: skipped (FTS uses auxiliary tables, not B-tree)
- **INSTANT ALTER**: handled transparently via `rec_validate()` which understands row versioning
- **Virtual/functional indexes**: validate materialized key values only, no expression evaluation

### 4. Graceful assertion handling
InnoDB's `btr_validate_level()` has 11 `ut_a()` assertions that crash on corrupted pages. Under `#ifdef XTRABACKUP` these are converted to soft errors that log a message and return false. Corrupted sibling links (e.g. all-zero pages with `FIL_PAGE_NEXT=0`) are handled by stopping traversal to prevent infinite loops.

### 5. LOB corruption detection (PS-9683)
Port of Percona Server fix to detect external LOB pages shared between two records. Uses `blob_ref_map` to track LOB first pages. Defensive checks added for corrupted LOB references (missing tablespace, out-of-bounds page number).

### 6. Option validation
`--check-tables` is only valid with `--prepare`. Using it with `--backup`, `--copy-back`, `--move-back`, or other modes produces an error and exits. It can be combined with `--export` (both run sequentially during prepare).

### 7. Test coverage

**`check_tables.sh`** (7 scenarios):
1. Positive multi-combo covering all major table/index types:
   encrypted, compressed, plain, general tablespace (5 tables),
   FTS, INSTANT ALTER (add/drop/re-add column churn), functional index,
   virtual column index, GIS/spatial index, HASH partition (3 parts),
   RANGE partition (3 parts), JSON multi-valued index, foreign key
   parent+child, invisible columns -- with double incremental backup
2. Physical `PAGE_INDEX_ID` corruption
3. Raw page checksum corruption
4. All-zero page corruption (unflushed page from redo)
5. Multiple corrupted tables — verifies all are reported
6. Invalid option combinations (`--copy-back --check-tables`, `--backup --check-tables`) — verifies rejection
7. `--prepare --export --check-tables` combined — verifies both export artifacts and index validation

**`check_tables_debug.sh`** (5 sub-tests):
Broken sibling link, wrong index ID, wrong min-record flag, catch-all injection, LOB duplicate detection

### DIFF
--- a/storage/innobase/btr/btr0btr.cc
+++ b/storage/innobase/btr/btr0btr.cc
@@ -4096,7 +4096,8 @@ static void btr_validate_report2(
 @return true if ok */
 static bool btr_validate_level(
     dict_index_t *index, const trx_t *trx, ulint level,
-    bool lockout IF_DEBUG(, Index_details &index_details)) {
+    bool lockout IF_DEBUG(, Index_details &index_details),
+    blob_ref_map *blob_map = nullptr) {
   buf_block_t *block;
   page_t *page;
   buf_block_t *right_block = nullptr; /* remove warning */
@@ -4169,13 +4170,40 @@ static bool btr_validate_level(
       ret = false;
     }
 
+#ifdef XTRABACKUP
+    /* In xtrabackup --check-tables we report corruption instead of crashing.
+       InnoDB uses ut_a here because space ID mismatch or wrong page level
+       during B-tree descent is unrecoverable for a running server. In
+       xtrabackup we only need a damage report. */
+    if (index->space != block->page.id.space() ||
+        index->space != page_get_space_id(page)) {
+      btr_validate_report1(index, level, block);
+      ib::error() << "B-tree corruption: space ID mismatch during descent."
+                  << " Expected " << index->space << ", got block space "
+                  << block->page.id.space() << ", page space "
+                  << page_get_space_id(page) << " in index " << index->name();
+      mtr_commit(&mtr);
+      mem_heap_free(heap);
+      return false;
+    }
+    if (page_is_leaf(page)) {
+      btr_validate_report1(index, level, block);
+      ib::error() << "B-tree corruption: unexpected leaf page "
+                  << page_get_page_no(page) << " during descent to level "
+                  << level << " in index " << index->name();
+      mtr_commit(&mtr);
+      mem_heap_free(heap);
+      return false;
+    }
+#else
     ut_a(index->space == block->page.id.space());
     ut_a(index->space == page_get_space_id(page));
+    ut_a(!page_is_leaf(page));
+#endif /* XTRABACKUP */
 #ifdef UNIV_ZIP_DEBUG
     page_zip = buf_block_get_page_zip(block);
     ut_a(!page_zip || page_zip_validate(page_zip, page, index));
 #endif /* UNIV_ZIP_DEBUG */
-    ut_a(!page_is_leaf(page));
 
     page_cur_set_before_first(block, &cursor);
     page_cur_move_to_next(&cursor);
@@ -4222,6 +4250,7 @@ static bool btr_validate_level(
   }
 
   do {
+    ulint cur_page_no;
     mem_heap_empty(heap);
     offsets = offsets2 = nullptr;
     if (!srv_read_only_mode) {
@@ -4237,7 +4266,32 @@ static bool btr_validate_level(
     ut_a(!page_zip || page_zip_validate(page_zip, page, index));
 #endif /* UNIV_ZIP_DEBUG */
 
+#ifdef XTRABACKUP
+    /* In xtrabackup --check-tables we report corruption instead of crashing.
+       Space ID mismatch in the page scan means the page does not belong to
+       this index's tablespace -- severe corruption. Stop traversing this
+       index because sibling links on a corrupted page are untrustworthy. */
+    if (block->page.id.space() != index->space) {
+      btr_validate_report1(index, level, block);
+      ib::error() << "B-tree corruption: page space ID "
+                  << block->page.id.space() << " != index space "
+                  << index->space << " on page " << block->page.id.page_no()
+                  << " in index " << index->name();
+      ret = false;
+      right_page_no = FIL_NULL;
+      goto node_ptr_fails;
+    }
+#else
     ut_a(block->page.id.space() == index->space);
+#endif /* XTRABACKUP */
+
+    DBUG_EXECUTE_IF("check_table_wrong_index_id", {
+      thread_local bool injected = false;
+      if (!injected && btr_page_get_index_id(page) == index->id) {
+        mach_write_to_8(page + PAGE_HEADER + PAGE_INDEX_ID, index->id + 1);
+        injected = true;
+      }
+    });
 
     if (fseg_page_is_free(seg, block->page.id.space(),
                           block->page.id.page_no())) {
@@ -4252,7 +4306,7 @@ static bool btr_validate_level(
 
       ret = false;
 
-    } else if (!page_validate(page, index)) {
+    } else if (!page_validate(page, index, true, blob_map)) {
       btr_validate_report1(index, level, block);
       ret = false;
 
@@ -4263,17 +4317,53 @@ static bool btr_validate_level(
       ret = false;
     }
 
+#ifdef XTRABACKUP
+    /* In xtrabackup --check-tables we report corruption instead of crashing.
+       Page level mismatch means the page has wrong metadata (e.g. all-zero
+       page from unflushed redo). Stop traversal -- sibling links are
+       untrustworthy on a corrupted page. */
+    if (btr_page_get_level(page) != level) {
+      btr_validate_report1(index, level, block);
+      ib::error() << "B-tree corruption: page level "
+                  << btr_page_get_level(page) << " != expected level " << level
+                  << " on page " << page_get_page_no(page) << " in index "
+                  << index->name();
+      ret = false;
+      right_page_no = FIL_NULL;
+      goto node_ptr_fails;
+    }
+#else
     ut_a(btr_page_get_level(page) == level);
+#endif /* XTRABACKUP */
     right_page_no = btr_page_get_next(page, &mtr);
     left_page_no = btr_page_get_prev(page, &mtr);
-    const ulint cur_page_no = page_get_page_no(page);
+    cur_page_no = page_get_page_no(page);
 
 #ifdef UNIV_DEBUG
     index_details.add_page(cur_page_no, level);
 #endif /* UNIV_DEBUG */
 
+#ifdef XTRABACKUP
+    /* In xtrabackup --check-tables we report corruption instead of crashing.
+       An empty non-root page indicates corruption (e.g. all-zero page from
+       unflushed redo, or truncated tablespace). Stop traversal -- an
+       all-zero page has FIL_PAGE_NEXT=0 which would loop through non-B-tree
+       pages indefinitely. */
+    if (page_is_empty(page) &&
+        !(level == 0 && page_get_page_no(page) == dict_index_get_page(index))) {
+      btr_validate_report1(index, level, block);
+      ib::error() << "B-tree corruption: page " << page_get_page_no(page)
+                  << " is empty but is not the root page"
+                  << " in index " << index->name()
+                  << ". Possible all-zero (unflushed) page.";
+      ret = false;
+      right_page_no = FIL_NULL;
+      goto node_ptr_fails;
+    }
+#else
     ut_a(!page_is_empty(page) ||
          (level == 0 && page_get_page_no(page) == dict_index_get_page(index)));
+#endif /* XTRABACKUP */
 
     if (right_page_no != FIL_NULL) {
       const rec_t *right_rec;
@@ -4285,6 +4375,14 @@ static bool btr_validate_level(
 
       right_page = buf_block_get_frame(right_block);
 
+      DBUG_EXECUTE_IF("check_table_break_sibling_link", {
+        thread_local bool injected = false;
+        if (!injected && btr_page_get_prev(right_page, &mtr) != 0) {
+          mach_write_to_4(right_page + FIL_PAGE_PREV, 0);
+          injected = true;
+        }
+      });
+
       const auto fil_page_prev = btr_page_get_prev(right_page, &mtr);
       if (fil_page_prev != cur_page_no) {
         btr_validate_report2(index, level, block, right_block);
@@ -4295,9 +4393,15 @@ static bool btr_validate_level(
 
         ret = false;
 #ifdef UNIV_DEBUG
-        /* In debug build, it is best to fail with an assert. */
-        const bool siblings_link_correct = false;
-        ut_ad(siblings_link_correct);
+        {
+          bool skip_assert = false;
+          DBUG_EXECUTE_IF("check_table_break_sibling_link",
+                          skip_assert = true;);
+          if (!skip_assert) {
+            const bool siblings_link_correct = false;
+            ut_ad(siblings_link_correct);
+          }
+        }
 #endif /* UNIV_DEBUG */
       }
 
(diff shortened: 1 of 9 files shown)


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -7089,6 +7155,56 @@ static void xtrabackup_prepare_func(int argc, char **argv) {
     datafiles_iter_free(it);
   }
 
+  if (xtrabackup_check_tables) {
+    xb::info() << "Starting table checks (--check-tables).";
+    bool check_failed = false;
```
jakub-nowakowski-percona, line 7165: `check_failed` can be modified from multiple threads, so it should be atomic or guarded.

storage/innobase/page/page0page.cc
```diff
@@ -1721,6 +1722,96 @@ bool page_rec_validate(
...
+        continue;
+      }
+
+      space_id_t blob_space_id = ref.space_id();
+      page_no_t blob_page_no = ref.page_no();
+
+      page_id_t blob_page_id(blob_space_id, blob_page_no);
+      bool is_free = fseg_page_is_free(nullptr, blob_space_id, blob_page_no);
+      if (is_free) {
+        ut_ad(0);
```
jakub-nowakowski-percona: Should we fail the assertion here? Won't it defeat the effort to report errors gracefully?

storage/innobase/btr/btr0btr.cc
```diff
@@ -4237,7 +4266,32 @@ static bool btr_validate_level(
...
+      ret = false;
+      right_page_no = FIL_NULL;
+      goto node_ptr_fails;
+    }
+#else
     ut_a(block->page.id.space() == index->space);
+#endif /* XTRABACKUP */
+
+    DBUG_EXECUTE_IF("check_table_wrong_index_id", {
+      static bool injected = false;
```
jakub-nowakowski-percona: I believe those static `injected` flags can be set from multiple threads, so  maybe they should be `thread_local`? There are several flags like that.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6734,6 +6743,63 @@ static void space_extend_thread_func(data_thread_ctxt_t *ctxt) {
...
+
+    if (!fsp_is_file_per_table(space->id, space->flags) &&
+        !fsp_is_shared_tablespace(space->flags)) {
+      continue;
+    }
+
+    auto result = xb::prepare::dict_load_from_spaces_sdi(space->id);
+    if (std::get<0>(result) != DB_SUCCESS) {
+      xb::warn() << "Cannot load dictionary for tablespace " << space->name;
+      continue;
```
jakub-nowakowski-percona, line 6767: Shouldn't this scenario result in an error rather than a warning? Is there a valid reason why we wouldn't be able to load a dictionary?

satya-bodapati: Agree, this should be an error.

storage/innobase/btr/btr0btr.cc
satya-bodapati: Good point.

storage/innobase/page/page0page.cc
satya-bodapati: it is debug only assertion, but yeah, we are better off removing it.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: [AI-assistant]: Fixed. `check_failed` writes are now guarded by `count_mutex`. Although all writers only set `true` (no torn-read risk), it's technically UB under the C++ memory model without synchronization. The `count_mutex` is already used for thread coordination, so we reuse it for the error flag.

The read of `check_failed` at line 7203 happens after the polling loop confirms `ct_count == 0` (all threads finished), so the last `mutex_exit` from a worker thread establishes happens-before for the read.

storage/innobase/page/page0page.cc
satya-bodapati: [AI-assistant]: Good catch. Removed the `ut_ad(0)` — it would crash debug builds on the exact corruption we're trying to report gracefully. The error message + `return false` is sufficient.

storage/innobase/btr/btr0btr.cc
satya-bodapati: [AI-assistant]: Good point. Changed both `static bool injected` flags to `static thread_local bool injected` in the `check_table_wrong_index_id` and `check_table_break_sibling_link` debug points. With `static`, if two parallel threads hit the same debug point, the first thread sets `injected = true` and the second thread would skip injection entirely (plus the data race itself is UB). With `thread_local`, each worker thread gets its own flag and injects corruption exactly once per thread.

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: [AI-assistant]: Agreed — changed from `xb::warn()` to `xb::error()` and now also sets `*(ctxt->error) = true` (guarded by `count_mutex`), so a tablespace whose dictionary can't be loaded fails the overall check.

Also added an explicit `fsp_is_system_tablespace(space->id)` skip before the existing filter, since the system tablespace (`innodb_system`, space_id=0) passes the `fsp_is_shared_tablespace()` check but doesn't have SDI, causing `dict_load_from_spaces_sdi()` to fail. Previously this was masked by the warning-only behavior.

jakub-nowakowski-percona: Should we `continue` or just exit the function with error set?

jakub-nowakowski-percona (approved):

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: we should continue, this a best effort, we try to verify as many tables as possible. Otherwise it is painful to know that there is always a probelm with another table...

also, if we know certain tables are not usable, other tables can still be restored using IMPORT TABLESPACE..

jakub-nowakowski-percona (approved):

storage/innobase/page/page0page.cc
```diff
@@ -1721,6 +1722,95 @@ bool page_rec_validate(
...
+      }
+
+      if (ref.length() == 0) {
+        continue;
+      }
+
+      space_id_t blob_space_id = ref.space_id();
+      page_no_t blob_page_no = ref.page_no();
+
+      page_id_t blob_page_id(blob_space_id, blob_page_no);
```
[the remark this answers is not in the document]
satya-bodapati: [AI-assistant]: Fixed. Added defensive checks before `fseg_page_is_free()`: validate that `fil_space_get(blob_space_id)` is non-null (missing tablespace) and `blob_page_no < fil_space_get_size()` (out-of-bounds). Both cases now return false with a diagnostic error message instead of crashing.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3567,11 +3567,13 @@ bool fseg_page_is_free(fseg_header_t *seg_header, /*!< in: segment header */
...
-  ut_a(seg_inode);
-  ut_ad(mach_read_from_4(seg_inode + FSEG_MAGIC_N) == FSEG_MAGIC_N_VALUE);
-  ut_ad(!((page_offset(seg_inode) - FSEG_ARR_OFFSET) % FSEG_INODE_SIZE));
+    ut_a(seg_inode);
+    ut_ad(mach_read_from_4(seg_inode + FSEG_MAGIC_N) == FSEG_MAGIC_N_VALUE);
+    ut_ad(!((page_offset(seg_inode) - FSEG_ARR_OFFSET) % FSEG_INODE_SIZE));
+  }
 
   descr = xdes_get_descriptor(space_id, page, page_size, &mtr);
   ut_a(descr);
```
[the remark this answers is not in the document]
satya-bodapati: [AI-assistant]: Fixed. Added `#ifdef XTRABACKUP` null checks in `fseg_page_is_free()`: if `fil_space_get()` returns nullptr, returns true (treat as free). If `xdes_get_descriptor()` returns nullptr, commits the mtr and returns true. In both cases this causes the caller to report the LOB page as free (corruption), which is the correct outcome for garbage data. The original `ut_a(descr)` is preserved for non-XTRABACKUP builds.

storage/innobase/include/page0page.h
```diff
@@ -761,15 +761,33 @@ bool page_simple_validate_old(
 bool page_simple_validate_new(
     const page_t *page); /*!< in: index page in ROW_FORMAT!=REDUNDANT */
 
+/** A blob map to track the first page no of external LOB and its parent record
+which is the <page_no, heap_no>. This is used to find duplicate external LOB
+pages that are shared between two records.  This can happen only on corruption.
+--check-tables uses this map to report corruption. */
+using blob_ref_map = std::unordered_map<page_no_t, std::pair<page_no_t, ulint>>;
+
```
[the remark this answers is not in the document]
satya-bodapati: [AI-assistant]: Fixed. Added `#include <unordered_map>` to `page0page.h`.

storage/innobase/xtrabackup/test/t/check_tables_debug.sh
```diff
@@ -0,0 +1,124 @@
+############################################################################
+# Test --check-tables corruption detection via DBUG_EXECUTE_IF debug points.
+# Requires a debug build of xtrabackup.
+#
+# Sub-tests:
+#   A) check_table_break_sibling_link  -- broken FIL_PAGE_NEXT/PREV chain
+#   B) check_table_wrong_index_id      -- page belongs to wrong index
+#   C) check_table_set_wrong_min_bit   -- min-record flag on wrong record
+#   D) check_table_inject_corruption   -- catch-all validation failure
```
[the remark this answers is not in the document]
satya-bodapati: [AI-assistant]: Fixed. Added sub-test E to the header comment.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -6734,6 +6743,72 @@ static void space_extend_thread_func(data_thread_ctxt_t *ctxt) {
...
+    fil_space_t *space = node->space;
+
+    if (fsp_is_system_tablespace(space->id)) {
+      continue;
+    }
+
+    if (!fsp_is_file_per_table(space->id, space->flags) &&
+        !fsp_is_shared_tablespace(space->flags)) {
+      continue;
+    }
```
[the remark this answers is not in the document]
satya-bodapati: [AI-assistant]: Good catch. Added explicit skips for `fsp_is_undo_tablespace(space->id)` and `fsp_is_system_temporary(space->id)` alongside the existing `fsp_is_system_tablespace()` check. These tablespaces have no SDI and would cause `dict_load_from_spaces_sdi()` to fail.

storage/innobase/page/page0page.cc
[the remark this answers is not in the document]
satya-bodapati: more defense checks, will take it

storage/innobase/fsp/fsp0fsp.cc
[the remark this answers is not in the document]
satya-bodapati: more defense checks, will do it

storage/innobase/include/page0page.h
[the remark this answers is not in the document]
satya-bodapati: can be ignored.

storage/innobase/xtrabackup/src/xtrabackup.cc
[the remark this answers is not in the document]
satya-bodapati: I think check at 6764 ensure that undo and tmep are skipped.. may be ebtter to write explicitly

storage/innobase/fsp/fsp0fsp.cc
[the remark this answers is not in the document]
satya-bodapati: may be add error message that we couldn't get extent descriptor for the page_id?

storage/innobase/include/page0page.h
```diff
@@ -39,6 +39,8 @@ this program; if not, write to the Free Software Foundation, Inc.,
 
 #include "univ.i"
 
+#include <unordered_map>
+
```
satya-bodapati, line 42: remove extra empty line

storage/innobase/xtrabackup/test/t/check_tables.sh
```diff
@@ -0,0 +1,384 @@
...
+vlog "INSTANT ALTER: add column c to t_instant after initial data"
+mysql test <<EOF
+ALTER TABLE t_instant ADD COLUMN c INT DEFAULT 42, ALGORITHM=INSTANT;
+ALTER TABLE t_instant ADD COLUMN d VARCHAR(50) DEFAULT 'added', ALGORITHM=INSTANT;
+ALTER TABLE t_instant DROP COLUMN d, ALGORITHM=INSTANT;
+ALTER TABLE t_instant ADD COLUMN e BIGINT, ALGORITHM=INSTANT;
+EOF
+
+vlog "Full backup"
+xtrabackup --backup --target-dir=$topdir/full \
```
satya-bodapati, line 72: always log xtrabackup comamnd to outputs to new file names (use based on scenario) and also tee the ouptut so taht it gets tor run.sh error log. remove file afterwards You already this below.

If you streaming becareful to not redirect stdout to stderr. Use this strategy: 2> >(tee $topdir/compress_encrypt_stream_backup.log >&2)

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3561,20 +3561,35 @@ bool fseg_page_is_free(fseg_header_t *seg_header, /*!< in: segment header */
 
   fil_space_t *space = fil_space_get(space_id);
 
+#ifdef XTRABACKUP
+  if (space == nullptr) {
```
satya-bodapati, line 3565: Also add proper error message here. from where we are not able to load the tablespace, print space_id.

storage/innobase/fsp/fsp0fsp.cc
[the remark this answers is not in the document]
satya-bodapati: Fixed. Added `ib::error()` message: `"Cannot get extent descriptor for space_id: X page_no: Y"` when `xdes_get_descriptor()` returns nullptr. Also added a similar message (`"Cannot load tablespace for space_id: X page_no: Y"`) for the `fil_space_get()` null check above it.

storage/innobase/include/page0page.h
satya-bodapati: Fixed. Removed the extra blank line.

storage/innobase/xtrabackup/test/t/check_tables.sh
satya-bodapati: Fixed. All xtrabackup backup and intermediate prepare commands now capture stderr to per-scenario log files using `2> >(tee $topdir/logfile.log >&2)`.

satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/142/

PR #1737, opened 2026-04-16
PXB-3747: report backup size and uncompressed size

satya-bodapati:
## Summary

xtrabackup now embeds two byte-accurate size fields in `xtrabackup_info`
and mirrors them to the error log:

- **`backup_size`** (bytes) — final bytes the backup wrote to disk
  (or to the stream). Always reported.
- **`uncompressed_backup_size`** (bytes) — logical pre-compression
  bytes (data + redo + metadata + server-encrypted InnoDB files +
  RocksDB SSTs that skip the compressor). Reported only when
  `--compress` is used.

Both values are byte-perfect against the actual on-disk artifact for
every combination of `--compress{,-lz4,-zstd}` / `--encrypt` /
`--stream=xbstream` / `--extra-lsndir` / incremental / sparse.
Strict-equality asserts in `t/backup_size_basic.sh` and
`t/backup_size_compress.sh` cover 15+ scenarios end-to-end.

JIRA: https://perconadev.atlassian.net/browse/PXB-3747

## How to use this feature

No new flags or options — the behavior is enabled for every backup.

1. **Final backup size (`backup_size`)** — always present in
   `xtrabackup_info`, e.g.:

       backup_size = 129456789

   and logged:

       Backup size: 123.45 MiB (129456789 bytes)

2. **Uncompressed size under `--compress` (`uncompressed_backup_size`)** —
   added only when `--compress{,-lz4,-zstd}` is used:

       uncompressed_backup_size = 479123456

   along with two extra log lines:

       Uncompressed backup size: 456.78 MiB (479123456 bytes)
       Compression ratio: 3.70x

   This helps operators estimate the disk space required to extract
   a compressed backup before running `xbstream -x` /
   `xtrabackup --decompress`, and makes backups taken at different
   dates comparable regardless of compressibility.

3. **Where `xtrabackup_info` lives** — the copy inside the backup
   directory follows the transport the user picked: it is compressed
   under `--compress`, encrypted under `--encrypt`, and wrapped in
   the xbstream under `--stream=xbstream`. To always obtain an
   unencrypted, uncompressed copy of `xtrabackup_info`, pass
   `--extra-lsndir=<dir>` — the file written there is plaintext
   regardless of `--compress` / `--encrypt`.

4. **Error-log output** — the two (or three, under `--compress`)
   lines above appear at the end of the backup, just before the
   "completed OK!" banner.

## Commits

Split into two focused, bisect-friendly commits:

1. **`PXB-3747 : Report final on-disk backup_size in xtrabackup_info`**
   — adds `backup_size` with zero `--compress` awareness; includes
   the shared strict-size test infrastructure.
2. **`PXB-3747 : Report uncompressed_backup_size for --compress backups`**
   — adds `uncompressed_backup_size`, the "Uncompressed backup size"
   log line, and the "Compression ratio" log line, wired up only
   for `--compress` runs.

## Commit 1 design — `backup_size`

The datasink pipeline has exactly one authoritative component: the
**leaf**. Every byte that lands in the backup artifact passes
through `ds_local` / `ds_stdout` / `ds_fifo`. Wrapper datasinks
(encrypt, xbstream, buffer, tmpfile, compress) only transform and
forward bytes.

```
    ds_data (head)
         |  ds_write()
         v
    [ wrappers: xbstream / encrypt / compress / ... ]
         |  pipe_ctxt
         v
    leaf (ds_local | ds_stdout | ds_fifo)   authoritative byte count
         |
         v
    my_write() / fwrite()
```

Rather than adding one-off `get_foo()` vtable slots per metric, the
patch introduces a tiny per-datasink metrics API: each datasink
can publish named `{name, value}` pairs via an optional
`report_metrics` slot, and callers read one off a specific node
with `ds_find_metric`. The three leaves publish an atomic
`bytes_written` counter bumped on every successful write.
`get_final_backup_size()` walks to the leaf and reads it;
`backup_size = N` is written to `xtrabackup_info` and logged.

`Tablespace_map::serialize()` and the transition-key dumper are
moved ahead of `backup_finish()` so their bytes are included in
the reported `backup_size`.

### Why two data pipelines exist

xtrabackup maintains two top-level data pipelines because not
every file benefits from compression:

- **`ds_data`** — normal InnoDB tablespaces, redo, metadata.
  Compressed when `--compress` is on.
- **`ds_uncompressed_data`** — server-encrypted InnoDB tablespaces
  (re-compressing ciphertext is wasteful) and RocksDB SST files
  (already compressed internally). Bypasses the compress wrapper.

Both pipelines share their encrypt / xbstream / leaf tail as an
init-time context-sharing optimization:

```
    ds_data              --> buffer --> compress --> encrypt --> leaf
                                                                  ^
                                                                  |
    ds_uncompressed_data -----------------------> encrypt --------+
```

### Why the counter cannot live on any `ds_ctxt_t`

The shared tail makes "attach a counter to the ctxt" broken no
matter which ctxt is picked:

- **Leaf ctxt.** Under `--stream=xbstream`, `ds_data`, `ds_redo`,
  `ds_meta`, and `ds_uncompressed_data` all terminate at the same
  leaf. A leaf-ctxt counter would sum bytes from every pipeline,
  and for the compressed pipeline those bytes are post-compression
  — neither the logical total nor the on-disk total.
- **Encrypt / xbstream / buffer ctxts.** `ds_data` and
  `ds_uncompressed_data` share these too. Counting there mixes
  post-compression bytes (from `ds_data`) with raw bytes (from
  `ds_uncompressed_data`); the sum is meaningless.

### Per-file `ds_file_t` handle is the right anchor

The counter attaches where a single logical write unambiguously
enters exactly one pipeline — on the per-file handle returned by
the top-level `ds_open()`. Each top-level open creates a fresh
`ds_file_t`, the caller decides whether that file's bytes count,
and `ds_write` / `ds_write_sparse` bump the counter with the
length supplied at the entry, before any wrapper transforms it:

```
    xtrabackup top-level open
         |  ds_open_track_uncomp(..., xb_global_track_uncomp_bytes())
         |                             |
         |                             +-- nullptr when !--compress
         v
    ds_file_t { uncomp_bytes -> xb_uncomp_bytes_counter }
         |  ds_write(len) / ds_write_sparse(packed_len)
         |     if (file->uncomp_bytes) uncomp_bytes->add_uncompressed(len)
         v
    [ compress / encrypt / buffer / xbstream wrappers ]
         |
         v
    leaf  (already counted as backup_size by commit 1)

    xb_uncomp_bytes_counter.get_uncompressed_backup_size()
         |
         v
    uncompressed_backup_size  (xtrabackup_info + error log)
```

Tracking is opt-in and scoped to the top-level opens on `ds_data`
/ `ds_redo` / `ds_meta` / `ds_uncompressed_data`, so the hot-path
cost is one atomic add per write on `--compress` runs; non-compress
runs skip even that. The same counter aggregates both pipelines,
so server-encrypted IBDs and RocksDB SSTs are counted exactly
once, at their logical size.

`xb_global_track_uncomp_bytes()` returns a real counter only when
`--compress` is active, so non-compress runs emit neither the
field nor the log lines.

## Test plan

- **`t/backup_size_basic.sh`** — `--target-dir`, `--stream=xbstream`,
  `--encrypt` on both, incremental chains, sparse tablespaces, full
  restore round-trip. Every scenario asserts `backup_size` matches
  the on-disk / on-stream byte count exactly; `uncompressed_backup_size`
  must **not** be present.
- **`t/backup_size_compress.sh`** — `--compress` target-dir,
  `--compress` xbstream, `--compress` + `--encrypt`, incremental
  chains with compression, RocksDB, server-encrypted InnoDB,
  redo-log encryption, sparse + `--compress`, and `--extra-lsndir`.
  Every scenario asserts `backup_size` matches the compressed
  on-disk size and, after decompression, `uncompressed_backup_size`
  matches the decompressed tree size byte-for-byte.
- **`suites/reducedlock/backup_size_reduced.sh`** — reduced-lock
  + concurrent TRUNCATE edge case (zero-size tablespace).
- **`-s compression`** — 12/12 runnable tests pass; shared
  strict-size helpers live in `inc/common.sh`.

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -1594,6 +1594,24 @@ bool backup_start(Backup_context &context) {
   return (true);
 }
 
+/** Report backup_size to the error log.  A successful backup run must
+have produced on-disk output, so the leaf counter must be non-zero;
+a zero value indicates a silent reporting bug: assert in debug, warn
+and skip in release. */
+static void report_backup_size() {
+  const unsigned long long backup_size = get_final_backup_size();
+
+  ut_ad(backup_size > 0);
+  if (backup_size == 0) {
+    xb::warn() << "Backup size reporting failed: leaf counter returned 0";
+    return;
+  }
+
+  xb::info() << "Backup size: "
+             << xtrabackup::utils::human_readable(backup_size) << " ("
+             << backup_size << " bytes)";
+}
+
 /* Finsh the backup. Release all locks. Write down backup metadata.
 @return true if success. */
 bool backup_finish(Backup_context &context) {
@@ -1649,6 +1667,8 @@ bool backup_finish(Backup_context &context) {
   if (!write_xtrabackup_info(mysql_connection)) {
     return (false);
   }
+
+  report_backup_size();
 
   return (true);
 }
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -55,6 +55,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <chrono>
 #include <fstream>
 #include <functional>
+#include <iomanip>
 #include <queue>
 #include <set>
 #include <sstream>
@@ -377,7 +378,8 @@ bool backup_file_print(const char *filename, const char *message, int len) {
   stat.st_mtime = time(nullptr);
   stat.st_size = len;
 
-  dstfile = ds_open(ds_data, filename, &stat);
+  dstfile = ds_open_track_uncomp(ds_data, filename, &stat,
+                                 xb_global_track_uncomp_bytes());
   if (dstfile == NULL) {
     xb::error() << "cannot open the destination stream for " << filename;
     goto error;
@@ -576,7 +578,9 @@ bool copy_file(ds_ctxt_t *datasink, const char *src_file_path,
 
   strncpy(dst_name, cursor.rel_path, sizeof(dst_name));
 
-  dstfile = ds_open(datasink, trim_dotslash(dst_file_path), &cursor.statinfo);
+  dstfile =
+      ds_open_track_uncomp(datasink, trim_dotslash(dst_file_path),
+                           &cursor.statinfo, xb_global_track_uncomp_bytes());
   if (dstfile == NULL) {
     xb::error() << "cannot open the destination stream for " << dst_name;
     goto error;
@@ -1594,10 +1598,14 @@ bool backup_start(Backup_context &context) {
   return (true);
 }
 
-/** Report backup_size to the error log.  A successful backup run must
-have produced on-disk output, so the leaf counter must be non-zero;
-a zero value indicates a silent reporting bug: assert in debug, warn
-and skip in release. */
+/** Report backup_size (and, under --compress, uncompressed_backup_size
+and the compression ratio) to the error log.  A successful backup run
+must have produced on-disk output, so the leaf counter must be
+non-zero; under --compress at least one top-level ds_open_track_uncomp()
+must have enabled an uncomp_bytes counter, so xb_uncomp_bytes_counter
+must be non-zero too.  A zero value therefore indicates a silent
+reporting bug: assert in debug, warn and skip in release to avoid
+emitting misleading numbers (e.g. "Compression ratio: inf"). */
 static void report_backup_size() {
   const unsigned long long backup_size = get_final_backup_size();
 
@@ -1610,9 +1618,31 @@ static void report_backup_size() {
   xb::info() << "Backup size: "
              << xtrabackup::utils::human_readable(backup_size) << " ("
              << backup_size << " bytes)";
+
+  if (xtrabackup_compress == XTRABACKUP_COMPRESS_NONE) {
+    return;
+  }
+
+  const unsigned long long uncompressed_backup_size =
+      get_uncompressed_backup_size();
+
+  ut_ad(uncompressed_backup_size > 0);
+  if (uncompressed_backup_size == 0) {
+    xb::warn() << "Uncompressed backup size reporting failed: metrics"
+                  " counter is 0 despite --compress";
+    return;
+  }
+
+  xb::info() << "Uncompressed backup size: "
+             << xtrabackup::utils::human_readable(uncompressed_backup_size)
+             << " (" << uncompressed_backup_size << " bytes)";
+
+  const double ratio = (double)uncompressed_backup_size / (double)backup_size;
+  xb::info() << "Compression ratio: " << std::fixed << std::setprecision(2)
+             << ratio << "x";
 }
 
-/* Finsh the backup. Release all locks. Write down backup metadata.
+/* Finish the backup. Release all locks. Write down backup metadata.
 @return true if success. */
 bool backup_finish(Backup_context &context) {
   /* release all locks */
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -1849,6 +1849,11 @@ char *get_xtrabackup_info(MYSQL *connection) {
   format_time(history_start_time, buf_start_time, time_buf_size);
   format_time(history_end_time, buf_end_time, time_buf_size);
 
+  /* Sampled here, after all data has drained to the leaf (see the
+  reordering in xtrabackup_backup_func()).  Embedded directly into
+  xtrabackup_info so consumers see the final on-disk byte count. */
+  const unsigned long long backup_size = get_final_backup_size();
+
   ut_a(uuid);
   ut_a(server_version);
   char *result = NULL;
@@ -1875,7 +1880,8 @@ char *get_xtrabackup_info(MYSQL *connection) {
                "format = %s\n"
                "compressed = %s\n"
                "encrypted = %s\n"
-               "lock_ddl_type = %s\n",
+               "lock_ddl_type = %s\n"
+               "backup_size = %llu\n",
                uuid,                                 /* uuid */
                opt_history ? opt_history : "",       /* name */
                tool_name,                            /* tool_name */
@@ -1901,7 +1907,8 @@ char *get_xtrabackup_info(MYSQL *connection) {
                xtrabackup_compress ? "compressed" : "N",     /* compressed */
                xtrabackup_encrypt ? "Y" : "N",               /* encrypted */
                ddl_lock_type_to_str(static_cast<lock_ddl_type_t>(opt_lock_ddl))
-                   .c_str()); /* lock-ddl */
+                   .c_str(), /* lock-ddl */
+               backup_size); /* backup_size */
 
   ut_a(ret != 0);
 
--- a/storage/innobase/xtrabackup/src/backup_mysql.cc
+++ b/storage/innobase/xtrabackup/src/backup_mysql.cc
@@ -1912,6 +1912,26 @@ char *get_xtrabackup_info(MYSQL *connection) {
 
   ut_a(ret != 0);
 
+  /* uncompressed_backup_size is only meaningful under --compress: it
+  reports the raw, pre-compression logical volume that fed the main
+  backup pipeline.  Without --compress, logical == physical, and the
+  backup_size line above already captures it. */
+  if (xtrabackup_compress != XTRABACKUP_COMPRESS_NONE) {
+    const unsigned long long uncompressed_backup_size =
+        get_uncompressed_backup_size();
+
+    char *tmp = NULL;
+    int ret2 = asprintf(&tmp, "%suncompressed_backup_size = %llu\n", result,
+                        uncompressed_backup_size);
+    if (ret2 < 0) {
+      xb::warn() << "Failed to append uncompressed_backup_size to"
+                 << " xtrabackup_info: " << strerror(errno);
+    } else {
+      free(result);
+      result = tmp;
+    }
+  }
+
   free(server_version);
   return result;
 }
--- a/storage/innobase/xtrabackup/src/datasink.cc
+++ b/storage/innobase/xtrabackup/src/datasink.cc
@@ -111,6 +111,10 @@ ds_file_t *ds_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *stat) {
   file = ctxt->datasink->open(ctxt, path, stat);
   if (file != NULL) {
     file->datasink = ctxt->datasink;
+    /* Save the ctxt this file is attached to so per-datasink state
+    (e.g. leaf bytes_written counters) can be reached from the write
+    paths via file->ctxt->ptr. */
+    file->ctxt = ctxt;
   }
 
   return file;
@@ -160,4 +164,30 @@ Set the destination pipe for a datasink (only makes sense for compress and
 tmpfile). */
 void ds_set_pipe(ds_ctxt_t *ctxt, ds_ctxt_t *pipe_ctxt) {
   ctxt->pipe_ctxt = pipe_ctxt;
+}
+
+const ds_ctxt_t *ds_leaf(const ds_ctxt_t *head) {
+  const ds_ctxt_t *c = head;
+  if (c == nullptr) return nullptr;
+  while (c->pipe_ctxt != nullptr) {
+    c = c->pipe_ctxt;
+  }
+  return c;
+}
+
+bool ds_find_metric(const ds_ctxt_t *node, std::string_view name,
+                    uint64_t *out) {
+  if (node == nullptr || node->datasink == nullptr ||
+      node->datasink->report_metrics == nullptr) {
+    return false;
+  }
+  std::vector<ds_metric> v;
+  node->datasink->report_metrics(node, v);
+  for (const auto &m : v) {
+    if (m.name == name) {
+      if (out != nullptr) *out = m.value;
+      return true;
+    }
+  }
+  return false;
 }
--- a/storage/innobase/xtrabackup/src/datasink.cc
+++ b/storage/innobase/xtrabackup/src/datasink.cc
@@ -113,10 +113,26 @@ ds_file_t *ds_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *stat) {
     file->datasink = ctxt->datasink;
     /* Save the ctxt this file is attached to so per-datasink state
     (e.g. leaf bytes_written counters) can be reached from the write
-    paths via file->ctxt->ptr. */
+    paths via file->ctxt->ptr.  Tracking defaults to off; callers that
+    want uncompressed-byte accounting call ds_track_uncomp() or use
+    the ds_open_track_uncomp() convenience. */
     file->ctxt = ctxt;
+    file->uncomp_bytes = nullptr;
   }
 
+  return file;
+}
+
+void ds_track_uncomp(ds_file_t *file, xb_uncomp_bytes *uncomp_bytes) {
+  if (file != nullptr) {
+    file->uncomp_bytes = uncomp_bytes;
+  }
+}
+
+ds_file_t *ds_open_track_uncomp(ds_ctxt_t *ctxt, const char *path,
+                                MY_STAT *stat, xb_uncomp_bytes *uncomp_bytes) {
+  ds_file_t *file = ds_open(ctxt, path, stat);
+  ds_track_uncomp(file, uncomp_bytes);
   return file;
 }
 
@@ -124,7 +140,15 @@ ds_file_t *ds_open(ds_ctxt_t *ctxt, const char *path, MY_STAT *stat) {
 Write to a datasink file.
 @return 0 on success, 1 on error. */
 int ds_write(ds_file_t *file, const void *buf, size_t len) {
-  return file->datasink->write(file, buf, len);
+  const int rc = file->datasink->write(file, buf, len);
+  /* Bump the backup-run uncompressed byte counter with the logical
+  byte count only on success.  Wrapper-internal opens do not set
+  file->uncomp_bytes, so every logical byte is counted exactly once
+  at the top. */
+  if (rc == 0 && file->uncomp_bytes != nullptr) {
+    file->uncomp_bytes->add_uncompressed(len);
+  }
+  return rc;
 }
 
 /************************************************************************
@@ -143,11 +167,18 @@ Write sparse chunk if supported.
 int ds_write_sparse(ds_file_t *file, const void *buf, size_t len,
                     size_t sparse_map_size, const ds_sparse_chunk_t *sparse_map,
                     bool punch_hole_supported) {
-  if (file->datasink->write_sparse != nullptr) {
-    return file->datasink->write_sparse(file, buf, len, sparse_map_size,
-                                        sparse_map, punch_hole_supported);
+  if (file->datasink->write_sparse == nullptr) {
+    return 1;
   }
-  return 1;
+  const int rc = file->datasink->write_sparse(file, buf, len, sparse_map_size,
+                                              sparse_map, punch_hole_supported);
+  if (rc == 0 && file->uncomp_bytes != nullptr) {
+    /* `len` is the packed (hole-excluded) payload size: callers pre-pack
+    the buffer and pass its length here, and local_write_sparse writes
+    exactly that many bytes across the sparse_map chunks. */
+    file->uncomp_bytes->add_uncompressed(len);
+  }
+  return rc;
 }
 
 /************************************************************************
--- a/storage/innobase/xtrabackup/src/datasink.h
+++ b/storage/innobase/xtrabackup/src/datasink.h
@@ -21,7 +21,9 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #ifndef XB_DATASINK_H
 #define XB_DATASINK_H
 
+#include <my_compiler.h>
 #include <my_dir.h>
+#include <atomic>
 #include <cstdint>
 #include <string_view>
 #include <vector>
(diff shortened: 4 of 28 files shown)


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/144/

storage/innobase/xtrabackup/src/datasink.cc
```diff
@@ -95,6 +96,12 @@ ds_ctxt_t *ds_create(const char *root, ds_type_t type) {
   ctxt = ds->init(root);
   if (ctxt != NULL) {
     ctxt->datasink = ds;
+    /* Per-datasink init() allocates ds_ctxt_t with my_malloc(), which
+    does not zero memory.  Normalize framework-level fields the
+    per-sink init() is not expected to set.  Do NOT touch fields that
+    a leaf's init() is responsible for (e.g. fs_support_punch_hole is
+    probed by local_init() and must be preserved). */
+    ctxt->pipe_ctxt = nullptr;
```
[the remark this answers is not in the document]
satya-bodapati: I think , it might be dangerous to override the punch hole support here. Waht about xbstream that supports it to?
On the other hand, the datasinks that didnt support shouldn't leave it unitialized, there must be a better to place to default initial to false and the datasink set their capabilites based on the tests it perform. 
local datasink checks if FS support punchole, xbstream always support punchole because it writes special format in xbstream to handle punched parts.

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -186,6 +200,12 @@ static int local_write_sparse(ds_file_t *file, const void *buf, size_t len,
       return 1;
     }
 
+    {
+      auto local_ctxt = static_cast<ds_local_ctxt_t *>(file->ctxt->ptr);
+      local_ctxt->bytes_written.fetch_add(sparse_map[i].len,
+                                          std::memory_order_relaxed);
+    }
```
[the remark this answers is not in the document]
satya-bodapati: This is right observation, we had this code before but it was lost somewhere in re-writes. We should fix it.

storage/innobase/xtrabackup/src/datasink.cc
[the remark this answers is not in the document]
satya-bodapati: Addressed in follow-up 7cc89789e5b ("PXB-3747 : Addressing review comments").

The fix follows your suggestion exactly:

- `ds_ctxt_t::pipe_ctxt` gets an in-class default of `nullptr` (matching the existing `fs_support_punch_hole = false` default), so every `new ds_ctxt_t` site starts with sane framework fields.
- The five sinks that allocate via `my_malloc` (local, stdout, xbstream, buffer, tmpfile) now pass `MY_FAE | MY_ZEROFILL`, so the framework fields start at zero/false there too.
- `xbstream_init()` flips `fs_support_punch_hole = true` because the xbstream wire format carries sparse chunks natively and round-trips holes regardless of the extraction filesystem. `local_init()` continues to probe `FALLOC_FL_PUNCH_HOLE` as before.
- `ds_create()` no longer overrides the capability bit; each sink now owns its own advertisement. The blanket-to-false line that regressed local\u2019s probe result is gone.

Verified by running t/backup_size_basic.sh and t/backup_size_compress.sh against a Debug build.

storage/innobase/xtrabackup/src/ds_local.cc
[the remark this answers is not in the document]
satya-bodapati: Addressed in follow-up 7cc89789e5b.

`local_close()` now bumps `ds_local_ctxt_t::bytes_written` by 1 right after the successful 1-byte trailing write (and before `my_free(file)`), so `backup_size` stays authoritative for sparse files under `--target-dir` and no longer under-reports by 1 byte per sparse file.

Verified with t/backup_size_basic.sh (exact-size assertions on scenarios that create sparse files).

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1594,6 +1596,19 @@ bool backup_start(Backup_context &context) {
...
+  else if (bytes >= 1ULL << 20)
+    snprintf(buf, sizeof(buf), "%.2f MiB", (double)bytes / (1ULL << 20));
+  else if (bytes >= 1ULL << 10)
+    snprintf(buf, sizeof(buf), "%.2f KiB", (double)bytes / (1ULL << 10));
+  else
+    snprintf(buf, sizeof(buf), "%llu bytes", bytes);
+  return buf;
+}
+
 /* Finsh the backup. Release all locks. Write down backup metadata.
```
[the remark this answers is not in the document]
satya-bodapati: Fixed in follow-up 7cc89789e5b ("Finsh" -> "Finish").

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1594,6 +1598,51 @@ bool backup_start(Backup_context &context) {
...
 
+/** Report backup_size (and, under --compress, uncompressed_backup_size
+and the compression ratio) to the error log.  A successful backup run
+must have produced on-disk output, so the leaf counter must be
+non-zero; under --compress at least one top-level ds_open_track_uncomp()
+must have enabled an uncomp_bytes counter, so xb_uncomp_bytes_counter
+must be non-zero too.  A
+zero value therefore indicates a silent reporting bug: assert in
+debug, warn and skip in release to avoid emitting misleading numbers
+(e.g. "Compression ratio: inf"). */
```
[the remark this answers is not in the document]
satya-bodapati: Fixed in follow-up 7cc89789e5b. Reflowed so the closing parenthesis lands on the same line as the 'Compression ratio: inf' example it belongs to.

storage/innobase/xtrabackup/test/inc/common.sh
```diff
@@ -87,6 +87,124 @@ function die()
...
+get_field() { grep "^$2 = " "$1" | awk '{print $3}'; }
+
+# file_size <path>
+#   Echo the size in bytes of a single regular file.
+file_size() { stat -c '%s' "$1"; }
+
+# sum_file_bytes <dir>
+#   Echo the sum of the sizes of every regular file under <dir>
+#   (recursive).  0 if the directory is empty or does not exist.
+sum_file_bytes() { find "$1" -type f -printf '%s\n' | awk '{s+=$1} END{print s+0}'; }
```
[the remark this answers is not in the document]
satya-bodapati: Fixed in follow-up 7cc89789e5b. `sum_file_bytes()` now early-returns `0` when the argument is not a directory, which honours the "0 if the directory is empty or does not exist" contract already documented in the helper and silences the stray `find(1)` stderr message.

satya-bodapati:
Pushed follow-up commit `7cc89789e5b` — **PXB-3747 : Addressing review comments**. Will squash before merge.

What it addresses:

1. `fs_support_punch_hole` initialization — `ds_ctxt_t::pipe_ctxt` now defaults to `nullptr` in the struct, and the five `my_malloc`-based sinks (local, stdout, xbstream, buffer, tmpfile) use `MY_FAE | MY_ZEROFILL`. `xbstream_init()` advertises `fs_support_punch_hole = true` since its wire format round-trips sparse chunks natively. `ds_create()` no longer overrides the capability bit.
2. `local_close()` counts the 1-byte trailing-hole write into `bytes_written`, so sparse files under `--target-dir` no longer under-report by 1 byte per file.
3. "Finsh" → "Finish" in `backup_copy.cc`.
4. Reflowed the `report_backup_size()` doc comment so the "Compression ratio: inf" example reads as a single sentence.
5. `sum_file_bytes()` in `inc/common.sh` early-returns `0` when the argument is not a directory, silencing the stray `find(1)` stderr message.

Verified by rebuilding in Debug and running `t/backup_size_basic.sh` and `t/backup_size_compress.sh` — both pass.

storage/innobase/xtrabackup/src/datasink.cc
```diff
@@ -96,12 +96,13 @@ ds_ctxt_t *ds_create(const char *root, ds_type_t type) {
...
-    a leaf's init() is responsible for (e.g. fs_support_punch_hole is
-    probed by local_init() and must be preserved). */
-    ctxt->pipe_ctxt = nullptr;
+    /* Every per-datasink init() is expected to leave the framework-level
+    fields of ds_ctxt_t in a well-defined state: ds_ctxt_t's in-class
+    initializers cover the `new ds_ctxt_t` call sites, and the `my_malloc`
+    call sites use MY_ZEROFILL.  pipe_ctxt therefore starts NULL, and
+    fs_support_punch_hole starts false; each sink then advertises its
+    own capability (local probes the filesystem, xbstream sets it to
+    true because its wire format carries sparse chunks natively). */
```
satya-bodapati: unnecessary comment. remove

storage/innobase/xtrabackup/src/ds_buffer.cc
```diff
@@ -66,9 +66,13 @@ static ds_ctxt_t *buffer_init(const char *root) {
   ds_ctxt_t *ctxt;
   ds_buffer_ctxt_t *buffer_ctxt;
 
-  ctxt = static_cast<ds_ctxt *>(
-      my_malloc(PSI_NOT_INSTRUMENTED,
-                sizeof(ds_ctxt_t) + sizeof(ds_buffer_ctxt_t), MYF(MY_FAE)));
+  /* MY_ZEROFILL so framework-level fields (pipe_ctxt, fs_support_punch_hole)
+  start at well-defined defaults.  buffer is a pass-through wrapper that
+  forwards to its pipe_ctxt; it owns no filesystem or stream of its own,
+  so fs_support_punch_hole stays false. */
```
satya-bodapati: we dont explain zero-fill, it is good practice in general. Remove

storage/innobase/xtrabackup/src/ds_local.cc
```diff
@@ -105,8 +105,12 @@ static ds_ctxt_t *local_init(const char *root) {
     return NULL;
   }
 
-  ctxt = static_cast<ds_ctxt_t *>(
-      my_malloc(PSI_NOT_INSTRUMENTED, sizeof(ds_ctxt_t), MYF(MY_FAE)));
+  /* MY_ZEROFILL so framework-level fields (pipe_ctxt, fs_support_punch_hole)
+  start at well-defined defaults; is_fallocate_punch_hole_supported() below
+  then flips fs_support_punch_hole on if the target filesystem supports
+  FALLOC_FL_PUNCH_HOLE, leaving it false otherwise. */
```
satya-bodapati: Remove this duplicated comment here and everywhere.

storage/innobase/xtrabackup/src/datasink.cc
satya-bodapati: Done. Removed the zero-fill / `ds_create()` explanation comments in f4a0a742753 (force-pushed). The `MY_FAE | MY_ZEROFILL` flag and the in-class `= nullptr` / `= false` defaults stand on their own.

storage/innobase/xtrabackup/src/ds_buffer.cc
satya-bodapati: Done. Removed the zero-fill / `ds_create()` explanation comments in f4a0a742753 (force-pushed). The `MY_FAE | MY_ZEROFILL` flag and the in-class `= nullptr` / `= false` defaults stand on their own.

storage/innobase/xtrabackup/src/ds_local.cc
satya-bodapati: Done. Removed the zero-fill / `ds_create()` explanation comments in f4a0a742753 (force-pushed). The `MY_FAE | MY_ZEROFILL` flag and the in-class `= nullptr` / `= false` defaults stand on their own.

satya-bodapati:
Follow-up on c935a9d7a17 (force-pushed): dropped the redundant `packed_len` loop in `ds_write_sparse()`. The `len` parameter is already the packed (hole-excluded) payload size — callers pre-pack the buffer (`dst_pos` in `write_ibd_buffer`, `chunk.length` in `xbstream.cc`) and `local_write_sparse` walks the sparse_map writing exactly that many bytes in total. Now we just `add_uncompressed(len)`. Verified with both t/backup_size_basic.sh and t/backup_size_compress.sh (byte-exact size assertions).

storage/innobase/xtrabackup/src/backup_mysql.cc
```diff
@@ -1849,6 +1849,11 @@ char *get_xtrabackup_info(MYSQL *connection) {
   format_time(history_start_time, buf_start_time, time_buf_size);
   format_time(history_end_time, buf_end_time, time_buf_size);
 
+  /* Sampled here, after all data has drained to the leaf (see the
+  reordering in xtrabackup_backup_func()).  Embedded directly into
+  xtrabackup_info so consumers see the final on-disk byte count. */
+  const unsigned long long backup_size = get_final_backup_size();
+
```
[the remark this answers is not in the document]
satya-bodapati: This is known limitation. we write backup_size in xtrabacup_info and xtrabackup_info bytes are counted. So we differ by few bytes. This is OK.

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1594,7 +1598,51 @@ bool backup_start(Backup_context &context) {
...
+    return;
+  }
+
+  xb::info() << "Uncompressed backup size: "
+             << xtrabackup::utils::human_readable(uncompressed_backup_size)
+             << " (" << uncompressed_backup_size << " bytes)";
+
+  const double ratio = (double)uncompressed_backup_size / (double)backup_size;
+  xb::info() << "Compression ratio: " << std::fixed << std::setprecision(2)
+             << ratio << "x";
```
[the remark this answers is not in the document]
satya-bodapati: Fixing commit message.

storage/innobase/xtrabackup/src/datasink.h
```diff
@@ -30,25 +34,89 @@ extern "C" {
...
+   Kept lexically distinct from the tier-1 ds_metric / report_metrics /
+   ds_find_metric system: that one describes per-datasink observations,
+   this one is a single backup-run aggregate.  No shared names.
+   --------------------------------------------------------------------- */
+
+struct xb_uncomp_bytes {
+  /** Record the raw byte count of one top-level ds_write /
+  ds_write_sparse.  Called by the datasink framework when the
+  issuing ds_file_t has tracking enabled. */
+  ALWAYS_INLINE void add_uncompressed(uint64_t n) {
```
jakub-nowakowski-percona, line 57: Is it important to force in-lining here?

storage/innobase/xtrabackup/src/datasink.h
```diff
@@ -30,25 +34,89 @@ extern "C" {
...
+ private:
+  std::atomic<uint64_t> uncompressed_bytes_{0};
+};
+
 typedef struct ds_ctxt {
   datasink_t *datasink;
   char *root;
   void *ptr;
-  struct ds_ctxt *pipe_ctxt;
+  struct ds_ctxt *pipe_ctxt = nullptr;
```
jakub-nowakowski-percona, line 75: We could null-initialize other pointers in this structure too.

storage/innobase/xtrabackup/src/datasink.h
```diff
@@ -30,25 +34,89 @@ extern "C" {
...
+   different datasink's output and denotes a different metric there;
+   queries are therefore always scoped to a specific node (see
+   ds_find_metric() below) rather than flattened across a pipeline.
+   --------------------------------------------------------------------- */
+
+/** A single datum published by a datasink's report_metrics.  name is a
+stable string literal owned by the datasink; value is a snapshot at
+read time. */
+struct ds_metric {
+  const char *name;
```
jakub-nowakowski-percona: Instead of explaining ownership of `name` in a comment, we could use `std::string_view`, which should make it unambiguous.

storage/innobase/xtrabackup/src/datasink.h
```diff
@@ -59,6 +127,12 @@ struct datasink_struct {
                       bool punch_hole_supported);
   int (*close)(ds_file_t *file);
   void (*deinit)(ds_ctxt_t *ctxt);
+  /* Optional: append zero or more {name, value} entries describing
+  this datasink's internal state.  NULL means "no metrics".  Only the
+  three leaf datasinks populate this today (to publish
+  "bytes_written"); wrappers leave it null and future wrappers can
+  opt in by implementing it without any framework change. */
+  void (*report_metrics)(const ds_ctxt_t *ctxt, std::vector<ds_metric> *out);
```
jakub-nowakowski-percona: Should `out` ever be null? If not, we could make it a reference. Unless it's part of some convention. If so, please ignore.

storage/innobase/xtrabackup/src/datasink.h
```diff
@@ -118,6 +217,28 @@ Set the destination pipe for a datasink (only makes sense for compress and
...
+first entry whose name equals @p name.  The lookup is scoped to a
+single node: pass ds_leaf(head) (or any other specific ctxt) to
+disambiguate when multiple datasinks in a pipeline publish metrics.
+
+Safe to call when the node is quiesced (e.g. from backup_finish).
+Returns false (and leaves *out untouched) if @p node is null, the
+datasink does not implement report_metrics, or the name is not
+published.
+@param[in]  node  a specific datasink ctxt (not a pipeline head)
+@param[in]  name  metric name to look up (strcmp comparison)
```
jakub-nowakowski-percona: Putting implementation details into interface docs seems a bit off to me (`strcmp comparison`). Also, `name` could be `std::string_view`.

storage/innobase/xtrabackup/src/ds_buffer.cc
```diff
@@ -65,9 +66,9 @@ static ds_ctxt_t *buffer_init(const char *root) {
   ds_ctxt_t *ctxt;
   ds_buffer_ctxt_t *buffer_ctxt;
 
-  ctxt = static_cast<ds_ctxt *>(
-      my_malloc(PSI_NOT_INSTRUMENTED,
-                sizeof(ds_ctxt_t) + sizeof(ds_buffer_ctxt_t), MYF(MY_FAE)));
+  ctxt = static_cast<ds_ctxt *>(my_malloc(
+      PSI_NOT_INSTRUMENTED, sizeof(ds_ctxt_t) + sizeof(ds_buffer_ctxt_t),
+      MYF(MY_FAE | MY_ZEROFILL)));
```
jakub-nowakowski-percona, line 71: Is `MY_ZEROFILL` needed here? I notice there are more changes like that made to `my_malloc` calls.

jakub-nowakowski-percona (approved):
Approved with optional remarks.

storage/innobase/xtrabackup/src/ds_buffer.cc
satya-bodapati: This is to ensure that when don't look at some uninitialized values of the members

storage/innobase/xtrabackup/src/datasink.h
satya-bodapati: not very important. but since we are going to do backup_size calculation always, trying to reduce the overhead of funciton calls to count backup size as much as possible.

storage/innobase/xtrabackup/src/datasink.h
satya-bodapati: Done in 8d20d4bc94b: every remaining pointer in ds_ctxt_t (datasink, root, ptr) and all five in ds_file_t (ptr, path, datasink, ctxt, uncomp_bytes) are now default-initialised to nullptr in-class. Newly constructed handles no longer rely on the allocation site remembering MY_ZEROFILL.

storage/innobase/xtrabackup/src/datasink.h
satya-bodapati: Done in 8d20d4bc94b: ds_metric::name is now std::string_view, producers unchanged (string literals convert implicitly), and the ownership sentence is gone — the type enforces the non-owning contract.

storage/innobase/xtrabackup/src/datasink.h
satya-bodapati: Done in 8d20d4bc94b: report_metrics' out parameter is now std::vector<ds_metric> &. Updated the three leaf implementations (local, stdout, fifo) to use out.push_back(...) accordingly; the single caller (ds_find_metric) passes a local stack vector.

storage/innobase/xtrabackup/src/datasink.h
satya-bodapati: Done in 8d20d4bc94b: the ds_find_metric @param name doc drops the '(strcmp comparison)' phrase, the name parameter is now std::string_view, and the body uses m.name == name. <cstring> dropped from datasink.cc — no strcmp left in this path.

satya-bodapati:
Follow-up commit **8d20d4bc94b — PXB-3747 : Address Jakub's review comments** (will squash before merge).

What it addresses from Jakub's review round:

- **J2 (`datasink.h` pointer init)** — every remaining pointer member in `ds_ctxt_t` (`datasink`, `root`, `ptr`) and all five in `ds_file_t` (`ptr`, `path`, `datasink`, `ctxt`, `uncomp_bytes`) is now default-initialised to `nullptr` in-class. Newly constructed handles no longer depend on the allocation site remembering `MY_ZEROFILL` / explicit field init.
- **J3 + J5 (`ds_metric::name` and `ds_find_metric` doc)** — `ds_metric::name` and `ds_find_metric`'s `name` parameter are now `std::string_view`. The ownership comment and the "(strcmp comparison)" implementation-leak are gone. `ds_find_metric` uses `m.name == name`; `strcmp` / `<cstring>` removed from `datasink.cc`. Producers stay the same — string literals convert implicitly.
- **J4 (`report_metrics` out param)** — the vtable slot is now `std::vector<ds_metric> &out`. The three leaf implementations (`ds_local`, `ds_stdout`, `ds_fifo`) use `out.push_back(...)` accordingly; the single caller (`ds_find_metric`) passes a stack-local vector. Other vtables' `nullptr /* report_metrics */` placeholders are unaffected.

Left as-is with the existing replies standing: J1 (`ALWAYS_INLINE`) and J6 (`MY_ZEROFILL`).

C1 (Compression ratio `3.70` vs `3.70x`): the PR description already shows `3.70x`; the Commit-2 message example will be fixed at squash time.

Verified: Debug rebuild clean; `t/backup_size_basic.sh` and `t/backup_size_compress.sh` both pass (byte-exact size assertions).

satya-bodapati:
Squashed the two follow-up commits into the two base commits. PR now shows exactly the intended two-commit history:

- **428d631ce2f** PXB-3747 : Report final on-disk backup_size in xtrabackup_info
- **1d57a2c7e9d** PXB-3747 : Report uncompressed_backup_size for --compress backups

What was folded:

- **c935a9d** ("Addressing review comments") split by hunk:
  - `report_backup_size()` doc reflow (`backup_copy.cc`), `ds_write_sparse` packed_len -> len opt (`datasink.cc`), and the "Finsh" -> "Finish" typo fix on `backup_finish()` -> Commit 2 (the typo fix's hunk context lives inside `report_backup_size`, which is Commit-2 territory).
  - `pipe_ctxt = nullptr` default, `ds_create()` cleanup, `MY_ZEROFILL` + xbstream `fs_support_punch_hole = true` + `local_close` trailing-byte accounting, `sum_file_bytes -d` guard -> Commit 1.
- **8d20d4b** ("Address Jakub's review comments") split by hunk:
  - `ds_file_t::uncomp_bytes = nullptr` default -> Commit 2 (the field is introduced by Commit 2).
  - All other `= nullptr` defaults, `ds_metric::name` and `ds_find_metric` -> `std::string_view`, `report_metrics` out-param -> reference -> Commit 1.

Commit 2's "How to use this feature" section now shows `Compression ratio: 3.70x` (was `3.70`), addressing Copilot's earlier flag.

Verification (byte-exact against the pre-fold tip):

```
$ git diff <new-tip> <pre-fold-safety-tag> -- storage/ test/ | wc -l
0
```

Re-ran both strict-size suites against the new history (Debug build, `MySQL 8.4.8 (InnoDB 8.4.8)` server):

- `t/backup_size_basic.sh`: passed.
- `t/backup_size_compress.sh`: passed (covers `--compress`, `--compress + --encrypt`, `--compress + --stream=xbstream`, `--extra-lsndir`, sparse, etc., all asserting byte-exact `backup_size` and `uncompressed_backup_size`).

Earlier per-thread review comments (Copilot's C1/C2; Jakub's J1/J2/J3/J4/J5/J6) reference commit IDs that no longer exist after the squash, so GitHub will mark them as outdated. The actual changes those comments were about are all in the two final commits.

PR #1740, opened 2026-04-23
PXB-3762: xtrabackup --backup --transition-key crashes when undo tablespace key exists in redo only

JoeJRW:
## Summary

Fix assertion `ut_a(found)` in `xb_set_encryption()` during
`xtrabackup --backup --transition-key=...`.

JIRA URL: https://perconadev.atlassian.net/browse/PXB-3762

## Background: what the keys map is for

xtrabackup uses one in-memory map (`encryption_info`) in two phases:

* **`--backup` - producer.** Keys seen in redo are *inserted* into the
  map and written out to `xtrabackup_keys` at end-of-backup.
* **`--prepare` - consumer.** `xtrabackup_keys` is *loaded* back into
  the map and looked up via `xb_fetch_tablespace_key()` to install
  encryption on each tablespace.

`xb_set_encryption()` is therefore a prepare-only helper. Calling it
during backup is wrong by design,  the map is still being filled, so
the lookup must fail.

## Root cause

`srv_undo_tablespace_read_encryption()` has three branches for loading
the undo tablespace key. Condition 2 correctly gates the prepare-only
path with `(use_dumped_tablespace_keys && !srv_backup_mode)`.
Condition 1 (redo holds a key newer than the on-disk undo page) is
missing the `!srv_backup_mode` half of that guard.

`--transition-key` sets `use_dumped_tablespace_keys = true` in both
backup and prepare. So during `--backup`, when redo contains a recent
master-key-rotation entry for the undo space, Condition 1 takes the
prepare-only path, the lookup finds nothing, and `ut_a(found)` fires.

## Fix

Mirror Condition 2's guard in Condition 1: only take the dump-load
path when `!srv_backup_mode`. During backup we now use the key we
already have right there in `recv_key` — which is also exactly what
we want, because backup's job is just to collect that key into the
map for later dumping.

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -392,7 +392,7 @@ static dberr_t srv_undo_tablespace_read_encryption(pfs_os_file_t fh,
 
   if (found && recv_key->lsn >= page_lsn) {
     // Condition 1: Use key from redo if available and appropriate
-    encryption_success = use_dumped_tablespace_keys
+    encryption_success = (use_dumped_tablespace_keys && !srv_backup_mode)
                              ? load_key_from_dump()
                              : set_encryption(recv_key->ptr, recv_key->iv);
   } else if (is_enc) {


satya-bodapati:
Hi @JoeJRW — thanks for the fix and the deterministic reproducer!

I've updated the PR title and description to lead with the producer/consumer framing of the `encryption_info` map, since that's what makes the bug "click" — `xb_set_encryption()` is a prepare-time consumer, and Condition 1 was calling it during the backup-time producer phase.

To keep the merged commit log in sync with the updated description, could you `git commit --amend` with the message below and `git push --force-with-lease`? I'll merge as soon as it lands.

```
PXB-3762: xtrabackup --backup --transition-key crashes when undo tablespace key exists in redo only
JIRA URL: https://perconadev.atlassian.net/browse/PXB-3762

The encryption_info keys map has two phases:
  - --backup  : producer. Keys are inserted from redo and dumped to
                xtrabackup_keys at end-of-backup.
  - --prepare : consumer. xtrabackup_keys is loaded into the map and
                looked up to set up encryption on each tablespace.

xb_set_encryption() / xb_fetch_tablespace_key() are the consumer side
and must only run at prepare time. Condition 2 in
srv_undo_tablespace_read_encryption() gates this correctly with
(use_dumped_tablespace_keys && !srv_backup_mode); Condition 1 was
missing the !srv_backup_mode half.

--transition-key sets use_dumped_tablespace_keys=true in both phases,
so when redo carried an undo-encryption key newer than the on-disk
undo page, --backup entered the consumer path against an empty map
and ut_a(found) fired.

Add !srv_backup_mode to Condition 1 to match Condition 2. During
backup we now use the key already in recv_key (which is what gets
inserted into the map anyway).
```

Thanks again!

JoeJRW:
@satya-bodapati Done — I amended the commit message with the text above and pushed it using `--force-with-lease`.

Thanks again for the review!

satya-bodapati:
@JoeJRW I managed to create a testcase that will fail with release build of server. Can you please modify your commit to remove testcase? I will add testcase as a followup commit

JoeJRW:
@satya-bodapati Done.

satya-bodapati:
Awesome, Thank you @JoeJRW! I will merge this PR as soon as github checks are finished.

PR #1742, opened 2026-04-28
PXB-3568 : xtrabackup crashes on RHEL9 with --compress=lz4

satya-bodapati:
JIRA URL: https://perconadev.atlassian.net/browse/PXB-3568

Problem:
xtrabackup crashes with SIGABRT (Signal 6) during a backup with --compress=lz4 on EL9-based images (RHEL 9, OracleLinux 9, Rocky 9) when the database is under concurrent write load. The crash also manifests as an SST failure in PXC clusters running PXC 8.0.42 and 8.0.45 Docker images. Without _GLIBCXX_ASSERTIONS the same condition silently corrupts heap memory instead of aborting.

Analysis:
The three compression datasinks - ds_compress.cc (qpress), ds_compress_lz4.cc, and ds_compress_zstd.cc - each defined structs named comp_thread_ctxt_t, ds_compress_ctxt_t, and ds_compress_file_t with different layouts. This violates the C++ One Definition Rule (ODR), which says a named type must have the same definition in every translation unit that references it.

Without LTO, every translation unit was compiled independently and saw only its own local definition, so no conflict surfaced at runtime. With LTO, GCC merges all translation units into a single intermediate representation at link time, detects the same-named structs with different layouts, picks one as canonical, and uses it for code generation across all callers.

GCC picked the qpress version of comp_thread_ctxt_t, which embeds a qlz_state_compress hash table and is 36920 bytes (0x9038). The LZ4 version is 40 bytes (0x28). As a result, the LZ4 compress_write() emitted the qpress 36920-byte stride when indexing its std::vector<comp_thread_ctxt_t>:

  imul $0x9038, %r12, %rax   ; per-element stride: WRONG (should be 0x28)

On the second iteration of the loop, the wrong stride produced an out-of-bounds offset, which tripped the std::vector::operator[] bounds-check assertion enabled by _GLIBCXX_ASSERTIONS (default on EL9 via redhat-rpm-config) and aborted.

GCC was already emitting -Wodr warnings flagging these exact structs in every build, but they were missed amid the build log noise.

Fix:
Rename the file-local types in each compression datasink with a compressor-specific prefix so every translation unit has a unique type identity, eliminating the ODR violation:

  ds_compress.cc (qpress): comp_thread_ctxt_t -> qpress_thread_ctxt_t
                           ds_compress_ctxt_t -> qpress_compress_ctxt_t
                           ds_compress_file_t -> qpress_compress_file_t
  ds_compress_lz4.cc     : comp_thread_ctxt_t -> lz4_thread_ctxt_t
                           ds_compress_ctxt_t -> lz4_compress_ctxt_t
                           ds_compress_file_t -> lz4_compress_file_t
  ds_compress_zstd.cc    : ds_compress_ctxt_t -> zstd_compress_ctxt_t
                           ds_compress_file_t -> zstd_compress_file_t

After the rename:
 * The three -Wodr warnings for these types are gone from the build.
 * LZ4 compress_write uses the correct $0x28 (40) stride.
 * qpress compress_write uses $0x9038 (36920), which is also correct because qpress's struct really is that size.
 * Full backup + decompress + --prepare + --copy-back + live SELECT round-trip passes on Ubuntu 22.04 and OracleLinux 9 with GCC 12.3
   + LTO + _GLIBCXX_ASSERTIONS.

### DIFF
--- a/storage/innobase/xtrabackup/src/ds_compress.cc
+++ b/storage/innobase/xtrabackup/src/ds_compress.cc
@@ -41,21 +41,21 @@ typedef struct {
   size_t to_size;
   uint32_t adler;
   qlz_state_compress state;
-} comp_thread_ctxt_t;
+} qpress_thread_ctxt_t;
 
 typedef struct {
   Thread_pool *thread_pool;
-} ds_compress_ctxt_t;
+} qpress_compress_ctxt_t;
 
 typedef struct {
   ds_file_t *dest_file;
-  ds_compress_ctxt_t *comp_ctxt;
+  qpress_compress_ctxt_t *comp_ctxt;
   size_t bytes_processed;
   char *comp_buf;
   size_t comp_buf_size;
   std::vector<std::future<void>> tasks;
-  std::vector<comp_thread_ctxt_t> contexts;
-} ds_compress_file_t;
+  std::vector<qpress_thread_ctxt_t> contexts;
+} qpress_compress_file_t;
 
 /* Compression options */
 extern char *xtrabackup_compress_alg;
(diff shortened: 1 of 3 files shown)


jakub-nowakowski-percona (approved):

satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-multijob/399/
All OEL9 Release build failures with compress=lz4 are gone..

PR #1748, opened 2026-04-30
PXB-3759 : xtrabackup crashes if server uses redo log encryption and

satya-bodapati:
innodb_log_buffer_size > 16MB

https://perconadev.atlassian.net/browse/PXB-3759

Problem
=======

xtrabackup crashes with SIGSEGV at end of backup when the source server has innodb_redo_log_encrypt=ON and innodb_log_buffer_size
> 32 MiB.

Analysis
========

Redo_Log_Writer::scratch_buf is hardcoded to 16 MiB, while the matching source buffer scales with srv_log_buffer_size / 2. With innodb_log_buffer_size=64M a single write_buffer() call may pass 32 MiB to Encryption::encrypt_log(), which copies that many bytes into scratch_buf without bounds-checking the destination. The overflow corrupts adjacent heap allocations; the crash typically surfaces later when freeing an allocation whose PFS metadata header was corrupted.

Fix
===

Size scratch_buf to redo_log_read_buffer_size so it always matches the largest len write_buffer() can receive. Add a release-build assertion against the same bound to prevent future drift between the two sizes.

Add suites/keyring/pxb-3759.sh as a regression test. It uses the xtrabackup_pause_after_redo_catchup debug-sync to deterministically build up >16 MiB of pending redo before resuming the backup.

### DIFF
--- a/storage/innobase/xtrabackup/src/redo_log.cc
+++ b/storage/innobase/xtrabackup/src/redo_log.cc
@@ -698,7 +698,7 @@ bool Redo_Log_Parser::parse_log(const byte *buf, size_t len, lsn_t start_lsn) {
 
 Redo_Log_Writer::Redo_Log_Writer() {
   scratch_buf.alloc_withkey(UT_NEW_THIS_FILE_PSI_KEY,
-                            ut::Count{16 * 1024 * 1024});
+                            ut::Count{redo_log_read_buffer_size});
 }
 
 bool Redo_Log_Writer::create_logfile(const char *name) {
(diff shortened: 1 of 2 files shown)


jakub-nowakowski-percona (approved):

satya-bodapati:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/498

xtradb platform shows failure because I used version as 8.0.455-36 instead of 8.0.45-36

Introduced test passed on all platforms

PR #1751, opened 2026-05-11
PXB-3768 : Fix 8.0 test failures on Jenkins

satya-bodapati:
https://perconadev.atlassian.net/browse/PXB-3768

Three of the redo-related tests waited for the InnoDB log governor to reclaim redo files behind a registered xtrabackup consumer. On loaded Jenkins workers, none of the governor's reclaim conditions reliably fires for the test's workload, so the wait runs to timeout. Trigger one of those conditions explicitly and tighten the synchronisation.

Fixes:
---------
1. t/pxb-3003-redo-log-purge-race.sh: Start at innodb_redo_log_capacity=16M and shrink to 8M after xtrabackup pauses. The shrink sets target < current capacity, which forces the governor to reclaim files behind the PXB consumer immediately. Add a PFS precondition that >= 2 redo files exist before backup starts and dump PFS state at suspend for diagnosability.

2. t/redo_log_consumer.sh: Anchor the wait grep on the consumer name (PXB or MEB) so the prior backup's log_checkpointer warnings don't false-match. Add a PFS fallback (>= 3 redo files implies the consumer is holding redo back) and bound the wait to 60s with a simple for-loop. On timeout, dump PFS and the mysqld error file tail and call die() so MTR fails with diagnostics instead of hanging until Jenkins kills the job. Also stop the background inserter before stop_server to avoid ERROR 2002 spam.

3. t/estimate_memory.sh: - Drop the obsolete 8.0.44 version guard (smart memory has been GA for some time). - Tee xtrabackup --prepare stderr to both the log file and stderr so failures are visible in the MTR result instead of producing an empty log on abrupt failure. This is what surfaced the smart-memory underflow described below. - Free-memory gate now uses min(MemAvailable, MemFree) and logs both the gate inputs and the kernel's MemFree / MemAvailable just before xtrabackup --prepare runs, so any result file shows the OS view next to xtrabackup's "Free memory: ..." line. Background: PXB-3770 - estimate- memory is broken on libproc2 (procps-ng 4.x, e.g. Ubuntu 24.04) builds: host_free_memory() misuses the libproc2 API and returns 0 even when MemAvailable is tens of GiB; the smart-memory code then unsigned-underflows the buffer pool size and SIGSEGVs InnoDB. The gate masks the crash on hosts with low MemFree; the diagnostic prints make any remaining failure self-explanatory until PXB-3770 lands.

4. t/PXB-2854_xbstream_memory_corruption.sh: Add require_qpress so the test skips cleanly on workers where the standalone qpress binary used to build the .qp fixture is not installed, instead of failing.

Related:
- PXB-3770 (xtrabackup binary fix for the libproc2 API misuse; branch + patch on the ticket).

### DIFF
--- a/storage/innobase/xtrabackup/test/t/PXB-2854_xbstream_memory_corruption.sh
+++ b/storage/innobase/xtrabackup/test/t/PXB-2854_xbstream_memory_corruption.sh
@@ -1,4 +1,8 @@
 # PXB-2854 - Quicklz decompression memory corruption issue
+# xbstream still decompresses legacy .qp streams; skip cleanly when
+# the standalone qpress binary used to build the fixture is missing.
+
+require_qpress
 
 head -c 100000 </dev/urandom > $topdir/payload.bin
 qpress $topdir/payload.bin $topdir/payload.qp
--- a/storage/innobase/xtrabackup/test/t/estimate_memory.sh
+++ b/storage/innobase/xtrabackup/test/t/estimate_memory.sh
@@ -1,8 +1,3 @@
-if is_server_version_higher_than 8.0.44
-then
-    die "We should check if smart memory is mature to transition to GA."
-fi
-
 . inc/common.sh
 require_debug_pxb_version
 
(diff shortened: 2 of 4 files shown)


satya-bodapati:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param/504/#showFailuresLink

jakub-nowakowski-percona (approved):

PR #1752, opened 2026-05-13
PXB-3768 [8.4]: Fix test failures on Jenkins

satya-bodapati:
Also add testcase for PXB-3762


satya-bodapati:
testcase for #1740 is added in this PR

jakub-nowakowski-percona (approved):

PR #1767, opened 2026-07-03
Fix restore crash when Smart memory estimation is used

alexole:
The `libproc2` variants of `host_free_memory()`/`host_total_memory()` declare `struct meminfo_info *mem_info;` uninitialized and pass it to `procps_meminfo_new(&mem_info)`. `procps_meminfo_new()` rejects a non-`NULL` `*info` with `-EINVAL`:

```c++
    // procps-ng library/meminfo.c, procps_meminfo_new()
    if (info == NULL || *info != NULL)
        return -EINVAL;
```

If the uninitialized stack slot holds a non-`NULL` value the call fails, the `if (rc < 0) return 0;` guard fires, and the function reports `0` bytes free.

With `--prepare --use-free-memory-pct` that `0` leads to a negative buffer pool size and crashes `--prepare` (exit 2). Initialize the pointer to `nullptr`, and release the handle with `procps_meminfo_unref()` before returning.

### DIFF
--- a/storage/innobase/xtrabackup/src/utils.cc
+++ b/storage/innobase/xtrabackup/src/utils.cc
@@ -169,12 +169,14 @@ unsigned long host_total_memory() {
   meminfo();
   return kb_main_total * 1024;
 #else
-  struct meminfo_info *mem_info;
+  struct meminfo_info *mem_info = nullptr;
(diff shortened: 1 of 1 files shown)


satya-bodapati:
https://perconadev.atlassian.net/browse/PXB-3881

PR #1774, opened 2026-07-31
PXB-3502 [8.4]: Process largest files first in all parallel phases

jakub-nowakowski-percona:
PXB-3502: Process largest files first in all parallel phases

https://perconadev.atlassian.net/browse/PXB-3502

Problem
=======

When xtrabackup processes files in parallel (backup, copy-back,
decompress, incremental apply), the work queue is FIFO. If the
largest tablespace or SST file happens to be discovered last, it
starts processing after all other threads have finished their
smaller files, resulting in a single thread running alone for the
duration of that large file. This "tail latency" problem makes
parallel restores and backups slower than necessary.

Fix
===

Replace FIFO scheduling with largest-first scheduling across all
parallel processing phases so that the longest-running task always
starts immediately, minimizing wall-clock time.

Affected scenarios (all use --parallel):

1. xtrabackup --backup: InnoDB tablespace copying is now ordered
   largest-first via sorted datafiles_iter_t.

2. xtrabackup --backup (RocksDB): SST and meta file copying is
   now ordered largest-first via queue-based work-stealing.

3. xtrabackup --backup --stream=xbstream: files are written to
   the stream in largest-first order, so downstream xbstream
   extraction and decompression inherits that order.

4. xtrabackup --prepare --incremental-dir (delta apply): InnoDB
   .delta files are applied largest-first via the priority queue.

5. xtrabackup --prepare --incremental-dir (RocksDB): SST files
   from the incremental directory are copied/moved largest-first.

6. xtrabackup --copy-back / --move-back: all data files (InnoDB,
   RocksDB, non-InnoDB) are processed largest-first.

7. xtrabackup --decompress: compressed files are decompressed
   largest-first via the priority queue.

Implementation
==============

1. datadir_queue (backup_copy.cc): Convert the internal std::queue
   to std::priority_queue with a comparator that orders entries by
   descending file_size. Tie-break on path for determinism. Empty
   directories get lowest priority.

2. datafiles_iter_t (xtrabackup.cc): Sort the tablespace node
   vector by descending byte size (page count * physical page size)
   at the end of datafiles_iter_new().

3. RocksDB (backup_copy.cc): Replace par_for()-based contiguous
   slicing with a queue-based work-stealing approach. RocksDB data
   and meta file lists are merged into a single datadir_queue so
   that all files compete globally for largest-first scheduling.
   Extract run_worker_threads() as a common template for thread
   lifecycle management shared between run_data_threads() and
   run_rocksdb_threads().

4. File size propagation: Populate file_size in datadir_entry_t
   during directory scanning (xb_process_datadir via stat()) and
   in Myrocks_datadir::scan_dir().

5. Log enhancement: Print both raw byte size and human-readable
   size in file copy/move log messages.

Tests
=====

Five test scripts verify ordering with --parallel=1 (which makes
scheduling deterministic):

- t/pxb_largest_first_innodb.sh: InnoDB backup and copy-back
- t/pxb_largest_first_delta.sh: InnoDB incremental delta apply
- suites/rocksdb/largest_first.sh: RocksDB backup and copy-back
- suites/rocksdb/largest_first_incremental.sh: RocksDB incremental
- suites/compression/largest_first_decompress.sh: LZ4 decompression

### DIFF
--- a/storage/innobase/xtrabackup/src/backup_copy.cc
+++ b/storage/innobase/xtrabackup/src/backup_copy.cc
@@ -60,6 +60,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <set>
 #include <sstream>
 #include <string>
+#include <vector>
 #include "changed_page_tracking.h"
 #include "common.h"
 #include "fil_cur.h"
@@ -104,16 +105,39 @@ xb_regex_t preg_filepath;
 static std::set<std::string> skip_copy_back_list;
 
 class datadir_queue {
-  std::queue<datadir_entry_t> queue;
+  // Comparator sorting datadir entries by their size.
+  struct queued_datadir_entry_compare {
+    bool operator()(const datadir_entry_t &lhs,
+                    const datadir_entry_t &rhs) const {
+      // Empty directories are considered to be smaller than files.
+      if (lhs.is_empty_dir != rhs.is_empty_dir) {
+        return lhs.is_empty_dir;
+      }
+
+      auto lhs_size = lhs.file_size >= 0 ? lhs.file_size : -1;
+      auto rhs_size = rhs.file_size >= 0 ? rhs.file_size : -1;
+
+      if (lhs_size != rhs_size) {
+        return lhs_size < rhs_size;
+      }
+
+      // If files are same size, order them by path to keep sorting
+      // deterministic.
+      return lhs.path > rhs.path;
+    }
+  };
+
+  std::priority_queue<datadir_entry_t, std::vector<datadir_entry_t>,
+                      queued_datadir_entry_compare>
+      queue;
   mysql_mutex_t mutex;
   mysql_cond_t cond;
-  bool final;
+  bool final = false;
 
  public:
   datadir_queue() {
     mysql_mutex_init(0, &mutex, MY_MUTEX_INIT_FAST);
     mysql_cond_init(0, &cond);
-    final = false;
   }
   ~datadir_queue() {
     mysql_cond_destroy(&cond);
@@ -140,7 +164,7 @@ class datadir_queue {
       mysql_mutex_unlock(&mutex);
       return false;
     }
-    entry = queue.front();
+    entry = queue.top();
     queue.pop();
     mysql_mutex_unlock(&mutex);
     return true;
@@ -424,9 +448,14 @@ bool backup_file_printf(const char *filename, const char *fmt, ...) {
   return (result);
 }
 
+/************************************************************************
+Common body for parallel thread execution. Takes a pre-populated and
+completed datadir_queue and spawns N worker threads that each receive a
+datadir_thread_ctxt_t*. Waits for all threads to finish and returns
+true if every thread succeeded. */
 template <typename F>
-bool run_data_threads(const char *dir, const char *suffix, F func, uint n,
-                      const char *thread_description) {
+bool run_worker_threads(datadir_queue &queue, F func, uint n,
+                        const char *thread_description) {
   datadir_thread_ctxt_t *data_threads;
   uint i, count;
   ib_mutex_t count_mutex;
@@ -439,25 +468,14 @@ bool run_data_threads(const char *dir, const char *suffix, F func, uint n,
   mutex_create(LATCH_ID_XTRA_COUNT_MUTEX, &count_mutex);
   count = n;
 
-  datadir_queue queue;
-
   for (i = 0; i < n; i++) {
     data_threads[i].n_thread = i + 1;
     data_threads[i].count = &count;
     data_threads[i].count_mutex = &count_mutex;
     data_threads[i].queue = &queue;
+    data_threads[i].ret = true;
     os_thread_create(PFS_NOT_INSTRUMENTED, 0, func, &data_threads[i]).start();
   }
-
-  xb_process_datadir(
-      dir, suffix,
-      [&](const datadir_entry_t &entry, void *arg) mutable -> bool {
-        queue.push(entry);
-        return true;
-      },
-      nullptr);
-
-  queue.complete();
 
   /* Wait for threads to exit */
   while (1) {
@@ -482,7 +500,25 @@ bool run_data_threads(const char *dir, const char *suffix, F func, uint n,
 
   ut::free(data_threads);
 
-  return (ret);
+  return ret;
+}
+
+template <typename F>
+bool run_data_threads(const char *dir, const char *suffix, F func, uint n,
+                      const char *thread_description) {
+  datadir_queue queue;
+
+  xb_process_datadir(
+      dir, suffix,
+      [&](const datadir_entry_t &entry, void *arg) mutable -> bool {
+        queue.push(entry);
+        return true;
+      },
+      nullptr);
+
+  queue.complete();
+
+  return run_worker_threads(queue, func, n, thread_description);
 }
 
 // Explicit instantiation for function pointer type:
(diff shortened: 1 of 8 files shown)


satya-bodapati:
@jakub-nowakowski-percona I didn't fully check the PR yet. one small suggestion that might help our test validation. Print the file size when we print the copying file message. I think we already have this info either in fil_node or the rocksdb files too. When printing I would prefer human readable size (I think we already have such utiity funciton). Maybe print both (byte version & human-readable version). Bytes version is for grep/awk and to do sort..

This way, we can verify that the file sizes printed are in decreasing order of size.

jakub-nowakowski-percona:
@satya-bodapati I added file sizes to log messages.

jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/163/

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -625,9 +625,13 @@ bool copy_file(ds_ctxt_t *datasink, const char *src_file_path,
   action = xb_get_copy_action();
   if (pos >= 0) {
     xb::info() << action << " " << src_file_path << " to " << dstfile->path
-               << " up to position " << pos;
+               << " up to position " << pos << ", size " << pos << " ("
+               << xtrabackup::utils::human_readable(pos) << ")";
   } else {
```
satya-bodapati: aren't we using wrong variable for the  size? pos is only hte position at which the data is read.

jakub-nowakowski-percona: If `pos >= 0`, `cursor.statinfo.st_size` is set to `pos` on line 612 anyway. Effectively, these are equal here. By using `pos` I am trying to be more explicit about the fact that we're using the truncated size. But I can change it to `cursor.statinfo.st_size` if you prefer that.

satya-bodapati: yeah, its bit non obvious, better to use st_size imo

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -104,16 +105,39 @@ xb_regex_t preg_filepath;
...
+  struct queued_datadir_entry_compare {
+    bool operator()(const datadir_entry_t &lhs,
+                    const datadir_entry_t &rhs) const {
+      // Empty directories are considered to be smaller than files.
+      if (lhs.is_empty_dir != rhs.is_empty_dir) {
+        return lhs.is_empty_dir;
+      }
+
+      auto lhs_size = lhs.file_size >= 0 ? lhs.file_size : -1;
+      auto rhs_size = rhs.file_size >= 0 ? rhs.file_size : -1;
```
satya-bodapati, line 118: when can file_size be reported as < 0?

storage/innobase/xtrabackup/src/backup_copy.cc
```diff
@@ -1301,63 +1339,84 @@ Myrocks_checkpoint::file_list Myrocks_checkpoint::data_files() const {
...
-    if (!*result) {
-      break;
-    }
-  }
+    mutex_enter(ctx->count_mutex);
+    --(*ctx->count);
+    mutex_exit(ctx->count_mutex);
+    ctx->ret = ret;
+  };
+
```
satya-bodapati, line 1391: the ret value is ignored isn't it? if there is problem with rocksdb file copy, would it abort?

satya-bodapati: previuosly backup_rocksdb_files did check the result and break the loop.

storage/innobase/xtrabackup/test/t/pxb_largest_first_innodb.sh
```diff
@@ -0,0 +1,94 @@
...
+mysql -e "FLUSH TABLES" test
+
+# Record actual file sizes for debugging
+ls -lS $mysql_datadir/test/*.ibd
+
+###############################################################################
+# Test 1: Backup with --parallel=1 processes largest .ibd first
+###############################################################################
+
+xtrabackup --backup --target-dir=$topdir/backup --parallel=1
```
satya-bodapati, line 35: I wonder if we can add atleast few scenarios with --parallel=X
db1 with 100MB files 100
db2 with 10MB files 100
db3 with 1Mb files 100

from our new verification logic, even though it is parallel, 100MB files are picked up first, 10Mb files, then 1MB files?

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -604,6 +604,32 @@ datafiles_iter_t *datafiles_iter_new(
...
+                return node && node->name ? node->name : "";
+              };
+
+              const auto lhs_size = file_size(lhs);
+              const auto rhs_size = file_size(rhs);
+              if (lhs_size != rhs_size) {
+                return lhs_size > rhs_size;
+              }
+
+              return file_name(lhs) < file_name(rhs);
```
satya-bodapati, line 630: can we use a common function? I think we duplicated it a bit. the other comparator used similar logic but not fully equivalent?

storage/innobase/xtrabackup/src/backup_copy.cc
jakub-nowakowski-percona: I can see only one place which doesn't set `file_size` and uses default -1: https://github.com/jakub-nowakowski-percona/percona-xtrabackup/blob/PXB-3502-8.4/storage/innobase/xtrabackup/src/xtrabackup.cc#L5839

But this check is mostly defensive. `file_size` is a `ssize_t`, and by default it's set to -1. If for any reason `datadir_entry_t` with negative `file_size
` reaches this comparator, I want it to be handled properly.

storage/innobase/xtrabackup/test/t/pxb_largest_first_innodb.sh
satya-bodapati: I guess with --parallel, there might be timing issue that even though the thread picksup 100MB file due to context switches etc, file with 10MB could be printed first. :thinking: if you think that is the case, lets drop the idea of testing with --parallel=x

storage/innobase/xtrabackup/src/backup_copy.cc
satya-bodapati: ok, thanks for explaining

storage/innobase/xtrabackup/src/xtrabackup.cc
jakub-nowakowski-percona: I think the amount of code that could be shared between sorting comparator of `fil_node_t` and `datadir_entry_t` isn't worth the effort. Let's leave it as is.

storage/innobase/xtrabackup/test/t/pxb_largest_first_innodb.sh
jakub-nowakowski-percona: It could be hard to make a deterministic test for a scenario with multiple threads running in parallel. Let's leave this test as is, and I will do some more manual testing to verify file processing behaves as expected for multiple threads.

storage/innobase/xtrabackup/src/backup_copy.cc
jakub-nowakowski-percona: I changed the way RocksDB files are processed to follow the way we do it for InnoDB. The behavior for RocksDB may be slightly different now, but it's to be expected.

storage/innobase/xtrabackup/src/backup_copy.cc
jakub-nowakowski-percona: Done

satya-bodapati (approved):
LGTM! Please squash before merge. Please mention al lthe scenarios that this bug fix will affect in commit message.

PR #1777, opened 2026-08-13
PXB-3609 : [8.4] xbcloud delete leaves the .md5 file behind

satya-bodapati:
https://perconadev.atlassian.net/browse/PXB-3609

xbcloud put --md5 uploads the checksum file as <backup_name>.md5, next to the backup directory and not inside it. xbcloud delete lists only what is under <backup_name>/, so the .md5 file xbcloud generated was never deleted and stayed in the bucket after the backup itself was gone.

Delete <backup_name>.md5 along with the backup.

At delete time we do not know whether the backup was taken with --md5, so the delete is unconditional and delete_object() takes a best_effort flag for it: an object that is not there, or that we have no permission on, is not reported as an error. S3 answers 204 for a DELETE of a missing key while Azure and Swift answer 404, and a user whose rights are scoped to <backup_name>/* gets 403 for the .md5 file whether it exists or not. In all of those cases we stay quiet and leave things as they were before this fix. Real errors are still reported, and the backup itself is deleted as usual.

We do not check whether the file exists before deleting it, so that delete does not start requiring s3:GetObject - a retention user with only ListBucket and DeleteObject must still be able to remove a backup.

Tests: md5_delete, md5_delete_permissions, md5_delete_prefix_scope.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/azure.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/azure.cc
@@ -223,7 +223,7 @@ void Azure_client::set_endpoint(const std::string &ep, bool development_storage,
 }
 
 bool Azure_client::delete_object(const std::string &container,
-                                 const std::string &name) {
+                                 const std::string &name, bool best_effort) {
   Http_request req(Http_request::DELETE, protocol, host,
                    "/" + container + "/" + name);
   signer->sign_request(container, name, req, time(0));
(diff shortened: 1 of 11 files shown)


satya-bodapati:
https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/158/

jakub-nowakowski-percona (approved):

PR #1780, opened 2026-08-18
PXB-3862 : [8.4] Merge scattered changed-page reads in page-tracking backups

satya-bodapati:
PXB-3862 : Merge scattered changed-page reads in page-tracking backups

https://perconadev.atlassian.net/browse/PXB-3862

Problem
-------
A page-tracking incremental reads only the pages the server marked as
changed. Consecutive changed pages are already read together in one
request; the problem is scattered changes. If the changed pages are

    1, 3, 5, 7, 9

xtrabackup issues five single-page reads. Every read request costs a
full I/O round trip, so with scattered changes the copy phase is bound
by the number of requests instead of the amount of data, and an
incremental can take several times longer than a full scan of the
same file.

Fix
---
Read the whole range 1-9 in one request. The unchanged pages 2, 4, 6,
8 are read as filler and discarded by the existing incremental write
filter, which drops every page whose FIL_PAGE_LSN is older than
incremental_lsn. The backup's content and size are byte-for-byte
unchanged; only the read pattern changes.

A gap is worth combining across exactly when its bytes cost less than
one read request. That cost, in bytes of sequential transfer, is
round_trip * bandwidth - and no fixed number fits both a local NVMe
and a network volume, so it is measured at backup start
(probe_storage, xb_io_probe.h): twelve scattered single-page reads
give the round trip, 16MB of sequential reads the bandwidth, on the
largest changed data file (at least 64MB). Then

    read_request_cost = round_trip * bandwidth / 1.5,
                        clamped to [64KB, 1MB], 512KB if unmeasurable
    merge_gap         = read_request_cost / physical_page_size

The /1.5 margin absorbs copy-pipeline overhead and measurement noise,
erring toward reading less; it was calibrated on two instrumented
machines whose break-evens bound it from both sides. The
per-tablespace conversion lets compressed tablespaces combine across
the same byte cost. Filler bytes are always the actual gap sizes
present in the data, never the limit, so a generous limit reads
nothing extra.

--page-tracking-merge-gap exposes the behaviour: "auto" (default) as
described; a page count pins one value for all tables and skips the
probe; 0 keeps the previous strict-consecutive reads.

Log messages
------------
Once per backup, the measurement:

  pagetracking: calibrated storage (./test/t1.ibd): request round
  trip 124 us, sequential read 984 MB/s -> one read request costs
  ~83KB of sequential transfer; gaps cheaper than this are combined

Per table with at least 1000 changed pages, when its copy finishes,
accumulated from what was actually read:

  pagetracking: test/t1.ibd: 3196 changed pages in 3196 ranges (avg
  gap 2.0 pages); merge-gap=4 (auto) combined them into 2 reads:
  request reduction 1598.0x, read amplification 2.98x; issued 16
  read batches

"ranges" is the requests merge-gap=0 would issue; "request reduction"
is the benefit and "read amplification" its price (bytes read divided
by changed bytes - read volume only, backup size is unaffected);
"issued" exceeds the group count only when a group is larger than
--read-buffer-size and is read in buffer-sized pieces. When the
typical gap costs more than one read request, an extra line names
both numbers and the pinned value to try, so a boundary case is
diagnosable from the log alone:

  pagetracking: test/t1.ibd: typical gap 8.9 pages (143KB) costs
  more than one read request (83KB); reads stay individual - if
  sequential read throughput is high, --page-tracking-merge-gap=9 may
  be faster

Testing: unit tests (xb_page_group-t) cover the read request cost
model across device classes and the storage probe's failure modes; a
framework testcase sweeps change densities and asserts only
hardware-independent invariants - ranges vs combined reads vs issued
requests, auto vs strict, the cost floor and ceiling, and restore
correctness - and was verified to fail against a build with the
combining silently disabled.

---

Branch layout
-------------
- `PXB-3862 : Merge scattered changed-page reads in page-tracking backups` — the complete fix (review feedback from this PR squashed in: probe reworked in page terms with the sequential phase first, range_get_next_page() takes the read-filter context, std::optional<Probe_result>, human_readable() in the log lines, ERANGE check, documented invariants, named READ_REQUEST_COST_MARGIN, concise option help).
- `PXB-3862 : TEMPORARY: bundle the validation benchmark kit` — benchmark tooling for the ongoing lab validation; will be dropped before merge.

Note: the option was renamed from `--page-tracking-max-gap` to `--page-tracking-merge-gap` during review (same semantics: a page count, "auto" default, 0 = old behaviour).


storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1083,6 +1096,27 @@ struct my_option xb_client_options[] = {
...
+     "changed pages, that is ignored so that the pages are treated as one "
+     "continuous group and read together (the gap between changed pages 1 "
+     "and 4 is 2: the unchanged pages 2 and 3). Grouping avoids many small "
+     "individual reads when changed pages are scattered. The gap pages are "
+     "read but never written to the backup, so this affects read volume "
+     "only, not backup size. The default \"auto\" chooses the gap for each "
+     "table from the distribution of its changed pages, never combining "
+     "across a gap that costs more than one read request on the storage "
+     "(probed once per backup; 512KB when it cannot be measured). A number "
+     "(in "
```
jakub-nowakowski-percona, line 1110: Kinda weird line break.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -457,6 +458,17 @@ static ulonglong global_max_value;
 bool opt_galera_info = false;
 bool opt_slave_info = false;
 bool opt_page_tracking = false;
+/* --page-tracking-max-gap: "auto" (default) chooses the gap per table
+from the changed-page distribution; a number pins it for all tables. */
+bool opt_page_tracking_max_gap_auto = true;
```
jakub-nowakowski-percona, line 463: Do we want to run gap-measuring benchmark by default? I am wondering if this could have a negative effect on some workloads, e.g. databases with few gaps.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -1947,6 +1981,24 @@ bool xb_get_one_option(int optid, const struct my_option *opt, char *argument) {
...
       break;
 
+    case OPT_PAGE_TRACKING_MAX_GAP: {
+      if (strcasecmp(argument, "auto") == 0) {
+        opt_page_tracking_max_gap_auto = true;
+        break;
+      }
+      char *endp = nullptr;
+      ulonglong val = strtoull(argument, &endp, 10);
+      if (endp == argument || *endp != '\0' || val > 65536) {
```
jakub-nowakowski-percona, line 1991: This should work, but according to https://en.cppreference.com/cpp/string/byte/strtoul, we should also test for `errno == ERANGE`.

storage/innobase/xtrabackup/src/xtrabackup.cc
```diff
@@ -4254,6 +4306,74 @@ static void cleanup_mysql_environment() {
...
+  }
+
+  xb_read_request_cost = pagetracking::read_request_cost_bytes(
+      probe.rtt_us, probe.bw_bytes_per_sec);
+
+  xb::info() << "pagetracking: calibrated storage (" << probe_path
+             << "): request round trip " << probe.rtt_us
+             << " us, sequential read "
+             << probe.bw_bytes_per_sec / (1024 * 1024)
+             << " MB/s -> one read request costs ~"
```
jakub-nowakowski-percona, line 4371: I see that we use MiB in other places (e.g. in `xtrabackup::utils::human_readable`). Maybe we could use `human_readable` here for consistency?
Same below with `KB`.

storage/innobase/xtrabackup/src/xb_io_probe.h
```diff
@@ -0,0 +1,212 @@
...
+sequential samples would not reach steady state); the probe rejects
+them and the caller keeps the fallback read request cost. */
+constexpr uint64_t PROBE_MIN_FILE_BYTES = 64 * 1024 * 1024;
+
+/** Result of probe_storage(); valid is false when the file could not
+be opened, is smaller than PROBE_MIN_FILE_BYTES, or a read failed. */
+struct Probe_result {
+  uint64_t rtt_us{0};           /*!< per-request round trip, median */
+  uint64_t bw_bytes_per_sec{0}; /*!< sequential read bandwidth */
+  bool valid{false};
```
jakub-nowakowski-percona, line 58: Alternatively, `probe_storage` could return `std::optional<Probe_result>`.

storage/innobase/xtrabackup/src/xb_io_probe.h
```diff
@@ -0,0 +1,212 @@
...
+    }
+    rtt_samples.push_back(
+        duration_cast<microseconds>(clock::now() - start).count());
+  }
+  std::nth_element(rtt_samples.begin(),
+                   rtt_samples.begin() + rtt_samples.size() / 2,
+                   rtt_samples.end());
+  result.rtt_us = std::max<uint64_t>(rtt_samples[rtt_samples.size() / 2], 1);
+
+  /* contiguous chunks from the middle of the file -> bandwidth.
```
jakub-nowakowski-percona, line 131: Parts of `fd` could have been cashed by the OS in the first phase of the measurement (round trip). Could it affect bandwidth measurement in any significant way? Should we run this benchmark twice, and discard the first measurement to account for that?

storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -112,6 +124,37 @@ static void rf_page_tracking_init(xb_read_filt_ctxt_t *ctxt,
...
+  }
+
+  /* A pinned --page-tracking-max-gap is expressed in innodb_page_size
+  pages; scale it by the physical page size so compressed tablespaces
+  (zip size 1K-8K) combine across the same byte limit instead of a
+  proportionally smaller one. In auto mode (the default) the limit is
+  the read request cost - measured single-threaded at backup start, or
+  the conservative fallback - converted to pages of this tablespace's
+  physical page size: combine every gap cheaper than one saved read. */
+  ut_ad(UNIV_PAGE_SIZE % ctxt->page_size == 0);
```
jakub-nowakowski-percona, line 144: What if this condition isn't met in release build? Should we silently continue?
Let's ensure we are dealing with an InnoDB file.

storage/innobase/xtrabackup/src/read_filt.cc
```diff
@@ -210,9 +253,26 @@ static void rf_page_tracking_get_next_batch(xb_fil_cur_t *cursor,
...
       ctxt->offset = next_page_id * ctxt->page_size;
       /* Find the end of the current page tracking block */
-      pagetracking::range_get_next_page(space);
+      {
+        ulint filler = 0, combined = 0;
+        pagetracking::range_get_next_page(space, ctxt->max_gap, &filler,
+                                          &combined);
+        ctxt->stat_filler_pages += filler;
+        ctxt->stat_combined_gaps += combined;
+      }
```
jakub-nowakowski-percona, line 275: Why do we need additional block brackets here?

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: As discussed, it is very tricky to get the option value correct for users,I think auto is the best. The option exists primarily to revert to old behavior or advanced tuning(user finds out the auto value chosen is bad).

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Done

storage/innobase/xtrabackup/src/xtrabackup.cc
satya-bodapati: Done

storage/innobase/xtrabackup/src/xb_io_probe.h
satya-bodapati: Done

storage/innobase/xtrabackup/src/xb_io_probe.h
satya-bodapati: Its rewritten a bit now. we use O_DIRECT and I dont think multiple trials are not required ebcause we do muiltple samples anyway

satya-bodapati: Logic is interms of pages. ALso since we use O_DIRECT, we do aligned reads.

storage/innobase/xtrabackup/src/read_filt.cc
satya-bodapati: the files discovered at the moment are innodb only. and every innodb file is page size unit based. Also we have minimum file size restriction. so we always read by certian unit size.

jakub-nowakowski-percona (approved):

jakub-nowakowski-percona (approved):

satya-bodapati:
https://pxb.cd.percona.com/job/percona-xtrabackup-8.1-test-param/187/#showFailuresLink

jenksins results are good now. I fixed the test to get required pages to be modified

PR #1784, opened 2026-08-31
PXB-3863 [8.4]: Fix xbcloud with Azure Hierarchical Namespace (HNS) storage

jakub-nowakowski-percona:
https://perconadev.atlassian.net/browse/PXB-3863

Problem
=======

xbcloud put, get, and delete fail on Azure storage accounts with Hierarchical Namespace (HNS) enabled. Azure HNS silently strips leading slashes from blob names, so a blob uploaded as "/backup/file.00000000000000000000" is stored as
"backup/file.00000000000000000000". When xbcloud later lists the container, the returned names no longer match the expected prefix, causing put to report an incomplete backup and get to fail downloading chunks.

Fix
===

Add a virtual normalize_name() method to Object_store with a default identity implementation. Azure_object_store overrides it to detect HNS via the Get Account Information REST API (x-ms-is-hns-enabled header) and strip leading slashes from the backup name. The HNS detection result is cached in a std::optional<bool> to avoid repeated REST calls. After the object store is constructed in main(), backup_name is passed through normalize_name() so all subsequent operations use the normalized path.

Additionally fix two pre-existing bugs found during investigation:

1. Remove trim() on the base64-decoded access key in Azure_signer::sign_request(). The decoded key is arbitrary binary data; trimming it corrupts keys whose decoded form happens to start or end with whitespace-like bytes (0x09, 0x0a, 0x0d, 0x20), causing authentication failures on affected accounts.

2. Fix the logical condition in chunk_name_to_file_name(): change (size < 22 && name[size-21] != '.') to use || so that short names are rejected before indexing into them.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/azure.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/azure.cc
@@ -197,8 +197,6 @@ void Azure_signer::sign_request(const std::string &container,
 
   std::string decoded_access_key = base64_decode(access_key);
 
-  trim(decoded_access_key);
-
   auto signature =
       base64_encode(hmac_sha256(decoded_access_key, string_to_sign));
 
--- a/storage/innobase/xtrabackup/src/xbcloud/azure.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/azure.cc
@@ -378,6 +378,32 @@ bool Azure_client::container_exists(const std::string &name, bool &exists) {
   return false;
 }
 
+bool Azure_client::is_hns_enabled() {
+  if (hns_enabled.has_value()) {
+    return hns_enabled.value();
+  }
+
+  Http_request req(Http_request::GET, protocol, host, "/");
+  req.add_param("restype", "account");
+  req.add_param("comp", "properties");
+  signer->sign_request("", "", req, time(0));
+
+  Http_response resp;
+  if (!http_client->make_request(req, resp)) {
+    hns_enabled = false;
+    return false;
+  }
+
+  if (!resp.ok()) {
+    hns_enabled = false;
+    return false;
+  }
+
+  auto it = resp.headers().find("x-ms-is-hns-enabled");
+  hns_enabled = it != resp.headers().end() && it->second == "true";
+  return hns_enabled.value();
+}
+
 bool Azure_client::upload_object(const std::string &container,
                                  const std::string &name,
                                  const Http_buffer &contents) {
--- a/storage/innobase/xtrabackup/src/xbcloud/azure.h
+++ b/storage/innobase/xtrabackup/src/xbcloud/azure.h
@@ -22,6 +22,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #define __XBCLOUD_AZURE_H__
 
 #include <iostream>
+#include <optional>
 #include "object_store.h"
 #include "xbcloud/http.h"
 #include "xbcloud/util.h"
@@ -92,6 +93,8 @@ class Azure_client {
   Http_request::protocol_t protocol;
   ulong max_retries;
   ulong max_backoff;
+
+  std::optional<bool> hns_enabled;
 
   static void upload_callback(Azure_client *client, std::string container,
                               std::string name, Http_request *req,
@@ -170,6 +173,8 @@ class Azure_client {
                                    const std::string &prefix,
                                    std::vector<std::string> &files,
                                    std::vector<std::string> &dirs);
+
+  bool is_hns_enabled();
 
   ulong get_max_retries() { return max_retries; }
 
@@ -265,6 +270,20 @@ class Azure_object_store : public Object_store {
       std::vector<std::string> &dirs) override {
     return azure_client.list_objects_files_and_dirs(container, directory + "/",
                                                     files, dirs);
+  }
+  // Strip leading slashes for HNS-enabled accounts.
+  // Azure HNS silently normalizes blob paths by removing leading slashes,
+  // so "/backup/chunk.00000" is stored as "backup/chunk.00000".  Without
+  // this, xbcloud's expected prefix no longer matches what the API returns.
+  // Example: "/db/mybackup" -> "db/mybackup"
+  // Flat accounts are unaffected — the name is returned as-is.
+  std::string normalize_name(const std::string &name) override {
+    if (azure_client.is_hns_enabled()) {
+      std::string result = name;
+      ltrim_slashes(result);
+      return result;
+    }
+    return name;
   }
 };
 }  // namespace xbcloud
--- a/storage/innobase/xtrabackup/src/xbcloud/util.h
+++ b/storage/innobase/xtrabackup/src/xbcloud/util.h
@@ -65,6 +65,11 @@ static inline void rtrim_slashes(std::string &s) {
           s.end());
 }
 
+static inline void ltrim_slashes(std::string &s) {
+  s.erase(s.begin(),
+          std::find_if(s.begin(), s.end(), [](int ch) { return ch != '/'; }));
+}
+
 static inline void to_lower(std::string &s) {
   s.resize(s.length());
   std::transform(s.begin(), s.end(), s.begin(), ::tolower);
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -1605,6 +1605,13 @@ int main(int argc, char **argv) {
     reinterpret_cast<Azure_object_store *>(object_store.get())
         ->set_extra_http_headers(extra_http_headers);
   }
+
+  backup_name = object_store->normalize_name(backup_name);
+  if (backup_name.empty()) {
+    msg_ts("%s: Backup name is empty after normalization.\n", my_progname);
+    return EXIT_FAILURE;
+  }
+
   /* validation */
   if (opt_threads > 1 && opt_fifo_dir == nullptr && opt_mode != MODE_DELETE) {
     msg_ts(
--- a/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
@@ -1037,7 +1037,7 @@ cleanup:
 @return true in case of success or false otherwise */
 bool chunk_name_to_file_name(const std::string &chunk_name,
                              std::string &file_name, my_off_t &idx) {
-  if (chunk_name.size() < 22 && chunk_name[chunk_name.size() - 21] != '.') {
+  if (chunk_name.size() < 22 || chunk_name[chunk_name.size() - 21] != '.') {
     /* chunk name is invalid */
     return false;
   }
--- a/storage/innobase/xtrabackup/src/xbcloud/object_store.h
+++ b/storage/innobase/xtrabackup/src/xbcloud/object_store.h
@@ -71,6 +71,15 @@ class Object_store {
                                            const std::string &directory,
                                            std::vector<std::string> &files,
                                            std::vector<std::string> &dirs);
+  /**
+   * Normalize a backup name according to storage-specific rules.
+   *
+   * Default implementation returns the name unchanged.
+   *
+   * @param name Backup name to normalize.
+   * @return Normalized name.
+   */
+  virtual std::string normalize_name(const std::string &name) { return name; }
   virtual ~Object_store() {}
 };
 


jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/164/

storage/innobase/xtrabackup/src/xbcloud/azure.h
```diff
@@ -266,6 +271,14 @@ class Azure_object_store : public Object_store {
     return azure_client.list_objects_files_and_dirs(container, directory + "/",
                                                     files, dirs);
   }
+  std::string normalize_name(const std::string &name) override {
```
satya-bodapati, line 280: can you add an example transformation here in code comments

storage/innobase/xtrabackup/src/xbcloud/util.h
```diff
@@ -65,6 +65,11 @@ static inline void rtrim_slashes(std::string &s) {
           s.end());
 }
 
+static inline void ltrim_slashes(std::string &s) {
```
satya-bodapati, line 68: what if there is no leading slash? non-HNS case?

storage/innobase/xtrabackup/src/xbcloud/azure.cc
```diff
@@ -378,6 +376,32 @@ bool Azure_client::container_exists(const std::string &name, bool &exists) {
   return false;
 }
 
+bool Azure_client::is_hns_enabled() {
```
satya-bodapati, line 379: Can you please check if we should utilize this for HNS delete (the other fix I did)?

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -1605,6 +1605,13 @@ int main(int argc, char **argv) {
     reinterpret_cast<Azure_object_store *>(object_store.get())
         ->set_extra_http_headers(extra_http_headers);
   }
+
+  backup_name = object_store->normalize_name(backup_name);
```
satya-bodapati, line 1609: this will ensure it handles delete, get and put in same way?

satya-bodapati (changes requested):
Can you also please paste results from manual testing of various combinations? (Think now the path to object storage is like directory)

1. like if original put /db/hns, what if we pass /db/ only for delete.
2. non existing /db/
3. any other weird combinations :)

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
jakub-nowakowski-percona: Yes, this affects all of these. I tested it, and it works in all scenarios.

storage/innobase/xtrabackup/src/xbcloud/util.h
jakub-nowakowski-percona: If there are no leading slashes then `normalize_name` will return the name user originally provided, without any changes. Trimming is done only for HNS containers.

storage/innobase/xtrabackup/src/xbcloud/azure.cc
jakub-nowakowski-percona: I checked delete with both HNS and flat containers, with `/` and without it. It worked in all scenarios.

storage/innobase/xtrabackup/src/xbcloud/xbcloud.cc
```diff
@@ -1037,7 +1037,7 @@ bool xbcloud_put(Object_store *store, const std::string &container,
 @return true in case of success or false otherwise */
 bool chunk_name_to_file_name(const std::string &chunk_name,
                              std::string &file_name, my_off_t &idx) {
-  if (chunk_name.size() < 22 && chunk_name[chunk_name.size() - 21] != '.') {
+  if (chunk_name.size() < 22 || chunk_name[chunk_name.size() - 21] != '.') {
```
satya-bodapati, line 1040: this is not problem in reality :) all AI tools flag this. this is because everyfile in backup is always prefixed with .0000001 (which is 21 chars long) but I think we can keep it. Otherwise every contribution will have this fix 😄

storage/innobase/xtrabackup/src/xbcloud/util.h
satya-bodapati: But I believe there is a case where trimming is not necessary, even with HNS containers? :thinking: See the bug scenario in the other delete bug (which didn't use any paths at all)?

So do we have
1. HNS enabled with paths passed from the user
2. HSN enabled without any paths passed from the user

storage/innobase/xtrabackup/src/xbcloud/azure.cc
satya-bodapati: I mean, since the original fix was not relying on is_hns_enabled(). Now that we have is_hsn_enabled(), is there anything to improve there?

jakub-nowakowski-percona:
> Can you also please paste results from manual testing of various combinations? (Think now the path to object storage is like directory)
> 
> 1. like if original put /db/hns, what if we pass /db/ only for delete.
> 2. non existing /db/
> 3. any other weird combinations :)

@satya-bodapati  I created a script that run a matrix of tests for this functionality:
- HNS and flat
- xbcloud put, get and delete
- with and without leading `/`

Here are the results: https://gist.github.com/jakub-nowakowski-percona/a78ed9aec05e4066b4d360fc618db0ce

storage/innobase/xtrabackup/src/xbcloud/util.h
jakub-nowakowski-percona: Can you give me an example of such case? I will test it.

storage/innobase/xtrabackup/src/xbcloud/azure.cc
jakub-nowakowski-percona: I don't think there is. Do you have any particular improvements in mind?

storage/innobase/xtrabackup/src/xbcloud/azure.h
jakub-nowakowski-percona: Done

storage/innobase/xtrabackup/src/xbcloud/azure.cc
satya-bodapati: nope, thanks for checking

satya-bodapati:
@jakub-nowakowski-percona Thanks for pasting results. in the bug report, there are two levels in the path. like /db/hns. Can you please test a similar style?

And the combinations, especially this: use /db/hns for put, then use/db/ for delete. or /db/ for get.

satya-bodapati:
@jakub-nowakowski-percona one additional request, can you please move additional fixes to a separate commit

jakub-nowakowski-percona:
> @jakub-nowakowski-percona one additional request, can you please move additional fixes to a separate commit

Sure, I will do that with further changes (if any). So far the only change was adding a comment.

jakub-nowakowski-percona:
> @jakub-nowakowski-percona Thanks for pasting results. in the bug report, there are two levels in the path. like /db/hns. Can you please test a similar style?
> 
> And the combinations, especially this: use /db/hns for put, then use/db/ for delete. or /db/ for get.

Ok, I will run deeper nested tests and paste. Combinations won't change anything because name gets normalized before any operations take place and they all operate on the same normalized name.

satya-bodapati:
> > @jakub-nowakowski-percona one additional request, can you please move additional fixes to a separate commit
> 
> Sure, I will do that with further changes (if any). So far the only change was adding a comment.

@jakub-nowakowski-percona I was referring to only these
Additionally fix two pre-existing bugs found during investigation:

    Remove trim() on the base64-decoded access key in Azure_signer::sign_request(). The decoded key is arbitrary binary data; trimming it corrupts keys whose decoded form happens to start or end with whitespace-like bytes (0x09, 0x0a, 0x0d, 0x20), causing authentication failures on affected accounts.

    Fix the logical condition in chunk_name_to_file_name(): change (size < 22 && name[size-21] != '.') to use || so that short names are rejected before indexing into them.

storage/innobase/xtrabackup/src/xbcloud/util.h
satya-bodapati: ```xtrabackup --user=root --password='XXXXXX' --backup --stream=xbstream --target-dir=/tmp | xbcloud put --azure-storage-account=xbcloudtest --azure-access-key="readacted" --azure-container-name=xbcloudbackupstorage --storage=azure --parallel=4 full_backup_$(date +%F)```

essentially a container name without any prefix slashes. Container is still HNS enabled, though. MySQL database directories become real directories in object storage because of HNS.

Also please test if we can download a specific file (we do support downloading a specific file from cloud). Not very well known though. So in above case, say download

`get full_backup_timestamp/test/t1.ibd`
`get /full_backup_timestamp/test/t1.ibd`

I dont think we support delete of a specific file..

jakub-nowakowski-percona:
@satya-bodapati I created a new test report which contains deeper nested backup names and also mixes slash usage for HNS containers (e.g. `put` with slash, but `get` without it).

https://gist.github.com/jakub-nowakowski-percona/75fb660c7fcb8a3c17da5d6a386df5f7

If you wish I can also send all the 36 individual log files for each operation, but I checked them and they show successful completion.

storage/innobase/xtrabackup/src/xbcloud/util.h
jakub-nowakowski-percona: Names without any slashes work. Downloading single files works too:
```
$ xbcloud put --storage=azure --azure-storage-account=accounthns --azure-access-key=$AZURE_KEY_HNS --azure-container-name=pxb3863 test < backup/test.xbstream
260831 16:20:19 xbcloud: [0] successfully uploaded chunk: test/testfile.txt.00000000000000000000, size: 79
260831 16:20:20 xbcloud: [0] successfully uploaded chunk: test/testfile.txt.00000000000000000001, size: 26
260831 16:20:20 xbcloud: [0] successfully uploaded chunk: test/xtrabackup_tablespaces.00000000000000000000, size: 80
260831 16:20:20 xbcloud: [0] successfully uploaded chunk: test/xtrabackup_tablespaces.00000000000000000001, size: 36
260831 16:20:20 xbcloud: Upload completed.

$ xbcloud get --storage=azure --azure-storage-account=accounthns --azure-access-key=$AZURE_KEY_HNS --azure-container-name=pxb3863 test/testfile.txt.00000000000000000000 > single_file.xbstream
260831 16:24:41 xbcloud: [0] Downloading test/testfile.txt.00000000000000000000.
260831 16:24:41 xbcloud: [0] Download successfull test/testfile.txt.00000000000000000000, size 79
260831 16:24:41 xbcloud: Download completed.

$ xbcloud get --storage=azure --azure-storage-account=accounthns --azure-access-key=$AZURE_KEY_HNS --azure-container-name=pxb3863 /test/testfile.txt.00000000000000000000 > single_file.xbstream
260831 16:24:56 xbcloud: [0] Downloading test/testfile.txt.00000000000000000000.
260831 16:24:56 xbcloud: [0] Download successfull test/testfile.txt.00000000000000000000, size 79
260831 16:24:56 xbcloud: Download completed.
```

jakub-nowakowski-percona:
> > > @jakub-nowakowski-percona one additional request, can you please move additional fixes to a separate commit
> > 
> > 
> > Sure, I will do that with further changes (if any). So far the only change was adding a comment.
> 
> @jakub-nowakowski-percona I was referring to only these Additionally fix two pre-existing bugs found during investigation:
> 
> ```
> Remove trim() on the base64-decoded access key in Azure_signer::sign_request(). The decoded key is arbitrary binary data; trimming it corrupts keys whose decoded form happens to start or end with whitespace-like bytes (0x09, 0x0a, 0x0d, 0x20), causing authentication failures on affected accounts.
> 
> Fix the logical condition in chunk_name_to_file_name(): change (size < 22 && name[size-21] != '.') to use || so that short names are rejected before indexing into them.
> ```

Done

satya-bodapati:
Can you prefix [8.4] in the PR /commit message. You also need to create 9.7 & turnk versions?

satya-bodapati (approved):
LGTM!

storage/innobase/xtrabackup/src/xbcloud/azure.cc
```diff
@@ -378,6 +376,32 @@ bool Azure_client::container_exists(const std::string &name, bool &exists) {
...
+  }
+
+  Http_request req(Http_request::GET, protocol, host, "/");
+  req.add_param("restype", "account");
+  req.add_param("comp", "properties");
+  signer->sign_request("", "", req, time(0));
+
+  Http_response resp;
+  if (!http_client->make_request(req, resp)) {
+    hns_enabled = false;
```
satya-bodapati, line 391: ;optional] better to start with hns_enabled=false before http request?

PR #1787, opened 2026-09-01
PXB-3863 [trunk]: Fix xbcloud with Azure Hierarchical Namespace (HNS) storage

jakub-nowakowski-percona:
https://perconadev.atlassian.net/browse/PXB-3863

Problem
=======

xbcloud put, get, and delete fail on Azure storage accounts with Hierarchical Namespace (HNS) enabled. Azure HNS silently strips leading slashes from blob names, so a blob uploaded as "/backup/file.00000000000000000000" is stored as
"backup/file.00000000000000000000". When xbcloud later lists the container, the returned names no longer match the expected prefix, causing put to report an incomplete backup and get to fail downloading chunks.

Fix
===

Add a virtual normalize_name() method to Object_store with a default identity implementation. Azure_object_store overrides it to detect HNS via the Get Account Information REST API (x-ms-is-hns-enabled header) and strip leading slashes from the backup name. The HNS detection result is cached in a std::optional<bool> to avoid repeated REST calls. After the object store is constructed in main(), backup_name is passed through normalize_name() so all subsequent operations use the normalized path.

Additionally fix two pre-existing bugs found during investigation:

1. Remove trim() on the base64-decoded access key in Azure_signer::sign_request(). The decoded key is arbitrary binary data; trimming it corrupts keys whose decoded form happens to start or end with whitespace-like bytes (0x09, 0x0a, 0x0d, 0x20), causing authentication failures on affected accounts.

2. Fix the logical condition in chunk_name_to_file_name(): change (size < 22 && name[size-21] != '.') to use || so that short names are rejected before indexing into them.

### DIFF
--- a/storage/innobase/xtrabackup/src/xbcloud/azure.cc
+++ b/storage/innobase/xtrabackup/src/xbcloud/azure.cc
@@ -197,8 +197,6 @@ void Azure_signer::sign_request(const std::string &container,
 
   std::string decoded_access_key = base64_decode(access_key);
 
-  trim(decoded_access_key);
-
   auto signature =
       base64_encode(hmac_sha256(decoded_access_key, string_to_sign));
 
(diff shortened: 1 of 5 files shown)


jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%209.x/job/percona-xtrabackup-9.x-multijob/16/

satya-bodapati (approved):

PR #1815, opened 2026-09-14
PXB-2559 [8.4]: Allow --history to write its record to another server

jakub-nowakowski-percona:
https://perconadev.atlassian.net/browse/PXB-2559

--history records the backup in PERCONA_SCHEMA.xtrabackup_history,
creating the schema and the table when they are not there yet. It does
this over the same connection that takes the backup, so the server being
backed up has to accept writes.

A Group Replication secondary and an asynchronous replica both run with
super_read_only=ON, and those are exactly the nodes an operator prefers
to back up. On such a node --history fails with error 1290 after the
backup itself has already been written, leaving the choice between
backing up the primary and giving up the history record.

Let the record be sent elsewhere. --history-host, --history-port,
--history-socket, --history-user and --history-password describe a
connection to a writable server, and --history-ssl-mode,
--history-ssl-ca, --history-ssl-capath, --history-ssl-cert and
--history-ssl-key secure it independently of the backup connection. Each
falls back to its backup counterpart when not given, so only the options
that actually differ have to be spelled out. Passing any of them without
--history, --incremental-history-name or --incremental-history-uuid is
rejected rather than quietly ignored.

The history server is reached over a second permanent connection.
mysql_history_connection is opened in xb_init() alongside
mysql_connection and aliases it when no --history-* option is given, so
a backup that keeps its history locally behaves as before and opens no
extra connection. Both the write path in write_xtrabackup_info() and the
read path in select_incremental_lsn_from_history() use it, so
--incremental-history-name and --incremental-history-uuid look for their
base LSN on the server that stores the records.

Opening that connection early means the host, the credentials and the
TLS settings are checked before any data is copied. A history server
that cannot be reached fails the backup up front, instead of letting it
run to completion and then exiting 0 with no record written.

check_all_privileges() validates the history table against the grants
held by the history account on the history server. With --history it
requires CREATE, ALTER and INSERT; with --incremental-history-name or
--incremental-history-uuid it requires SELECT. CREATE is required even
when the schema and the table already exist, because CREATE DATABASE IF
NOT EXISTS and CREATE TABLE IF NOT EXISTS are issued on every run and
the server checks the privilege before it checks existence.

--history-password is hidden the way --password is: from the process
list, from the recognized-arguments log line, from --help output and
from the tool_command stored in the history record. Failure to apply the
TLS settings aborts the connection rather than silently leaving it
unprotected.

A failure of the statement that writes the record is now reported. The
exit code is deliberately left unchanged, but the record no longer goes
missing without a word in the log.

Tests: history_dsn covers the DSN options, the incremental read path,
the early validation, the privilege check and --history-ssl-mode on two
plain servers; gr/history_host covers the super_read_only secondary the
feature exists for; check_for_passwords gained --history-password.


jakub-nowakowski-percona:
Tests: https://pxb.cd.percona.com/view/PXB%208.1/job/percona-xtrabackup-8.1-multijob/173/

